PR #1, opened 2023-02-08
mrc-3968 skeleton repo

hillalex:
This is just the result of running `gradle init` and `npm init` in the relevant folders, plus a dummy `schema.sql` file for the db.


LekanAnni (approved):
Excellent 👍

PR #5, opened 2023-02-16
mrc- 3976 add spring boot, hibernate and spring data

LekanAnni:
Add spring boot


api/app/build.gradle.kts
```diff
@@ -1,19 +1,22 @@
...
- * For more details take a look at the 'Building Java & JVM projects' chapter in the Gradle
- * User Manual available at https://docs.gradle.org/7.6/userguide/building_java_projects.html
- */
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     // Apply the org.jetbrains.kotlin.jvm Plugin to add support for Kotlin.
-    id("org.jetbrains.kotlin.jvm") version "1.7.10"
+    kotlin("jvm") version "1.7.22"
+    //id("org.jetbrains.kotlin.jvm") version "1.7.10"
```
hillalex, line 6: can this commented line be removed?

hillalex (changes requested):
This looks good, can we have a README and a github action as well? Also, I think this was introduced in the last PR but I didn't notice - it looks like `make db` in the github action is not building the docker image, see: https://github.com/mrc-ide/packit/actions/runs/4195595003/jobs/7275362266#step:6:7 (and see pulling `mrcide/packit-db:mrc-3976` returns a "manifest does not exist" error). I'm not so familiar with `make` I'm afraid, but I guess it needs a flag to tell it to always run that task, i.e. never assume it is "up to date" (since we want a docker image for every branch)?

api/app/src/main/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+spring.datasource.url=jdbc:postgresql://packit:5452/packit
+spring.datasource.username=
+spring.datasource.password=
```
hillalex, line 3:
```suggestion
spring.datasource.url=jdbc:postgresql://localhost:5432/packit
spring.datasource.username=packituser
spring.datasource.password=changeme
```

this is what I needed to get it connecting to the db

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v3
+      - name: Set up Java JDK ${{ matrix.java-version }}
+        uses: actions/setup-java@v3
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: "temurin"
+          cache: "gradle"
+      - name: test backend
+        working-directory: api
+        run: ./gradlew :app:test --stacktrace
```
hillalex, line 27: So now we have
* a shell script for the db
* a `Makefile` for the front-end 
* raw bash commands like this for the back-end

I think we should pick an approach (either `Make` or shell scripts) and use it consistently throughout!

Makefile
```diff
@@ -13,6 +13,3 @@ frontend:
 #Install dependencies, run eslint, build and test frontend
 frontend.deploy:
 	$(MAKE) frontend
-
```
hillalex, line 16: also, why do we have 2 tasks here, one that just calls the other? can't we just get rid of `frontend.deploy` here and always use `frontend`?

hillalex (commented):
Just a couple more questions about the utility of the Makefile!

.github/workflows/backend-test.yml
LekanAnni: Better to get rid of Makefile, the combination is very confusing.

db/scripts/run
```diff
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
 set -e
 GIT_BRANCH=$(git -C "$PACKAGE_ROOT" symbolic-ref --short HEAD)
-docker run --rm -d --network=host --name packit-db mrcide/packit-db:$GIT_BRANCH
+docker run --rm -d \
+ --network=bridge \
```
LekanAnni, line 5: I changed this to a bridge network as localhost was getting propagated in host computer on Mac. We should create a custom network when we dockerize all apps for communication between containers.

hillalex (approved):

PR #6, opened 2023-02-21
Mrc-3969 Create dummy endpoint for returning index data

LekanAnni:
This PR adds dummy endpoint for `index/packet` and make dummy data available for db.

* I have added database `flyway` migration framework.
* I added h2 in memory database for testing the db. however I have ignored the only test using h2 due to an issue with h2 converting tables to uppercase. I could not fix that on time so I have created a ticket to do so https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4012/Fix-tables-being-transformed-to-UPPERCASE-in-H2.
* I deleted the schema folder from db Dockerfile as migration is now ran dynamically.


.github/workflows/backend-test.yml
```diff
@@ -13,9 +13,27 @@ jobs:
     strategy:
       matrix:
         java-version: [ 17 ]
+    # Service containers to run with `runner-job`
+    services:
+      packit-db:
+        image: mrcide/packit-db:${{ github.head_ref || github.ref_name }}
+        # Set health checks to wait until postgres has started and add container name
+        options: >-
+          --health-cmd pg_isready
```
hillalex, line 22: this is cool, so much easier than managing all this ourselves on buildkite!

db/dummy/packet.json
```diff
@@ -0,0 +1,1000 @@
+{"id":1,"name":"Filbert","display_name":"Filbert Bundock","parameters":[{},{}],"published":false}
```
hillalex, line 1: `parameters` will be a dictionary rather than an array, so like `"parameters": {a: 1, b: 2}`

scripts/run-dependencies
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+HERE=$(realpath "$(dirname $0)")
+
+## Run database and add dummy
```
hillalex, line 7:
```suggestion
## Start database 
```

scripts/run-dependencies
```diff
@@ -0,0 +1,12 @@
...
+"$HERE"/../db/scripts/build
+"$HERE"/../db/scripts/run
+
+# Populate db
```
hillalex, line 11:
```suggestion
# Populate db with dummy packet data
```

api/app/src/main/kotlin/packit/repository/IndexRepository.kt
```diff
@@ -0,0 +1,8 @@
+package packit.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import packit.model.Packet
+
+@Repository
+interface IndexRepository: JpaRepository<Packet, Long>
```
hillalex, line 8: I guess we don't have detekt in here yet (would be a good idea to add that next!) - these files should all  have empty lines at the end.

api/app/src/main/kotlin/packit/service/BaseIndexService.kt
```diff
@@ -0,0 +1,20 @@
+package packit.service
+
+import org.springframework.stereotype.Service
+import packit.model.Packet
+import packit.repository.IndexRepository
+
+
+interface IndexService
+{
+    fun getPacket(): List<Packet>
```
hillalex, line 10:
```suggestion
    fun getPackets(): List<Packet>
```

api/app/src/main/kotlin/packit/service/BaseIndexService.kt
```diff
@@ -0,0 +1,20 @@
...
+}
+
+@Service
+class BaseIndexService(private val indexRepository: IndexRepository) : IndexService
+{
+    override fun getPacket(): List<Packet>
```
hillalex, line 16:
```suggestion
    override fun getPackets(): List<Packet>
```

api/app/src/main/kotlin/packit/model/Packet.kt
```diff
@@ -0,0 +1,19 @@
+package packit.model
+
+import jakarta.persistence.*
+
+@Entity
+@Table(name = "packet")
+data class Packet(
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
```
hillalex, line 9: These ids will always come from `outpack` and be Strings, so I don't think we want this `GeneratedValue` attribute

hillalex (changes requested):

api/app/src/main/kotlin/packit/model/Packet.kt
```diff
@@ -0,0 +1,18 @@
+package packit.model
+
+import jakarta.persistence.*
+
+@Entity
+@Table(name= "packet")
+data class Packet(
+    @Id
+    var id: String,
```
LekanAnni, line 9: I am using var as kotlin data class is immutable by design.

.github/workflows/backend-test.yml
LekanAnni: It's awesome. We can do the same in docker-compose file. I think we should properly have a ticket for dockerising the apps.

api/app/src/main/kotlin/packit/controllers/IndexController.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import packit.model.Packet
+import packit.service.IndexService
+
+@RestController
+@RequestMapping("/index")
```
hillalex, line 11: Sorry I didn't catch this before. I think a more natural naming convention here would have `PacketController` with` @RequestMapping("/packet")` and then the below the method is `fun index()` with `@GetMapping("/")`. Then later we will also have a `fun packet(id)` with `@GetMapping("/<id>")`

api/app/src/main/kotlin/packit/repository/IndexRepository.kt
```diff
@@ -0,0 +1,8 @@
+package packit.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import packit.model.Packet
+
+@Repository
+interface IndexRepository: JpaRepository<Packet, Long>{}
```
hillalex, line 8:
```suggestion
interface PacketRepository: JpaRepository<Packet, Long>{}
```

api/app/src/main/kotlin/packit/service/BaseIndexService.kt
```diff
@@ -0,0 +1,20 @@
+package packit.service
+
+import org.springframework.stereotype.Service
+import packit.model.Packet
+import packit.repository.IndexRepository
+
+
+interface IndexService
```
hillalex, line 8:
```suggestion
interface PacketService
```

api/app/src/main/kotlin/packit/service/BaseIndexService.kt
```diff
@@ -0,0 +1,20 @@
...
+{
+    fun getPackets(): List<Packet>
+}
+
+@Service
+class BaseIndexService(private val indexRepository: IndexRepository) : IndexService
```
hillalex, line 14:
```suggestion
class BasePacketService(private val indexRepository: IndexRepository) : IndexService
```

scripts/run-dependencies
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+HERE=$(realpath "$(dirname $0)")
+
+## Start database
+"$HERE"/../db/scripts/build
+"$HERE"/../db/scripts/run
+
+# Populate db with dummy data
+docker exec packit-db psql -U packituser -d packit -a -f /packit-dummy/import.sql
```
hillalex, line 12: This fails because the db isn't ready, you need to leave [this line](https://github.com/mrc-ide/packit/pull/6/files#diff-4dd3e8495a99556b02e15b8567535ee7ad8ad4ef83868fca5b1a37e5556d481dL11) in place. But this will also fail because you have moved initialising the schema out of `db/scripts/run`, so the `packet` table won't exist when you try to import the dummy data!

api/README.md
```diff
@@ -16,3 +16,8 @@ Execute tests on the command line or through IntelliJ
 To run a specific test alone, add `--test` + the \
 [fully qualified class name](https://docs.gradle.org/current/userguide/java_testing.html#full_qualified_name_pattern)\
 to the command. For example, the command for running AppTest.kt would be: `./api/gradlew -p api/app :app:test --tests AppTest`
+
+## Database migration
+Database uses [flyway](https://flywaydb.org/documentation/database/h2) to run migration, the scripts are located in \
+`api/resources/db/migration`. Follow the naming pattern in migration folder to alter the schema \
```
hillalex, line 22: This is a problem for adding the dummy data - the schema needs to exist before the data is imported. I suggest you back this out for now and stick with the simpler approach of just running the schema import file (no Flyway). We did say we don't need to worry about migrations for now, since we can just destroy and rebuild the db every time we deploy.

hillalex (changes requested):
A few things:
* Naming of controller/service/repo (`index` -> `packet`)
* The order of things is a bit confused, need to wait for db to be ready, then import schema, then import dummy data. 
* The github action is not working as expected. Importing the dummy data is failing here for the reason given above: https://github.com/mrc-ide/packit/actions/runs/4242185644/jobs/7373361383#step:4:20 The failure of this step should probably just fail the build, but even if it doesn't then the tests will fail; the fact the build passed suggests the tests are not being run. 
* The format of the `parameters` column in the dummy data is wrong. This field should be a dictionary object, not an array, e.g. `"parameters": {a: 1, b: 2}`.

LekanAnni:
This ticket will be closed. I have left it for reference purpose until mrc-3969 is merged.

PR #7, opened 2023-02-23
Mrc-3969-revert Create dummy endpoint for returning index data

LekanAnni:
This PR adds dummy endpoint for /packet/ endpoint and make dummy data available for db.

I am closing mrc-3969, I had issues pushing and I didn't want to force push. I have discarded flyway for now. I have updated dummy data to include the correct datatypes.


.github/workflows/backend-test.yml
```diff
@@ -1,6 +1,10 @@
 name: backend-test
 run-name: ${{ github.actor }} is testing packit backend features
 on:
+  workflow_run:
+    workflows: [ "build-db" ]
+    types:
+      - completed
```
LekanAnni, line 7: db will be built before running backend workflow

api/app/src/main/kotlin/packit/model/Packet.kt
```diff
@@ -0,0 +1,25 @@
...
+@Entity
+@Table(name = "packet")
+data class Packet(
+    @Id
+    val id: String,
+    val name: String,
+    val displayName: String,
+    @Column(name = "parameters", columnDefinition = "json")
+    val parameters: String,
+    val published: Boolean,
```
LekanAnni, line 15: I have restored val for simplicity. I think we should observe the immutability but this seems to work for now.

hillalex (approved):
Great. Just one last thing: the endpoint works for `http://localhost:8080/packet/` but not `http://localhost:8080/packet`. The trailing slash should be optional, but feel free to add a new ticket for this.

LekanAnni:
> Great. Just one last thing: the endpoint works for `http://localhost:8080/packet/` but not `http://localhost:8080/packet`. The trailing slash should be optional, but feel free to add a new ticket for this.

Done

PR #11, opened 2023-04-05
mrc-4129 Sort table column on packet explorer page

LekanAnni:
This PR adds functionality to sort sortable table data.

### DIFF
--- a/app/src/app/components/contents/explorer/Table.tsx
+++ b/app/src/app/components/contents/explorer/Table.tsx
@@ -1,5 +1,5 @@
 import React, {useState} from "react";
-import {AssessorsProperty, Header, PacketTableProps} from "../../../../types";
+import {Header, Packet, PacketTableProps} from "../../../../types";
 import {FaSort} from "react-icons/fa";
 
 const headers: Header[] = [
@@ -15,8 +15,8 @@ export default function Table({data}: PacketTableProps) {
     const sortPackets = (accessor: string) => {
         data.sort((a, b) => {
             setAscending(!ascending);
-            return (`${a[accessor as keyof AssessorsProperty]}`
-                .localeCompare(`${b[accessor as keyof AssessorsProperty]}`)) * (ascending ? -1 : 1);
+            return (`${a[accessor as keyof Packet]}`
+                .localeCompare(`${b[accessor as keyof Packet]}`)) * (ascending ? -1 : 1);
         });
     };
 
--- a/app/src/app/components/contents/explorer/Table.tsx
+++ b/app/src/app/components/contents/explorer/Table.tsx
@@ -10,14 +10,22 @@ const headers: Header[] = [
 ];
 
 export default function Table({data}: PacketTableProps) {
-    const [ascending, setAscending] = useState(true);
+    const [ascendingName, setAscendingName] = useState(false);
+    const [ascendingStatus, setAscendingStatus] = useState(false);
 
     const sortPackets = (accessor: string) => {
-        data.sort((a, b) => {
-            setAscending(!ascending);
-            return (`${a[accessor as keyof Packet]}`
-                .localeCompare(`${b[accessor as keyof Packet]}`)) * (ascending ? -1 : 1);
-        });
+        let ascending = false;
+        if (accessor === "displayName") {
+            ascending = ascendingName;
+            setAscendingName(!ascending);
+            setAscendingStatus(false);
+        } else if (accessor === "published") {
+            ascending = ascendingStatus;
+            setAscendingStatus(!ascending);
+            setAscendingName(false);
+        }
+        data.sort((a, b) => (`${a[accessor as keyof Packet]}`
+            .localeCompare(`${b[accessor as keyof Packet]}`)) * (ascending ? -1 : 1));
     };
 
     return (
--- a/app/src/app/components/contents/explorer/Table.tsx
+++ b/app/src/app/components/contents/explorer/Table.tsx
@@ -1,21 +1,37 @@
-import React from "react";
-import {Header, PacketTableProps} from "../../../../types";
+import React, {useState} from "react";
+import {AssessorsProperty, Header, PacketTableProps} from "../../../../types";
+import {FaSort} from "react-icons/fa";
+
+const headers: Header[] = [
+    {label: "Name", accessor: "displayName", sortable: true},
+    {label: "Version", accessor: "id", sortable: false},
+    {label: "Status", accessor: "published", sortable: true},
+    {label: "Parameters", accessor: "parameters", sortable: false}
+];
 
 export default function Table({data}: PacketTableProps) {
-    const headers: Header[] = [
-        {label: "Name", accessor: "name", sortable: true},
-        {label: "Version", accessor: "version", sortable: true},
-        {label: "Status", accessor: "status", sortable: true},
-        {label: "Parameters", accessor: "parameters", sortable: false}
-    ];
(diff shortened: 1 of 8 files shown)


app/src/tests/components/contents/explorer/Table.test.tsx
```diff
@@ -55,50 +78,94 @@ describe("table component", () => {
...
+        expect(rows[2].querySelectorAll("td")[0]).toHaveTextContent("packet 2");
+        expect(rows[3].querySelectorAll("td")[0]).toHaveTextContent("packet 1");
+    });
+
+    it("can sort data by status header as expected", async () => {
+        const store = getStore();
+
+        render(<Provider store={store}> <Table data={packets}/></Provider>);
+
+        const {getByText, getByTestId} = screen;
```
EmmaLRussell, line 155: Unused `getByTestId` here

app/src/app/store/packets/packets.ts
```diff
@@ -14,7 +14,7 @@ export const packetsSlice = createSlice({
     extraReducers: (builder) => {
         builder
             .addCase(actions.fetchPackets.fulfilled, (state, action) => {
-                state.packets = action.payload;
+                state.packets = action.payload.sort((a, b) => a.name.localeCompare(b.name));
```
EmmaLRussell, line 17: Should be a test for this default sort?

EmmaLRussell (commented):
This looks good! I guess the sortable columns will evolve a bit as new features are added to the index page (we're going to get date eventually?). I'm not sure how meaningful it is to sort by Status really, but this is good for proof of concept. 

The only thing that seems a bit odd about the behaviour results from having a single `ascending` flag which gets toggled even when you click on another column. So if you click on "Name" a couple of times so that the rows go back to being sorted ascending, then click on "Status" I'd expect Status to sort ascending initially, but it actually sorts descending as the flag has toggled to false..

LekanAnni:
> This looks good! I guess the sortable columns will evolve a bit as new features are added to the index page (we're going to get date eventually?). I'm not sure how meaningful it is to sort by Status really, but this is good for proof of concept.
> 
> The only thing that seems a bit odd about the behaviour results from having a single `ascending` flag which gets toggled even when you click on another column. So if you click on "Name" a couple of times so that the rows go back to being sorted ascending, then click on "Status" I'd expect Status to sort ascending initially, but it actually sorts descending as the flag has toggled to false..

I am not sure I understand what you mean. This works as expected, initial sorting is by `display name` in mutation. I think this will do for a POC

EmmaLRussell:
> > This looks good! I guess the sortable columns will evolve a bit as new features are added to the index page (we're going to get date eventually?). I'm not sure how meaningful it is to sort by Status really, but this is good for proof of concept.
> > The only thing that seems a bit odd about the behaviour results from having a single `ascending` flag which gets toggled even when you click on another column. So if you click on "Name" a couple of times so that the rows go back to being sorted ascending, then click on "Status" I'd expect Status to sort ascending initially, but it actually sorts descending as the flag has toggled to false..
> 
> I am not sure I understand what you mean. This works as expected, initial sorting is by `display name` in mutation. I think this will do for a POC

What I mean is, I think the user would expect the first time they click on a column that it would sort ascending by that column, and it doesn't if you follow the sequence I describe.

EmmaLRussell:
I just tried it again and, actually, you don't need to do the initial sorts on the name column. If the first sort click you do is on the Status column, it sorts it descending, whereas I think the expectation would be that it would sort ascending initially.

app/src/app/store/packets/packets.ts
```diff
@@ -14,7 +14,7 @@ export const packetsSlice = createSlice({
     extraReducers: (builder) => {
         builder
             .addCase(actions.fetchPackets.fulfilled, (state, action) => {
-                state.packets = action.payload.sort((a, b) => a.name.localeCompare(b.name));
+                state.packets = action.payload;
```
EmmaLRussell: Sorry, I don't really understand this change. I wasn't suggesting that you change the initial ordering from sorted by name, that seems fine. I was just saying that when the user then wants to sort by another column (clicks on 'Status'), it should sort ascending since that's the first Status sort. 

It should probably toggle the ascending flag on sort, unless a new column from the previous sort has been selected, in which case it should always revert to ascending = true.

I might not have explained this very well, happy to chat about it!

hillalex (commented):
Everything looks good to me aside from agreeing with Emma about the sort behaviour.

app/src/app/components/contents/explorer/Table.tsx
```diff
@@ -10,14 +10,22 @@ const headers: Header[] = [
...
+            ascending = ascendingName;
+            setAscendingName(!ascending);
+            setAscendingStatus(false);
+        } else if (accessor === "published") {
+            ascending = ascendingStatus;
+            setAscendingStatus(!ascending);
+            setAscendingName(false);
+        }
+        data.sort((a, b) => (`${a[accessor as keyof Packet]}`
+            .localeCompare(`${b[accessor as keyof Packet]}`)) * (ascending ? -1 : 1));
```
EmmaLRussell, line 28:
```suggestion
        const newAscending = accessor === ascendingCol ? !ascending : true;
        setAscending(newAscending);
        setAscendingCol(accessor);
        
        data.sort((a, b) => (`${a[accessor as keyof Packet]}`
            .localeCompare(`${b[accessor as keyof Packet]}`)) * (newAscending ? -1 : 1));
```

app/src/app/components/contents/explorer/Table.tsx
```diff
@@ -10,14 +10,22 @@ const headers: Header[] = [
 ];
 
 export default function Table({data}: PacketTableProps) {
-    const [ascending, setAscending] = useState(true);
+    const [ascendingName, setAscendingName] = useState(false);
+    const [ascendingStatus, setAscendingStatus] = useState(false);
```
EmmaLRussell, line 14: I really meant doing a more general approach here - I wouldn't be surprised if more sortable columns come along, and I don't think there's much value in retaining `ascending` values for columns which aren't the current sort column. I think you'd always expect the column to sort ascending when you switch from one sort column to another.

Something like the following..?

```suggestion
    const [ascending, setAscending] = useState(true);
    const [ascendingCol, setAscendingCol] = useState("");
```

app/src/app/components/contents/explorer/Table.tsx
EmmaLRussell: Sorry, I should have called it `sortCol'`rather than `ascendingCol`, I just meant the column that you're sorting by (whether asc or desc), and `ascending` is still the flag to say whether it's currently ascending or descending. 
So if you've got sortCol="displayName" and ascending=false, then you click on "status", it should then set `newAscending` to true since accessor ("status") != sortCol ("displayName"), so it should then sort by status ascending.

EmmaLRussell: OK, I think I just had the sort direction the wrong way round. This works for me:
```
    const [ascending, setAscending] = useState(true);
    const [sortCol, setSortCol] = useState("");

    const sortPackets = (accessor: string) => {
        const newAscending = (accessor === sortCol) ? !ascending : true;
        setAscending(newAscending);
        setSortCol(accessor);

        data.sort((a, b) => (`${a[accessor as keyof Packet]}`
            .localeCompare(`${b[accessor as keyof Packet]}`)) * (newAscending ? 1 : -1));
    };
```

PR #12, opened 2023-05-02
mrc-4189 scheduled job to import packets

hillalex:
This sets up the scheduled job which reads the checksum from outpack and determines whether packets should be refreshed. Actual code for importing the packets in a further PR.

Also includes a bit of simplication around running dependencies/test setup, and a demo orderly directory for testsing (see the README in the demo directory https://github.com/mrc-ide/packit/pull/12/files#diff-aec57cd2b5263590e3454601b8012373c551a611ed30c124b5f452f5103ad6c8)


.github/workflows/backend-test.yml
```diff
@@ -11,33 +11,21 @@ on:
...
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-      - name: Connect to PostgreSQL
-        run: db/scripts/add-schema-and-dummy
-        env:
-          POSTGRES_HOST: localhost
-          POSTGRES_PORT: 5432
+      - name: Run dependencies
+        run: ./scripts/run-dependencies
```
hillalex, line 28: consolidated our approach to running up dependencies - this script can be used for local dev and for testing

api/app/src/main/kotlin/packit/service/BasePacketService.kt
```diff
@@ -16,4 +18,25 @@ class BasePacketService(private val packetRepository: PacketRepository) : Packet
     {
         return packetRepository.findAll()
     }
+
+    override fun getChecksum(): String
+    {
+        return packetRepository.findAllIds()
```
EmmaLRussell, line 24: Are we confident that we'll be assembling our ids in the same order as ourpack when calculating the two checksums?

api/app/src/main/kotlin/packit/service/OutpackServerClient.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    fun getChecksum(): String
+    {
+        return get("checksum")
+    }
+
+    fun getMetadata(): List<OutpackMetadata>
+    {
+        val url = "$baseUrl/metadata"
+        val response = restTemplate.exchange(
```
EmmaLRussell, line 30: Couldn't this use the `get` method? Also, not sure why `getMetadata` is in this PR - doesn't seem to be related to the scheduled job?

EmmaLRussell (commented):
I'm seeing an error logged when I run the app using `./api/gradlew -p api/app :app:run`:
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'appConfig': Injection of autowired dependencies failed
```
Dependencies are running. Is there anything else I need to do?

api/app/src/main/kotlin/packit/service/BasePacketService.kt
hillalex: good point, no! this should be sorted.

api/app/src/main/kotlin/packit/service/OutpackServerClient.kt
hillalex: You're right, it's used in the next one. Can remove from here. See comment on the `get` method - unfortunately passing around generic types doesn't work: https://github.com/mrc-ide/packit/pull/12/files/9448f2474fd6bface58247c8655124aab039696b#diff-07850fffb48c54cb4734c8aa8681a522c8fe3cd48c45d1620b42c4648e555040R41

hillalex:
Ok, removed that unused method and added explicit sorting of ids. `AppConfig` issue should be fixed now too, it was because I'd erroneously changed a property name in application.properties. We need some blackbox tests to catch bugs like that (also to test the scheduler). I'll add a ticket for that.

hillalex:
Ok, removed that unused method and added explicit sorting of ids. `AppConfig` issue should be fixed now too, it was because I'd erroneously changed a property name in application.properties. We need some blackbox tests to catch bugs like that (also to test the scheduler properly). I'll add a ticket for that.

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
```diff
@@ -8,6 +8,6 @@ import packit.model.Packet
 @Repository
 interface PacketRepository : JpaRepository<Packet, String>
 {
-    @Query("select p.id from Packet p")
+    @Query("select p.id from Packet p order by p.id asc")
     fun findAllIds(): List<String>
```
LekanAnni: You surely don't need the Query. `findAllById` will return all Ids. `findAllByIdOrderByIdAsc` will return ordered list.I think you don't need to write anyCode inside the this repository.

LekanAnni: I guess it might not work because the Id is stored as String.

hillalex: You do need to do a little work to select only a single column, see e.g. https://stackoverflow.com/questions/22007341/spring-jpa-selecting-specific-columns. Using a specific query seemed like the neatest way to me (creating a custom object with only the id field just adds more boilerplate).

hillalex: You do need to do a little work to select only a single column, see e.g. https://stackoverflow.com/questions/22007341/spring-jpa-selecting-specific-columns. Using a specific query seemed like the neatest way to me (creating a custom object with only the id field just adds more boilerplate).

EmmaLRussell (approved):

api/app/src/main/kotlin/packit/service/Scheduler.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.stereotype.Service
+
+@Service
+class Scheduler(
+        private val packetService: PacketService,
+        private val outpackServerClient: OutpackServerClient
+)
+{
+
+    @Scheduled(fixedDelay = 1000)
```
LekanAnni, line 14: I think polling interval is too short, I do not think data will change every second, we could make it 10 seconds.

LekanAnni (changes requested):
Looks great, 

Just a concern regarding scheduling interval.

LekanAnni (approved):
LGTM 👍

PR #13, opened 2023-05-02
mrc-3977 add detekt for kotlin linting

hillalex:
Also commits the Intellij code style files and applies styles to the whole project. Using the same style rules as OrderlyWeb.


EmmaLRussell (commented):
LGTM. Seeing this on the build:
```
Property 'formatting>TrailingComma' is deprecated. Rule is split between `TrailingCommaOnCallSite` and `TrailingCommaOnDeclarationSite` now..
```
..thought not sure where `TrailingComma` is being referenced - it doesn't seem to be in detekt.yml

hillalex:
> ..thought not sure where `TrailingComma` is being referenced - it doesn't seem to be in detekt.yml

Looks like its deprecated in ktlint which the detekt default ruleset is built on. Guessing this warning will go away in future releases of detekt.

PR #14, opened 2023-05-04
mrc-4189 import packet data via the scheduler

hillalex:
~~Depends on https://github.com/mrc-ide/packit/pull/12~~

This does the actual importing of packets.

Also a bit of test tidy up - the database tests were not using Postgres and failing to catch bugs as a consequence. 

Dummy data now removed as test packet data is imported via the scheduler and `outpack_server`.


api/app/src/main/kotlin/packit/repository/PacketRepository.kt
```diff
@@ -10,4 +10,7 @@ interface PacketRepository : JpaRepository<Packet, String>
 {
     @Query("select p.id from Packet p order by p.id asc")
     fun findAllIds(): List<String>
+
+    @Query("select p from Packet p order by p.time desc limit 1")
```
hillalex, line 14: Using the JPA built in method syntax (like `findTopOrderByTimeDesc`) gives a `SqlGrammarException` and I can't for the life of me work out why. The Postgres dialect seems to be set correctly and the error is really opaque. So I gave up and did this instead.

api/app/src/main/resources/application-test.properties
```diff
@@ -1,7 +1,8 @@
-spring.datasource.driver-class-name=org.h2.Driver
-spring.datasource.url=jdbc:h2:mem:test
-spring.datasource.username=sa
-spring.datasource.password=
-spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.H2Dialect
```
hillalex, line 5: Using a fake db server and a different SQL dialect was resulting in bugs not being surfaced in the tests. There's no need to write unit tests against repositories because they don't contain any logic - all we care about is whether they integrate correctly with a Postgres database.

api/app/src/main/resources/application.properties
```diff
@@ -1,8 +1,9 @@
-spring.datasource.url=jdbc:postgresql://localhost:5432/packit
+spring.datasource.url=jdbc:postgresql://localhost:5432/packit?stringtype=unspecified
```
hillalex, line 1: Adding `stringtype=unspecified` is required for inserting json strings into the db - otherwise everything gets sent as a `varchar` which fails when it hits the db.

api/app/src/test/kotlin/packit/integration/repository/PacketRepositoryTest.kt
```diff
@@ -1,29 +1,24 @@
-package packit.unit.repository
+package packit.integration.repository
```
hillalex, line 1: I deleted the repositorry unit tests which just duplicated the tests in the integration tests but with a fake db.

scripts/run-dependencies
```diff
@@ -9,12 +9,6 @@ HERE=$(realpath "$(dirname $0)")
 "$HERE"/../db/scripts/build
 "$HERE"/../db/scripts/run
 
-# Populate db with dummy data
```
hillalex, line 12: no longer needed :tada:  when the app runs, the scheduler immediately imports data from `outpack_server`.

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
```diff
@@ -10,4 +10,7 @@ interface PacketRepository : JpaRepository<Packet, String>
...
+    fun findMostRecent(): Packet?
```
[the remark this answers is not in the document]
hillalex: Does changing the code to this work for you? When I try this I get a SQL syntax error, see comment here: https://github.com/mrc-ide/packit/pull/14#discussion_r1184871708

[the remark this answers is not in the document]
LekanAnni: I think you might have tried with `findTopOrderByTimeDesc` instead of `findTopByOrderByTimeDesc`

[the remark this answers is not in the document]
hillalex: I tried `findTopByOrderByTimeDesc` - could you try the same and see what happens?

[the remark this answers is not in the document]
LekanAnni: Works for me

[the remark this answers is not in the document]
hillalex: Ok that's super weird. Are you in tomorrow? Maybe we can compare our setups to figure out what might be different between my laptop and yours. I'll commit that change and see if it works on CI.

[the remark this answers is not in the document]
LekanAnni: Referencing your comment here `Using the JPA built in method syntax (like findTopOrderByTimeDesc) gives a SqlGrammarException and I can't for the life of me work out why. The Postgres dialect seems to be set correctly and the error is really opaque. So I gave up and did this instead.` I think you tried with `findTopOrderByTimeDesc`

api/app/src/main/resources/application.properties
```diff
@@ -1,8 +1,9 @@
...
 spring.datasource.username=packituser
 spring.datasource.password=changeme
 spring.jpa.show-sql=false
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
 spring.jpa.properties.hibernate.format_sql=true
+spring.jpa.open-in-view=false
```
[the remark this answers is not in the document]
hillalex: https://www.baeldung.com/spring-open-session-in-view#1-spring-boot

EmmaLRussell (commented):
Code looks fine to me. Can't help with the JPA issue I'm afraid, 
With current code, I'm getting: 
```
org.hibernate.exception.SQLGrammarException: JDBC exception executing SQL [select p1_0.id,p1_0.display_name,p1_0.name,p1_0.parameters,p1_0.published,p1_0.time from packet p1_0 order by p1_0.time desc fetch first ? rows only]
``` 
when I run the API - that's what you were expecting?

hillalex:
> Code looks fine to me. Can't help with the JPA issue I'm afraid, With current code, I'm getting:
> 
> ```
> org.hibernate.exception.SQLGrammarException: JDBC exception executing SQL [select p1_0.id,p1_0.display_name,p1_0.name,p1_0.parameters,p1_0.published,p1_0.time from packet p1_0 order by p1_0.time desc fetch first ? rows only]
> ```
> 
> when I run the API - that's what you were expecting?

Yup, that's the issue I see too. We can revert to my previous solution unless @LekanAnni has any ideas about what might be going wrong.

api/app/src/main/resources/application-test.properties
```diff
@@ -1,7 +1,8 @@
...
+spring.datasource.url=jdbc:postgresql://localhost:5432/packit?stringtype=unspecified
+spring.datasource.username=packituser
+spring.datasource.password=changeme
 spring.jpa.show-sql=true
+spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
```
LekanAnni, line 5:
```suggestion
spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQL82Dialect
```
Recommended to have this in test config
`spring.jpa.hibernate.ddl-auto=create`

api/app/src/main/resources/application.properties
```diff
@@ -1,8 +1,9 @@
-spring.datasource.url=jdbc:postgresql://localhost:5432/packit
+spring.datasource.url=jdbc:postgresql://localhost:5432/packit?stringtype=unspecified
 spring.datasource.username=packituser
 spring.datasource.password=changeme
 spring.jpa.show-sql=false
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
```
LekanAnni, line 5:
```suggestion
spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQL82Dialect
```

LekanAnni (changes requested):
I think I have cracked it. The dialect was the problem.

This can remove from ` runtimeOnly("com.h2database:h2")` gradlew dependencies now.

hillalex:
> I think I have cracked it. The dialect was the problem.
> 
> This can remove from ` runtimeOnly("com.h2database:h2")` gradlew dependencies now.
> 
> Works nicely without the @query annotations.

Thanks for looking into this! It's incredibly confusing because we're actually using Postgres 10 for the database but the `PostgreSQL10Dialect` has been removed, the `PostgreSQL95Dialect` causes the same SQL grammar error, and the Hibernate 6 migration docs say that [we should be using `PostgreSQLDialect` instead of a version specific dialect](https://stackoverflow.com/questions/74744188/what-happened-to-postgresql10dialect-in-hibernate-6-x)! But in fact as you've discovered, only the `PostgreSQL82Dialect` seems to generate the right SQL here. If it works it works I guess :woman_shrugging:

api/app/src/main/resources/application-test.properties
hillalex: Because we're no longer using an embedded db but using a db that is already initialised (the dockerised Postgres db) we don't want Hibernate to try and initialise it.

LekanAnni (approved):
LGTM 👍

PR #15, opened 2023-05-17
mrc-4232 test integration between front-end and API

hillalex:
Adds a jest test suite for testing that the back-end is serving responses as expected, and that the front-end is able to receive and parse them. This also implicitly tests that the scheduler is running (because packet data is returned as expected).


app/src/tests/integration/actions.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+        const asyncThunk = api.get<Packet[]>(PacketsMutationType.GetPackets, "/bad-url")();
+        await asyncThunk(dispatch, jest.fn(), jest.fn());
+
+        expect(dispatch.mock.calls[0][0]).toMatchObject({
+            type: "GetPackets/pending",
+            payload: undefined,
+        });
+        const result = dispatch.mock.calls[1][0];
+        expect(result["type"]).toBe("GetPackets/rejected");
+        expect(result["payload"]["error"]).toBe("Not Found");
```
hillalex, line 19: this comes from the built in Spring error format, open ticket to do custom error handling in the API

hillalex:
I cannot really explain what codecov is on about.

app/package.json
```diff
@@ -39,12 +39,14 @@
   "jest": {
     "moduleNameMapper": {
       "^axios$": "axios/dist/axios.js"
-    }
+    },
+    "coveragePathIgnorePatterns": ["/node_modules/", "tests"]
```
EmmaLRussell, line 43: Should this be `src/tests`? or is `coveragePathIgnorePatterns` sensitive to /s?

.github/workflows/frontend-test.yml
```diff
@@ -11,6 +11,9 @@ on:
   pull_request:
     branches:
       - 'main'
+env:
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
```
EmmaLRussell, line 15: What's this getting used for?

EmmaLRussell (commented):
LGTM

app/package.json
hillalex: it's a regex so this will match "tests" anywhere in the path

.github/workflows/frontend-test.yml
hillalex: `api/scripts/run` runs the dockerised api using the current branch name as a tag

PR #18, opened 2023-05-22
mrc-4255 deploy containers

hillalex:
This PR makes a couple of changes to the API docker container to support deployment. Unfortunately we can't use the built-in `:bootBuildImage` task as it doesn't support customisation in the way we need to get config into the container at deployment. So this now follows the pattern in `hint`:

* config that involves secrets/needs to be injected at deploy-time is separated out of `application.properties` into `config.properties` and we have `AppConfig` and `DbConfig` classes to pull them into the Spring app.
* use a custom Dockerfile that waits for config to be available at an `/etc` path before starting the app

### DIFF
--- a/api/scripts/run
+++ b/api/scripts/run
@@ -9,6 +9,5 @@ PACKAGE_ROOT=$(realpath $HERE/..)
 docker run --rm -d \
   --network=host \
   --name packit-api \
+  --volume $PACKAGE_ROOT/app/src/main/resources/config.properties:/etc/packit/config.properties \
   "$TAG_BRANCH"
-
-docker cp $PACKAGE_ROOT/app/src/main/resources/config.properties packit-api:/etc/packit/config.properties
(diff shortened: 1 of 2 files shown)


api/scripts/run
```diff
@@ -4,7 +4,11 @@ set -e
...
 . $HERE/common
 
+PACKAGE_ROOT=$(realpath $HERE/..)
+
 docker run --rm -d \
   --network=host \
   --name packit-api \
   "$TAG_BRANCH"
+
+docker cp $PACKAGE_ROOT/app/src/main/resources/config.properties packit-api:/etc/packit/config.properties
```
LekanAnni, line 14: Should this be mounted in docker volume §?

LekanAnni (commented):
LGTM

A minor comment

api/app/src/main/kotlin/packit/AppConfig.kt
```diff
@@ -1,12 +1,55 @@
...
+import java.net.URL
+import java.util.*
 
-@Configuration
-class AppConfig
+// prevent auto-wiring of default Properties
+class PackitProperties : Properties()
+
+@Component
+class AppConfig(private val props: PackitProperties = properties)
```
LekanAnni, line 13: I think there are missing tests as well.

LekanAnni (changes requested):

PR #19, opened 2023-05-31
Add packet page mrc 4282

LekanAnni:
This PR adds parameters widget to packet view. Implementation contains creation of `/packets/{id}` endpoint and consumed in frontend. When a user clicks packet name from packet runner page, the selected packet will be displayed on packet runner page. However, when a user clicks on packet runner from the sidebar, if a selected packetId is in the state, the selected packet will be displayed hence first packets will be displayed.


EmmaLRussell (changes requested):
> When a user clicks packet name from packet runner page, the selected packet will be displayed on packet runner page. However, when a user clicks on packet runner from the sidebar, if a selected packetId is in the state, the selected packet will be displayed hence first packets will be displayed.

So the packet details page *is* the package runner page? I don't have a link to the mockups to hand so I might be misremembering, but I expected those to be distinct. Conceptually, running a packet is different from looking at the details of a completed packet, and I think a user looking at the details page is going to be interested in the artefacts of the completed packet, which it wouldn't be appropriate to show on the runner page. Also, the parameters in the runner page will need to be editable so the user can set parameter values to use in the run, rather than read-only as they should be in the packet details page. 

Clicking on the sidebar 'Packet explorer' link doesn't seem to work from the packet runner page if you got there by clicking on a packet in the explorer - I'd expect that to return to the packet list, but it seems to think that's still the current page. 

The inline layout for the parameters looks good when the parameter names and values are terse are they are in the demo data, but may look less neat with real, typically more more verbose parameter names - but I think we can tweak that later if need be.

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
```diff
@@ -1,14 +1,56 @@
...
+    @BeforeEach
+    fun `init`()
+    {
+        packetRepository.save(packet)
+    }
+
+    @AfterEach
+    fun teardown()
+    {
+        packetRepository.delete(packet)
```
hillalex, line 33: to avoid having to do this manual teardown each time, I think go ahead and add
```
@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
@Transactional
```

to the base `IntegrationTest` class.

hillalex (changes requested):
The logic looks good, but it's not the `PacketRunner.tsx` component that should be rendering the packet details. We need a new packet details page equivalent to the report page in OrderlyWeb. I can't now find the mockups for this but I know we had some, could you add the mockup links to the YouTrack tickets? https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4044/Mockup-for-report-view-page and https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3972/Mockup-for-index-page

Here's an example of the old report page, for reference:
https://montagu.vaccineimpact.org/reports/report/VIMC-model-run-DALYs-guidance/20230519-145852-467240e6/

The other thing missing here is routing. We want the packet details page to be available at a distinct route, like 
`/packets/<packit-id>/` (the runner page will be at something like `/run`, the workflow runner at `/run-workflow` etc)

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -16,4 +18,10 @@ class PacketController(private val packetService: PacketService)
     {
         return ResponseEntity.ok(packetService.getPackets())
     }
+
+    @GetMapping("/{id}")
+    fun findPacket(@PathVariable id: String): ResponseEntity<Optional<Packet>>
+    {
+        return ResponseEntity.ok(packetService.getPacket(id))
+    }
```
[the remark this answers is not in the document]
hillalex: The packet details page will need more details than currently available from the index endpoint, so I think it's right that we have a separate endpoint. But the retrieval logic and returned model will have to change to accomodate the extra info. I was actually imagining this returning dummy data for now as needed by the front-end, then wiring up the database/import/retrieval later, as we did for the index page.

app/src/app/components/contents/packets/PacketRunner.tsx
```diff
@@ -1,9 +1,31 @@
 import React from "react";
+import {useSelector} from "react-redux";
+import {RootState} from "../../../../types";
 
 export default function PacketRunner() {
+    const {packet, packets} = useSelector((state: RootState) => state.packets);
+    const data = Object.keys(packet).length > 0 ? packet : packets[0];
```
hillalex, line 7: This should handle the case where `packets` has length 0 as well. At the moment calls like `data.id` on line 13 will throw errors if `data` is null.

LekanAnni:
Mock up now added to YT

LekanAnni:
I have made changes to the structure, the app now uses react router. 
To:do 
* Fix optional and render error nicely
* Get real data from outback server.

app/src/app/components/Routes/Router.tsx
```diff
@@ -0,0 +1,22 @@
+import {Route, Routes} from "react-router-dom";
```
EmmaLRussell, line 1: Tiny point, but the folder name `Routes` should probably be lower-cased like the others

app/src/app/components/contents/packets/PacketDetails.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    const {packet} = useSelector((state: RootState) => state.packets);
+
+    const paramLen = (): boolean => {
+        return Object.entries(packet.parameters).length > 0;
+    };
+
+    if (Object.keys(packet).length === 0) {
+        return <div>Loading...</div>;
+    }
```
EmmaLRussell, line 28: I guess at some point we'll have to deal with the cases which will never complete 'Loading':
1. `packetId` is not set in the route so we never attempt to fetch packet
2. `packetId` is not valid (dealing with the 404 response to be added in mrc-4331)

Could deal with 1. now or in another ticket.

LekanAnni:
> It would be good to get some e2e tests in soon.

I agree, ticket added https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4333/Add-e2e-tests

EmmaLRussell (approved):

PR #21, opened 2023-06-21
mrc-4331 Return error when calling endpoint

LekanAnni:
This PR removes support for wrapping response from the api in optional interface. Throws error when a packet can not be retrieved instead. 

PR also includes API error handlers, ensuring error messages are returned in json format when occurred. Errors should be returned in JSON format as the backend will only be used as API service. Should we want to display page not found in frontend, react will handle that. A ticket has been created here  https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4349/Add-not-found-page.

### DIFF
--- a/api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
+++ b/api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
@@ -6,7 +6,7 @@ import org.springframework.web.bind.annotation.ControllerAdvice
 import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.servlet.NoHandlerFoundException
 import packit.model.ErrorDetail
-import packit.model.ErrorDetail.Companion.defaultError
+import java.util.*
 
 @ControllerAdvice
 class PackitExceptionHandler
@@ -19,10 +19,12 @@ class PackitExceptionHandler
 
     @ExceptionHandler(PackitException::class)
     fun handlePackitException(
-        e: PackitException,
+        error: PackitException
     ): ResponseEntity<String>
     {
-        return ErrorDetail(e.httpStatus, e.message ?: "")
+        val resourceBundle = getBundle()
+
+        return ErrorDetail(error.httpStatus, resourceBundle.getString(error.key))
             .toResponseEntity()
     }
 
@@ -38,7 +40,11 @@ class PackitExceptionHandler
     {
         val message = originalMessage ?: ""
 
-        return ErrorDetail(status, message, defaultError)
-            .toResponseEntity()
+        return ErrorDetail(status, message).toResponseEntity()
+    }
+
+    private fun getBundle(): ResourceBundle
+    {
+        return ResourceBundle.getBundle("errorBundle", Locale("en"))
     }
 }
--- a/api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
+++ b/api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
@@ -1,0 +1,44 @@
+package packit.exceptions
+
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.servlet.NoHandlerFoundException
+import packit.model.ErrorDetail
+import packit.model.ErrorDetail.Companion.defaultError
+
+@ControllerAdvice
+class PackitExceptionHandler
+{
+    @ExceptionHandler(NoHandlerFoundException::class)
+    fun handleNoHandlerFoundException(error: Exception): Any
+    {
+        return handleErrorPage(error)
+    }
+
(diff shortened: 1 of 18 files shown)


app/src/tests/store/packets/reducers.test.tsx
```diff
@@ -72,14 +77,19 @@ describe("packetsSlice reducer", () => {
     });
 
     it("should handle fetchPacketById when rejected", async () => {
-        const error = Error("Could not parse API response");
+        const packerError = {
```
hillalex, line 80:
```suggestion
        const packetError = {
```

app/src/tests/store/packets/reducers.test.tsx
```diff
@@ -72,14 +77,19 @@ describe("packetsSlice reducer", () => {
...
+            error: {
+                detail: "Packet does not exist",
+                error: "OTHER_ERROR"
+            }
+        };
 
         const packetState = packetsReducer(
             initialPacketsState,
-            actions.fetchPacketById.rejected(error, "", "1")
+            actions.fetchPacketById.rejected(null, "", "1", packerError)
```
hillalex, line 89:
```suggestion
            actions.fetchPacketById.rejected(null, "", "1", packetError)
```

app/src/tests/store/packets/reducers.test.tsx
```diff
@@ -72,14 +77,19 @@ describe("packetsSlice reducer", () => {
...
         );
 
         expect(packetState.packet).toEqual({});
-        expect(packetState.error?.message).toBe(error.message);
+        expect(packetState.packetError).toBe(packerError);
```
hillalex, line 93:
```suggestion
        expect(packetState.packetError).toBe(packetError);
```

api/app/src/main/kotlin/packit/service/BasePacketService.kt
```diff
@@ -41,9 +41,16 @@ class BasePacketService(
...
 
-    override fun getPacket(id: String): Optional<Packet>
+    override fun getPacket(id: String): Packet
     {
-        return packetRepository.findById(id)
+        val packet = packetRepository.findById(id)
+
+        if (packet.isEmpty)
+        {
+            throw PackitException("packetDoesNotExist")
```
hillalex, line 50: httpStatus here should be 404

app/src/types.ts
```diff
@@ -33,8 +32,16 @@ export interface PacketTableProps {
 
 export interface PacketsState {
     packets: Packet[]
-    error: SerializedError | null
-    packet: Packet
+    error:  null | Error
```
hillalex, line 35: So this `error` property is specifically for errors when fetching packets and `packetError` is specifically for errors fetching a single packet by id? Maybe they should be called `fetchPacketsError` and `fetchPacketError` or similar to make it clear?

api/app/src/main/kotlin/packit/model/OutpackResponse.kt
```diff
@@ -1,9 +1,21 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper
 import java.io.Serializable
 
 data class OutpackResponse<T>(
         val status: String,
         val data: T,
         val errors: Any?
 ) : Serializable
+
+data class ErrorResponse(val error: ErrorDetail)
```
hillalex, line 13: I think it's fine to have all response models in this file but can we rename the file from `OutpackResponse` to just `Responses`?

api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
```diff
@@ -0,0 +1,50 @@
...
+        error: PackitException
+    ): ResponseEntity<String>
+    {
+        val resourceBundle = getBundle()
+
+        return ErrorDetail(error.httpStatus, resourceBundle.getString(error.key))
+            .toResponseEntity()
+    }
+
+    private fun handleErrorPage(e: Exception): ResponseEntity<String>
```
hillalex, line 31: this doesn't return a page and only gets called once, suggest we remove it altogether and just move line 33 to line 17?

api/app/src/main/kotlin/packit/model/ErrorDetail.kt
```diff
@@ -0,0 +1,50 @@
...
+    val error: String = defaultError,
+)
+{
+
+    companion object
+    {
+        const val defaultError = "OTHER_ERROR"
+    }
+
+    fun <T> toResponseEntity() = ResponseEntity
```
hillalex, line 19: At the moment at least this `T` isn't needed.

api/app/src/main/kotlin/packit/model/OutpackResponse.kt
hillalex: I think in due course we'll want to also modify the success responses to return in the above format (the same as the `OutpackResponse` format) but that can wait for a separate PR

hillalex, line 13:
```suggestion
data class PackitErrorResponse(val error: ErrorDetail)
```

hillalex (changes requested):
Looking good, a few small suggestions

PR #23, opened 2023-07-05
mrc-4281 Render packet details

LekanAnni:
This PR introduces new endpoints for retrieving packet files and enhances the functionality of the existing `packet/${id}` endpoint by fetching data from the metadata endpoint. These changes enable the display of packet files in the frontend, specifically for content types that are recognized as HTML.

However, there are a couple of improvements that would be needed for the incoming PR.
* It would be beneficial for **Outpack Server** to append file extension to the filename in headers for better clarity and usability. 
* All the files currently have the content type `application/octet` indicating that the file types are unknown. It would be valuable to include the appropriate content type for these files in the Outpack Server repository (https://github.com/mrc-ide/outpack_server).

### DIFF
--- a/api/app/src/main/kotlin/packit/service/BasePacketService.kt
+++ b/api/app/src/main/kotlin/packit/service/BasePacketService.kt
@@ -94,7 +94,7 @@ class BasePacketService(
 
         val headers = HttpHeaders().apply {
             contentType = MediaType.valueOf(contentMediaType)
-            contentDisposition = ContentDisposition.parse("$disposition; filename='$filename'")
+            contentDisposition = ContentDisposition.parse("$disposition; filename=$filename")
         }
 
         return byteArrayResource to headers
--- a/api/app/src/test/kotlin/packit/integration/IntegrationTest.kt
+++ b/api/app/src/test/kotlin/packit/integration/IntegrationTest.kt
@@ -4,6 +4,7 @@ import org.assertj.core.api.Assertions.assertThat
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.boot.test.web.client.TestRestTemplate
+import org.springframework.http.ContentDisposition
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
@@ -26,6 +27,10 @@ abstract class IntegrationTest
     {
         assertEquals(responseEntity.statusCode, HttpStatus.OK)
         assertEquals(responseEntity.headers.contentType, MediaType.TEXT_HTML)
+        assertEquals(
+            responseEntity.headers.contentDisposition,
+            ContentDisposition.parse("attachment; filename=report.html")
+        )
         assertThat(responseEntity.body).isEqualToIgnoringNewLines("<html><body><h1>TEST</h1></body></html>")
     }
 }


app/src/apiService.ts
```diff
@@ -24,7 +25,36 @@ export class ApiService implements API {
...
+    }
+
+    private handleDownloadError = (error: AxiosError) => {
+        console.log(error.response && error.response.data || error);
+        let errorMessage = {error: {detail: "Could not parse API response", error: "error"}};
+        if (error instanceof AxiosError && error.response) {
+            errorMessage = error.response.data as Error;
+        }
+        return errorMessage;
+    };
```
EmmaLRussell, line 43: Looks like this is identical to how non-download errors are handled, so you could call this method for both,

app/src/app/components/contents/packets/PacketDetails.tsx
```diff
@@ -43,7 +48,7 @@ export default function PacketDetails() {
                 {hasParameters() && <ParameterList parameters={packet.parameters}/>}
             </div>
             <div data-testid="runner-content" className="content-box">
-                <p>content</p>
+                {hasHash() && <PacketFile hash={packet.files[0].hash}/>}
```
EmmaLRussell, line 51: I don't think we can assume that the first file in a packet will be a displayable html file, can we? Orderly has the concept of a 'primary resource' or something like that, but have we decided what we're going to do for Packit? Just display the first html file we find in the list, if any?

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
```diff
@@ -41,16 +52,21 @@ class PacketControllerTest : IntegrationTest()
...
+    fun `get packet file by hash`()
+    {
+        val result = restTemplate
+            .getForEntity(
+                "/packets/file/sha256:c7b512b2d14a7caae8968830760cb95980a98e18ca2c2991b87c71529e223164",
+                String::class.java
+            )
 
-        assertSuccess(result)
+        assertFileSuccess(result)
```
EmmaLRussell, line 70: If there are known packets and files, can we test for expected content too?

app/src/app/store/packets/packets.ts
```diff
@@ -22,12 +24,19 @@ export const packetsSlice = createSlice({
...
+            .addCase(actions.fetchPacketMetadataById.fulfilled, (state, action) => {
                 state.packet = action.payload;
                 state.packetError = null;
             })
-            .addCase(actions.fetchPacketById.rejected, (state, action) => {
+            .addCase(actions.fetchPacketMetadataById.rejected, (state, action) => {
                 state.packetError = action.payload ?? null;
+            })
+            .addCase(actions.fetchFileByHash.fulfilled, (state, action) => {
+                state.fileUrl = action.payload;
```
EmmaLRussell, line 35: Isn't `fetchFileByHash` supposed to download the file content, rather than just provide the url`?

EmmaLRussell (commented):
I'm not seeing any file details on the packet page, although I can see that the metadata is being fetched:
image

From your comment, is it the case that html files should be visible in iframes in this branch, but none of the demo files have content type of html so it's currently difficult to test manually? 
Are non-html files to be handled in a separate branch, e.g. list with download buttons?

api/app/src/test/kotlin/packit/integration/exceptionss/PackitExceptionHandlerTest.kt
```diff
@@ -23,16 +23,10 @@ class PackitExceptionHandlerTest : IntegrationTest()
     }
 
     @Test
-    fun `throws exception when packet is not found`()
+    fun `throws exception when client error occurred`()
     {
-        val entity = restTemplate.getForEntity("/packets/nonsense", String::class.java)
+        val entity = restTemplate.getForEntity("/packets/metadata/nonsense", String::class.java)
```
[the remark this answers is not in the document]
LekanAnni: Outpack Server does not handle exceptions for no route case hence the client returns 500. I believe this should be added in future but fine for this POC.

app/src/apiService.ts
LekanAnni: Absolutely

app/src/app/components/contents/packets/PacketDetails.tsx
LekanAnni: Agreed, I think I mentioned this will be fixed in future ticket when we actually return html in the demo files.

api/app/src/main/kotlin/packit/service/OutpackServerClient.kt
```diff
@@ -2,29 +2,80 @@ package packit.service
...
 
+    override fun getMetadataById(id: String): PacketMetadata
+    {
+        return getEndpoint("metadata/$id/json")
+    }
+
+    override fun getFileBy(hash: String): Pair<ByteArray, HttpHeaders>
+    {
+        val url = "$baseUrl/file/$hash"
+        log.debug("Fetching {}", url)
```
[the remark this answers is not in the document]
LekanAnni: I thought that was useful, isn't it ?

api/app/src/test/kotlin/packit/integration/exceptionss/PackitExceptionHandlerTest.kt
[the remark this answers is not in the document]
EmmaLRussell: In `BasePacketService.getMetadataById` it's assuming the server will give back null data if not found and will return its 404 in that case, so for that to make sense, that needs to be the actual behaviour of the server. Would it be better for the server to return a 404 and we can just pass that through in packit? Either way, needs to be a ticket on outpack server.

api/app/src/main/kotlin/packit/service/BasePacketService.kt
```diff
@@ -1,25 +1,28 @@
...
 import java.security.MessageDigest
 import java.time.Instant
 
 interface PacketService
 {
     fun getPackets(): List<Packet>
-    fun getPacket(id: String): Packet
     fun getChecksum(): String
     fun importPackets()
+    fun getMetadataBy(id: String): PacketMetadata
```
hillalex, line 18: The corresponding methods in the `OutpackServer` class are called `getMetadataById` and `getFileByHash` - can we use the same naming convention here for consistency?

api/app/src/main/kotlin/packit/model/Metadata.kt
```diff
@@ -0,0 +1,28 @@
+package packit.model
+
+data class PacketMetadata(
```
hillalex, line 3: Can we make these model names consistent, so either just call this one `Packet`, or add `Metadata` to the others, i.e. `GitMetadata`, `FileMetadata` etc.?

app/src/app/store/packets/thunks.ts
```diff
@@ -1,18 +1,21 @@
 import {AsyncThunk} from "@reduxjs/toolkit";
-import {Packet, RejectedErrorValue} from "../../../types";
+import {Packet, PacketMetadata, RejectedErrorValue} from "../../../types";
 import {api} from "../../../apiService";
 
 export interface PacketsActions {
     fetchPackets: AsyncThunk<Packet[], void, RejectedErrorValue>;
-    fetchPacketById: AsyncThunk<Packet, string, RejectedErrorValue>;
+    fetchPacketMetadataById: AsyncThunk<PacketMetadata, string, RejectedErrorValue>;
```
hillalex, line 7: why this change of method name? I prefer the original `fetchPacketById`

app/src/app/components/contents/packets/PacketFile.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const {fileUrl} = useSelector((state: RootState) => state.packets);
+
+    useEffect(() => {
+        if (hash) {
+            dispatch(actions.fetchFileByHash(hash));
+        }
+    }, [hash]);
+
+    return (
+        <iframe className="packit-iframe" src={fileUrl}></iframe>
```
hillalex, line 23: the url here should just be `/packets/file/{hash}`

app/src/app/components/contents/packets/PacketDetails.tsx
hillalex: The logic in orderlyweb is: 
* if the first file is of a displayable type, then display it
* otherwise do not display anything

app/src/app/store/packets/thunks.ts
```diff
@@ -1,18 +1,21 @@
...
-    GetPacket = "GetPacket"
+    GetPacket = "GetPacket",
+    GetFile = "GetFile"
 }
 
 export const actions: PacketsActions = {
     fetchPackets: api.get<Packet[], void>(PacketsMutationType.GetPackets, "/packets"),
-    fetchPacketById: api.get<Packet, string>(PacketsMutationType.GetPacket, "/packets")
+    fetchPacketMetadataById: api.get<PacketMetadata, string>(PacketsMutationType.GetPacket, "/packets/metadata"),
+    fetchFileByHash: api.download<string, string>(PacketsMutationType.GetFile, "/packets/file")
```
hillalex, line 20: I don't think we will ever want to programatically fetch file contents in React

hillalex (changes requested):
Bit of confusion here about the front-end. I would suggest leaving out the inline artefact for now (the iframe) and just rendering a list of files that can be downloaded via the browser by the user (i..e a list of `<a>` tags with urls like `/packets/file/{hash}`. The React action `fetchFileByHash` can be removed as will never be needed.

api/app/config/detekt.yml
```diff
@@ -1,3 +1,6 @@
+build:
+  maxIssues: 3
```
LekanAnni, line 2: Suddenly started getting the below error, every fix I tried did not work, so I introduced a maxIssue count 

`Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end [ImportOrdering]`

hillalex (approved):
I've added a report with an html artefact to the demo folder for manual testing. It isn't working at the moment because the attachment header prompts the browser to download the file rather than display it in the iframe, but I think you're doing the right thing by just passing headers through from `outpack_server`, so I will modify the endpoint there to take a query parameter `inline` and only return an attachment header if `inline=false`. This is the approach we used in OrderlyWeb.

api/app/src/main/kotlin/packit/service/BasePacketService.kt
```diff
@@ -74,4 +65,31 @@ class BasePacketService(
...
+            throw PackitException("doesNotExist", HttpStatus.NOT_FOUND)
+        }
+
+        val inputStream = response.first.toString().byteInputStream()
+
+        val inputStreamResource = InputStreamResource(inputStream)
+
+        val headers = HttpHeaders().apply {
+            contentType = response.second.contentType
+            contentDisposition = response.second.contentDisposition
```
hillalex, line 90: Actually maybe it _is_ appropriate for Packit to be the one controlling this. Can this endpoint accept a query parameter `inline` and if present, don't add this `contentDisposition` header? It would also be ideal if this endpoint accepted a filename (the property `path` from the file metadata) so that it can use that as the attachment header filename when present...

LekanAnni:
Html files can display now but the content is not being read. I have created a ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4403/Make-file-render-in-iframe to fix that here

api/app/src/main/kotlin/packit/service/BasePacketService.kt
```diff
@@ -74,4 +67,33 @@ class BasePacketService(
...
+        val response = outpackServerClient.getFileByHash(hash)
+
+        if (response?.first == null || response.first.isEmpty())
+        {
+            throw PackitException("doesNotExist", HttpStatus.NOT_FOUND)
+        }
+
+        val inputStream = response.first.toString().byteInputStream()
+
+        val inputStreamResource = InputStreamResource(inputStream)
```
hillalex, line 88: This conversion is causing the problem with rendering. Just change the return type of this method to `ByteArrayResource` and then here do:

```suggestion
        val byteArrayResource = ByteArrayResource(response.first)
```

LekanAnni:
> Html files can now display data on the iframe but the content is not being render properly. I have created a ticket to look into it https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4403/Make-file-render-in-iframe

This is obsolete, ticket is now deleted.

EmmaLRussell (approved):

PR #24, opened 2023-07-06
mrc-4372 display downloadable files 

LekanAnni:
PR displays downloadable files. 

Step to test
* From packet explorer, click on a packet
* then additional sidebar items will appear, click on downloads.

Each download heading, where does the data come from ?

Endpoint returns below sample.
```
files": [
            {
                "hash": "sha256:4bda5343742c17873fcc68c6172ff4e426b1cbd6b2aacad4eec36664b928756e",
                "path": "data.rds",
                "size": 97
            },
            {
                "hash": "sha256:501aa01ab4176a4afd517a8c132429b12bc2ad3eb3430487eec0c5453305eef5",
                "path": "orderly.R",
                "size": 103
            }
        ]
```

shouldn't this be a nested array where a packet contains download heading and array of files as seen below in the old orderly?

Screenshot 2023-07-14 at 13 29 59

I am not sure we are returning all the list of files but this is what the endpoint provided. Kindly pre-review to give pointers.


EmmaLRussell (commented):
I don't think the download is working - the file contents aren't R or rds files, and the filenames aren't as expected. For example, when I download "orderly.R" for the parameters packet, I get a file with a long SHA name, with contents "[B@6b3008a0" - I'd expect to see the R file from demos/orderly/src/parameters.

api/app/config/detekt.yml
```diff
@@ -1,3 +1,6 @@
+build:
```
hillalex, line 1: can we get rid of this and just fix the issues please? I fixed them in a previous commit, have they crept back in? or, if we don't care about lexicographic ordering of imports, just disable that rule altogether

LekanAnni: This is actually an old PR. I just updated with master.

EmmaLRussell (approved):
Works for me!

app/src/app/components/contents/download/Download.tsx
```diff
@@ -0,0 +1,60 @@
...
+                <span className="p-2 pb-0 h1">{packet.custom?.orderly.description.display || packet.name}</span>
+                <span className="p-2 pt-0 small">{packet.id}</span>
+            </div>
+            <ul className="list-unstyled">
+                {packet.files.map((data, key) =>
+                    (
+                        <li key={key} className="pb-2">
+                            <div className="card custom-card">
+                                <div className="card-header">
+                                    Outputs for {data.path} model
```
hillalex, line 44:
```suggestion
                        Download {data.path} 
```

app/src/helpers.ts
```diff
@@ -0,0 +1,11 @@
+export function bytesToSize(bytes: number): string {
+    const units = ["byte", "kilobyte", "megabyte", "gigabyte", "terabyte"];
```
hillalex, line 2:
```suggestion
    const units = ["bytes", "kilobytes", "megabytes", "gigabytes", "terabytes"];
```

PR #25, opened 2023-07-14
mrc-4396 - display metadata

LekanAnni:
This PR displays sample metadata. I had like to understand what needs to be displayed here. Dependencies have not been looked into so I have no idea what is expected.


EmmaLRussell (commented):
So this is what I can see for the parameters report:
image (3)

I think this looks good for the metadata we have at the moment. Dependencies will definitely be a whole separate feature. Couple of minor comments:
- something's gone wrong with the start date I assume (1970!)
- It would be good to see some indication of how long the packet took to run (we do have the end time as well I think)
- Wide gap between git branch and commit looks a bit odd
- Might be more readable to use the shortened git SHA?

hillalex:
> * something's gone wrong with the start date I assume (1970!)

Looks good overall, the time that comes back from `outpack_server` is seconds since epoch.

app/src/helpers.ts
```diff
@@ -13,17 +13,30 @@ export function bytesToSize(bytes: number): string {
...
+    const remainingMinutes = minutes % 60;
 
     let formattedTime = "";
-    if (minutes > 0) {
-        formattedTime += `${minutes} minute${minutes > 1 ? "s" : ""}`;
+    if (hours > 0) {
+        formattedTime += `${hours} hour${hours > 1 ? "s" : ""}`;
+    }
+    if (remainingMinutes > 0) {
+        formattedTime += ` ${remainingMinutes} minute${remainingMinutes > 1 ? "s" : ""}`;
```
EmmaLRussell: It looks like not all these code paths are being tested, and that's not triggering codecov - is that because `helpers.ts` isn't getting picked up by codecov because it's in the root `src` folder, rather than `src/app`?

EmmaLRussell (commented):
I'm not seeing Elapsed Time for any of the demo packets, even though the end and start times are different. Is this working for you?

image

EmmaLRussell:
Coverage issue seems to be fixed now, thanks, but still not seeing any elapsed times in display (see initial comment of last review).

EmmaLRussell (approved):
Great, thanks!

PR #30, opened 2023-10-11
Authenticate via Git API

LekanAnni:
This PR allows user to authenticate with GitHub personal token. 

Config includes orgs that are allowed.

paths to login `/auth/login/github`


EmmaLRussell (commented):
Thanks for the chat about this ticket @LekanAnni 
My notes from that:

For Lekan to do before he goes:
- add some documentation to the README, how to configure and use the endpoints.

For us to take on, probably in separate tickets:
- include support for specifying allowed github team as well as org (replicate logic from OrderlyWeb) and note that this should apply to web login as well as API
- when adding above support, consider using [org.kohsuke.github](https://github-api.kohsuke.org/apidocs/org/kohsuke/github/package-summary.html) to replace GithubClient (it works well in OrderlyWeb and is being maintained, but balance that with risks of 3rd party dependency)
- note that the request and response here is different to that in OW (which accepts Github PAT in header `Authorization: token [Github PAT]` rather than in request body, and returns JSON with `access_token`, `expires_in` and `token_type` rather than just the token as here. This should be fine as users will be accessing the API via the R package, but we'll need to modify that for the change in interface.
- I think the `login/github` endpoint should be something like `login/api` since we will be adding montagu access, and the auth provider will be defined in the app config rather than something the user should be deciding on. 
- add a full integration test

EmmaLRussell:
As suggested above, I've added the following tickets to extend the functionality here:

-   use [org.kohsuke.github](https://github-api.kohsuke.org/apidocs/org/kohsuke/github/package-summary.html) to replace GithubClient: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4682

- include support for specifying allowed github team as well as org (replicate logic from OrderlyWeb) and note that this should apply to web login as well as API: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4683

-     change the login/github endpoint to login/api since we will be adding montagu access, and the auth provider will be defined in the app config rather than something the user should be deciding on: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4684/

-     add a full integration test: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4685

EmmaLRussell (approved):

PR #31, opened 2023-10-12
Mrc-4638 - Validate and extract user details from Token

LekanAnni:
- Extracts email from jwt
- move apiService token method to helpers


app/src/helpers.ts
```diff
@@ -45,9 +47,20 @@ export const getElapsedTime = (time: TimeMetadata) => {
...
     return !!user.token;
 };
+
+export const useBearerToken = (
+    injectedStore: any,
+    config: InternalAxiosRequestConfig<any>)
+    : InternalAxiosRequestConfig<any> => {
+    if (validateToken(getCurrentUser() || injectedStore.getState().login.user)) {
+        config.headers.authorization = `Bearer ${getCurrentUser().token
+        || injectedStore.getState().login.user.token}`;
```
LekanAnni, line 59: I am unable to test this line

app/src/app/components/routes/ProtectedRoute.tsx
```diff
@@ -18,15 +19,29 @@ export default function ProtectedRoute() {
...
-                return;
             }
 
-            if (user && user.token) {
-                dispatch(saveUser(user));
-            }
+            const decodedToken = decodeToken(token) as PackitDecodedToken;
+
+            const newUser = (decodedToken && decodedToken.email)
+                ? {...user, email: decodedToken.email}
```
LekanAnni, line 41: Could not test this line, not sure why the local storage does not work for object in test.

app/src/app/components/routes/ProtectedRoute.tsx
```diff
@@ -18,15 +19,29 @@ export default function ProtectedRoute() {
     useEffect(() => {
         const user = getCurrentUser();
 
-        if (authConfig.enableAuth) {
-            if (!isAuthenticated && !user?.token) {
+        if (!authConfig.enableAuth) {
+            return;
+        }
+
+        if (!isAuthenticated && !user?.token) {
```
EmmaLRussell, line 26: What case does this cover, where we've authenticated but there isn't a user token? Could you add a comment to describe please?

app/src/app/components/routes/ProtectedRoute.tsx
```diff
@@ -18,15 +19,29 @@ export default function ProtectedRoute() {
...
+            navigate("/login");
```
EmmaLRussell, line 27: Just for the sake of DRY, would be nice to pull out this line as it gets done in two places

EmmaLRussell (commented):
This works nicely for basic auth, but I see "null" in the user email menu item when I log in with github, 

Got a few questions and comments, obviously @LekanAnni you don't need to address all of these, but would be nice if you could shed some light on the questions! :)

EmmaLRussell:
I'm going to return to this ticket after doing https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4682/Use-kohsuke-github-package
It looks like the email is not being returned in the user info from github, even after adding user:email to requested scope - looking at OW code suggests this needs to be requested in a separate request, and I think we should delay doing this until we're using the github client package,

EmmaLRussell:
Closing this  as user details have been extracted from token in another branch

PR #36, opened 2023-11-29
Mrc-4720 Backend For all reports page + misc

absternator:
This PR contains the BE work needed for the all reports page. Note: some FE stuff has been added just to ensure the BE things work as expected. Thus may be best to not review FE and review it once i make new ticket for all FE changes.
 
Below are things done: 

- Schema index add 
-  add endpoint + rest for all reports 
- FE: add Home page to test BE changes (this will be updated in next PR so can probs skip over)
- Added more demo data for testing purposes (run outpack on same packet groups)


app/src/apiService.ts
absternator, line 1: Only formatting can skip

app/src/types.ts
```diff
@@ -1,123 +1,140 @@
...
+  workflowRunner,
+  projectDoc
 }
 
 export interface RejectedErrorValue {
-    rejectValue: Error
+  rejectValue: Error;
 }
-
+// TODO: depricate below
```
absternator, line 15: here

db/schema/schema.sql
```diff
@@ -31,45 +31,47 @@ CREATE TABLE "permission"
...
 CREATE TABLE "settings"
 (
-    "auth_allow_guest"  BOOLEAN
+    "auth_allow_guest" BOOLEAN
 );
 
 CREATE TABLE "packet"
 (
-    "id"  TEXT NOT NULL,
+    "id" TEXT PRIMARY KEY NOT NULL,
```
absternator, line 69: here and below only

EmmaLRussell (commented):
LGTM, we've talked about most of the back end stuff already. As suggested, I haven't reviewed the front end bits. 

Some suggestions about naming, and I'm not convinced we need an interface/getters for the DTO - we should be consistent with Packet, which is a data class. Or were you forced into doing it that way by JPA?

It doesn't look like the build actions have been running for this branch - any idea why?

Could you describe the changes you made in orderly? This was running existing reports so you had multiple packets per report?

I don't think you should need to commit `demos/orderly/draft` changes - those are ephemeral in orderly, so should probably roll those back.

EmmaLRussell:
Oh, of course the github worflows aren't running because they're configured to only run for PRs into main - just ran into this one myself.. Can temporarily update to include PRs to mrc-4716

api/app/src/main/kotlin/packit/model/PacketIdCountsDTO.kt
```diff
@@ -0,0 +1,13 @@
+package packit.model
+
+// Projection class for PacketRepository.findIdCountDataByName
+interface IPacketIdCountsDTO
```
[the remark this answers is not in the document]
absternator: also im running all my tests locallly and making sure they pass.. maybe we can just rely on build actions running on the main feature branch this will be merged into

EmmaLRussell (commented):
Thanks, `PacketGroupSummary` definitely feels easier to reason about. 

Had a chat with Rich about packet ids - they should always be chronologically ordered, which will make life a bit easier.

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
```diff
@@ -11,4 +14,26 @@ interface PacketRepository : JpaRepository<Packet, String>
     @Query("select p.id from Packet p order by p.id asc")
     fun findAllIds(): List<String>
     fun findTopByOrderByTimeDesc(): Packet?
+    @Query(
+        value = "SELECT " +
```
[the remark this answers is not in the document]
absternator: as discussed will leave query as is and make a seperate ticket to fetch start & end times that will be used for sorting

.github/workflows/backend-test-and-build.yml
```diff
@@ -2,15 +2,17 @@ name: backend-test
...
       - completed
   push:
     branches:
-      - 'main'
+      - "main"
   pull_request:
     branches:
-      - 'main'
+      - "main"
+      # TODO: remove wone 4716 is merged
```
EmmaLRussell, line 14: wone? Is that like a mashup of when and once? 😆

EmmaLRussell (approved):
Codecov complaining but I guess the front end changes are tested in the next branch.

PR #37, opened 2023-11-30
mrc-4683 restrict auth to GitHub team

EmmaLRussell:
This branch restricts Github user access to members of a configured team, if provided. If not provided, membership of the configured org alone is required (as previously). In this branch, we remove the ability to define multiple allowed orgs, as this is not required. 

Manual testing is as in https://github.com/mrc-ide/packit/pull/35 - set the value of `auth.githubAPITeam` in `config.properties` to the name of a team you are or are not a member of.


api/app/src/main/kotlin/packit/clients/GithubUserClient.kt
```diff
@@ -38,14 +37,25 @@ class GithubUserClient(private val config: AppConfig, private val githubBuilder:
         return UserPrincipal.create(user, mutableMapOf())
     }
 
-    fun checkGithubOrgMembership()
+    fun checkGithubMembership()
```
absternator, line 40: should we probably be  returning something here? like true or something? seems a bit weird to let method run and relying on throwing error?

api/app/src/main/kotlin/packit/clients/GithubUserClient.kt
```diff
@@ -38,14 +37,25 @@ class GithubUserClient(private val config: AppConfig, private val githubBuilder:
...
     {
         checkAuthenticated()
 
-        val userOrgs = ghUser!!.allOrganizations.map {org -> org.login}
-        val inAllowedOrg = allowedOrgs.any {allowed -> userOrgs.contains(allowed)}
+        val userOrg = ghUser!!.allOrganizations.firstOrNull { org -> org.login == config.authGithubAPIOrg }
```
absternator, line 44: also `!!` feels kinda unsafe... is it better to null coalsece? (i.e `ghUser?.allOrganisations`)

api/app/src/main/kotlin/packit/clients/GithubUserClient.kt
```diff
@@ -38,14 +37,25 @@ class GithubUserClient(private val config: AppConfig, private val githubBuilder:
...
+        var userAllowed = userOrg != null
 
-        if (!inAllowedOrg)
+        val allowedTeam = config.authGithubAPITeam
+        if (userAllowed && !allowedTeam.isEmpty())
```
absternator, line 48:
```suggestion
        if (userAllowed && !allowedTeam.isNotEmpty())
```

absternator (commented):
LGTM and works like a charm :heart_eyes:  
Just a general Question for me to understand on how this is going to work? maybe a quick chat... because the app would define these teams so unsure how that links with users

api/app/src/main/kotlin/packit/clients/GithubUserClient.kt
```diff
@@ -38,14 +37,25 @@ class GithubUserClient(private val config: AppConfig, private val githubBuilder:
...
+        {
+            // We've confirmed user is in org, and required team is not empty, so we need to check team membership too
+            val team = userOrg!!.teams[allowedTeam] ?: throw PackitAuthenticationException(
+                "githubConfigTeamNotInOrg",
+                HttpStatus.UNAUTHORIZED
+            )
+            userAllowed = ghUser!!.isMemberOf(team)
+        }
+
+        if (!userAllowed)
```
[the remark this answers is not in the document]
EmmaLRussell: It's potentially setting userAllowed to false on line 55 as well though, and I this way I can only throw the exception in one place.

api/app/src/main/kotlin/packit/clients/GithubUserClient.kt
EmmaLRussell: That's kind of the idea though, since this is being called from both the OAuth2UserService (browser login) and GithubAPILoginService (API login) - they're both delegating the throwing of the exception to the user client if there's a problem with the user authentication. I'll add an annotation to the method though, to make it more explicit what the usage is.

absternator (approved):

api/app/src/main/kotlin/packit/clients/GithubUserClient.kt
```diff
@@ -38,14 +37,26 @@ class GithubUserClient(private val config: AppConfig, private val githubBuilder:
...
```
M-Kusumgar, line 59: this logic feels very weird to me to follow, like i get what it is trying to do but the `userAllowed` checks are weird, it goes into a condition that relies on `userAllowed` and then it potentially changes the value of `userAllowed` and we do another if statement on it. Here's my go at writing this function, feel free to not do it this way!
```kotlin
checkAuthenticated()

val userOrg = ghUser!!.allOrganizations.firstOrNull { org -> org.login == config.authGithubAPIOrg }
val allowedTeam = config.authGithubAPITeam
val orgContainsTeam = if (allowedTeam.isEmpty()) true else userOrg?.teams?.containsKey(allowedTeam)

if (userOrg == null || !ghUser!!.isMemberOf(userOrg.teams[allowedTeam])) {
    throw PackitAuthenticationException("githubUserRestrictedAccess", HttpStatus.UNAUTHORIZED)
}

if (orgContainsTeam == false) {
    throw PackitAuthenticationException(
        "githubConfigTeamNotInOrg",
        HttpStatus.UNAUTHORIZED
    )
}
```

M-Kusumgar (approved):
Looks good to me, minor comment about structure of code in one of the functions but happy either way, worked for me!

api/app/src/main/kotlin/packit/clients/GithubUserClient.kt
EmmaLRussell: I don't think that will quite work, because it's purporting to check if user `isMemberOf` team which might be null because  the team isn't in the org (so I don't think it would compile) - we could put an extra check for it to fail that `if` if `!orgContainsTeam`, so that it goes on to raise the `gitHubConfigTeamNotInOrg` error - but I think that would be even more confusing, particularly since `orgContainsTeam` is true if there isn't a configured team... 

I feel like the pattern of keeping a running 'ok' flag and doing successive checks which set it to false if there's anything wrong is fairly standard. The logic as is sort of makes sense as it's doing the team check only once it knows it can rely on the user org. But yeah it is a bit weird. I'll add some comments and change `userAllowed` to `userOK`, as I feel like that sounds a bit less final..?!

PR #39, opened 2023-12-04
mrc-4741 Remove basic auth and simplify user classes

EmmaLRussell:
This ticket simplifies the auth implementation, removing the temporary basic auth functionality, and simplifying user classes, preparatory to adding Montagu auth in a future branch. 

User classes previously had integer id and email fields - both of these are being removed and replaced with username (unique for the provider - this will actualy be an email address in the case of Montagu login, but will be user id for Github login) and optional display name. 

In a future ticket, where we add Montagu auth, I'll refactor the config slightly, to replace `auth.enableGithubLogin` with `auth.provider` (either github or montagu). 

Codecov is moaning about global coverage decrease, but the patch is fully covered.


api/app/src/main/kotlin/packit/AppConfig.kt
```diff
@@ -16,9 +16,8 @@ class AppConfig(private val props: PackitProperties = properties)
     val dbUrl: String = propString("db.url")
     val dbUser: String = propString("db.user")
     val dbPassword: String = propString("db.password")
-    val authBasicSecret: String = propString("auth.basic.secret")
+    val authJWTSecret: String = propString("auth.jwt.secret")
```
EmmaLRussell, line 19: This relates to generating the JWT token so this seemed like a less confusing name...

app/src/tests/integration/IntegrationTest.ts
```diff
@@ -1,21 +1,5 @@
...
-        async (prop, thunkAPI) => api(store)
-            .postAndReturn("/auth/login", prop, thunkAPI));
-    const fetchTokenAction = fetchTokenThunk(testUser);
-    await fetchTokenAction(dispatch, jest.fn(), jest.fn());
-    const result = dispatch.mock.calls[1][0];
-    expect(result["type"]).toBe("GetToken/fulfilled");
-
-    return result["payload"];
+// Long-lived JWT token for test backend configuration which we can use to authenticate in integration tests
+export const testUser = {
```
EmmaLRussell, line 2: The integration tests previously used the now-retired basic auth login to access protected endpoints - we don't yet have a full integration test solution for the Github login (and doing a full login might not be appropriate for testing non-auth functionality), so here I've just replaced that with a long-lived (20 year) JWT token generated from the token issuer code.

absternator (approved):
LGMT nice work with the cleanups love it

api/app/src/main/kotlin/packit/security/profile/TokenToPrincipalConverter.kt
```diff
@@ -17,10 +17,9 @@ class TokenToPrincipalConverter : TokenToPrincipal
     override fun convert(jwt: DecodedJWT): UserPrincipal
     {
         return UserPrincipal(
-            jwt.getClaim("email").asString(),
-            "",
+            jwt.getClaim("userName").asString(),
+            jwt.getClaim("displayName").let{ if (it.isNull) null else it.asString() },
```
M-Kusumgar, line 21: i think
```suggestion
            jwt.getClaim("displayName")?.toString(),
```

M-Kusumgar (approved):
Looks good! love it when we remove code XD

PR #40, opened 2023-12-07
Mrc 4721 Packet Group table page

absternator:
The following PR is for the packet group page, which shows a table of all packets in the packet version. The work is a simple table and a TODO is left when requirements are fleshed out to make the table more feature-rich..... The following are part of PR: 
- Endpoint for packets by packet name 
- Packet group page with table 
- unit tests 
- some extra orderly demo data 
image


app/src/app/components/contents/common/DataTable.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <div className="rounded-md border">
+      <Table>
+        <TableHeader>
+          {table.getHeaderGroups().map((headerGroup) => (
+            <TableRow key={headerGroup.id}>
+              {headerGroup.headers.map((header) => (
+                <TableHead key={header.id}>
+                  {header.isPlaceholder ? null : flexRender(header.column.columnDef.header, header.getContext())}
```
EmmaLRussell, line 21: What does `isPlaceholder` mean here? Why would a column be a placeholder.

app/src/app/components/contents/PacketGroup/packetColumns.tsx
```diff
@@ -0,0 +1,55 @@
+import { ColumnDef, createColumnHelper } from "@tanstack/react-table";
+import { Link } from "react-router-dom";
+import { Packet } from "../../../../types";
+import { CheckSquare, X } from "lucide-react";
+import { Separator } from "../../Base/Separator";
+
+const columnHelper = createColumnHelper<Packet>();
+
+export const packetColumns = [
```
EmmaLRussell, line 9: So this thing gets plugged into the tanstack table, via `DataTable` and defines how to render values into table cells?

EmmaLRussell (commented):
Looks good! I like the parameter view. 

I would change the Published icons though - the two don't match each other (tick has a box round it but cross doesn't it) and both misleadingly look like they should be clickable (the tick resembles a checkbox and the cross resembles a close button). So I think just Yes/No or Published/Internal text would be better. Or maybe just slightly different icon choices would work better (should probably have tooltips in that case).

M-Kusumgar (commented):
Looks good! I have a couple of comments:

I agree with emma about the icons, they should be consistent and probably nice to have some text. I wonder if you even need that column? Could just be a green pill saying published and yellow or red one saying private or something like that?

When you hover on the header of the table it highlights, I think it should only highlight when you hover on a packet row. I also think when you click anywhere on the packet row it should take you to the details about the packet , currently it only takes you to the correct page if you click in the packet column.

I have to say I am not a huge fan of the way we are displaying parameters, I think the pills look quite busy and crowded. Perhaps a bit more padding and bigger margins? I definitely think the "|" should be replaced with a ":" or "=".

app/src/app/components/contents/common/DataTable.tsx
absternator: its a tanstack table thing where there can be a placeholer header

app/src/app/components/contents/PacketGroup/packetColumns.tsx
absternator: yup

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const PacketGroup = () => {
+  const { packetName } = useParams();
+
+  return (
+    <div className="flex justify-center">
+      <div className="h-full flex-1 flex-col space-y-8 p-8 max-w-7xl">
+        <div className="flex items-center justify-between space-y-2">
+          <div>
+            <h2 className="text-2xl font-bold tracking-tight">{capitalizeFirstLetter(packetName ?? "")}</h2>
+            <p className="text-muted-foreground">Here&apos;s a list of all packets for the {packetName} packet group</p>
```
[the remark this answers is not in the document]
absternator: yup have upated and just use what is returned from api

absternator:
The table has be rearranged as suggested... Published column removed.. Also parameters have been mae more spacious and used : as seperator.Also no hover highlight on header of table

app/src/app/components/contents/PacketGroup/PacketTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const PAGE_SIZE = 50;
+
+  const { data, isLoading, error } = useSWR<PageablePackets>(
+    `${appConfig.apiUrl()}/packets/${packetName}?pageNumber=${pageNumber}&pageSize=${PAGE_SIZE}`,
+    (url: string) => fetcher({ url, authRequired: true })
+  );
+
+  if (error) return <ErrorComponent message="Error fetching packets" error={error} />;
+
+  if (isLoading)
```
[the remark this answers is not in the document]
absternator: I dont think first thing is feasibile. But agree with skeleton.. have changed to 2.. 1 looked odd when testing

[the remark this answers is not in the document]
absternator: also the other comment  "I also think when you click anywhere on the packet row it should take you to the details about the packet , currently it only takes you to the correct page if you click in the packet column."... I personally disagree and provies weird UX experience so have not done that

M-Kusumgar (approved):
Looks good to me! i like the ":" and the colours for the published and internal! One minor comment

EmmaLRussell (approved):

PR #41, opened 2023-12-07
mrc-4758 Fix download file when authenticated

EmmaLRussell:
This branch fixes the download file links when auth is enabled and user is logged in - previously this was failing because we were rendering plain `<a>` links to the downloads, which of course to do not pass the required `Authorization` header. 

I've added a `download` lib method which passes the header if required, and spoofs a link click to the actual download. 
I've replaced the previous links with a `DownloadButton` component - and made this look more like a button than a link.
This also shows an error message if there is an error downloading the file. 

NB this PR is merging to the theme update epic branch in order to pick up the new `lib` structure and msw support.


.github/workflows/backend-test-and-build.yml
```diff
@@ -7,11 +7,11 @@ on:
       - completed
```
absternator, line 7: Might need to do another pull from 4716

app/src/app/components/contents/download/DownloadButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+import {bytesToSize} from "../../../../helpers";
+import {useSelector} from "react-redux";
+
+interface DownloadButtonProps {
+    file: FileMetadata
+}
+
+export default function DownloadButton({ file }: DownloadButtonProps) {
+    const [error, setError] = useState("");
+    const { isAuthenticated } = useSelector((state: RootState) => state.login );
```
absternator, line 16: pretty sure you dont have check authenicated here.... as you have to be authenticated to see this page

app/src/lib/download.ts
```diff
@@ -0,0 +1,25 @@
+import {getAuthHeader} from "./auth/getAuthHeader";
+
+export const download = async (url: string, filename: string, authRequired = true) => {
+    const headers = authRequired ? getAuthHeader(): undefined;
```
absternator, line 4: can use existing fetcher to achieve all this data fetching pretty sure....

app/src/lib/download.ts
```diff
@@ -0,0 +1,25 @@
...
+        headers,
+    });
+
+    if (!res.ok) {
+        const json = await res.json();
+        const msg = json.error?.detail ? `Error: ${json.error.detail}` : `Error downloading ${filename}`;
+        throw new Error(msg);
+    }
+
+    const blob = await res.blob();
```
absternator, line 16: try catch around this

app/src/lib/download.ts
```diff
@@ -0,0 +1,25 @@
...
+    const fileUrl = URL.createObjectURL(blob);
```
absternator, line 17: sorry just my ignorance but is this the only way to do this? seems really hacky for some reason

app/src/tests/components/contents/download/DownloadButton.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+import userEvent from "@testing-library/user-event";
+
+describe("DownloadButton", () => {
+    const file = {
+        path: "test.txt",
+        size: 1024,
+        hash: "fakeHash"
+    };
+
+    const getStore = (props: Partial<LoginState> = {isAuthenticated: true}) => {
```
absternator, line 22: as above dont think we neeed to test if user is authenticated or not

app/src/tests/components/contents/download/DownloadButton.test.tsx
```diff
@@ -0,0 +1,81 @@
+const mockDownload = jest.fn();
```
absternator, line 1: maybe best to move under imports

app/src/tests/lib/auth/getAuthHeader.test.ts
```diff
@@ -0,0 +1,14 @@
+const mockGetBearerToken = jest.fn();
```
absternator, line 1: below imports

app/src/tests/lib/auth/getAuthHeader.test.ts
```diff
@@ -0,0 +1,14 @@
...
+jest.mock("../../../lib/auth/getBearerToken", () => ({
+    getBearerToken: () => mockGetBearerToken()
```
absternator, line 3:
```suggestion
    getBearerToken: jest.fn(() => "1234")
```
Can avoid creating another var

app/src/tests/lib/download.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import {download} from "../../lib/download";
+import {mockFileBlob} from "../mocks";
+import {server} from "../../msw/server";
+import {rest} from "msw";
+
+const url = `${downloadFileUri}?filename=test.txt`;
+
+describe("download", () => {
+    const fakeObjectUrl = "fakeObjectUrl";
+    const realFetch = window.fetch;
```
absternator, line 16: shouldnt need to mock fetch... can use msw handler to intercept request.. msw purpose is so we can avoid this :smile:

absternator (commented):
LGTM the code is sweet as just some minor comments... Havent tested yet il wait till parent branch is merged in as lots of changes

app/src/tests/components/contents/download/DownloadButton.test.tsx
EmmaLRussell: I think it has to go before the import so that the real download doesn't get pulled in by the imported method - will confirm this though,

app/src/tests/components/contents/download/DownloadButton.test.tsx
EmmaLRussell: This is because we might be running an instance with auth switched off.

app/src/tests/lib/auth/getAuthHeader.test.ts
EmmaLRussell: Ah yeah, can do here actually since we don't need to explicitly expect the mock

app/src/tests/lib/download.test.ts
EmmaLRussell: I want to spy on  it though... :mag:

EmmaLRussell: Unless msw has some spy magic I don't know about?

app/src/lib/download.ts
EmmaLRussell: It seems to be sadly... Yeah, that was my feeling too, but this seems to be the standard approach.

app/src/lib/download.ts
EmmaLRussell: Yeah, I wondered about that, but would need to modify the existing fetcher, which currently just returns JSON. I figured it wouldn't actually share very much code other than appending the auth header (since this has to deal with a blob on success and json on failure). I could put in the capacity to the current fetcher to allow blob return but I think that would overcompilicate that probably..

.github/workflows/backend-test-and-build.yml
EmmaLRussell: Cheers, yep will repoint the PR once 4716 merged to main,

app/src/tests/components/contents/download/DownloadButton.test.tsx
EmmaLRussell: Actually, we could just let the getAuthHeader method return nothing if there's no current user, that might be simpler as the clients of the function wouldn't need to know, but might be confusing since the header returned might be empty,

app/src/lib/download.ts
absternator: ohh didnt realize it does a blob on the res directly.... yup good

app/src/tests/components/contents/download/DownloadButton.test.tsx
absternator: it wont.. when you use jest.mock it auto gets hoiste to top of file :smile:

app/src/tests/components/contents/download/DownloadButton.test.tsx
absternator: ohh yeah that auth instance off scenario

app/src/tests/lib/download.test.ts
absternator: what exactly are you spying on? is the direct spyon fetch needed for the test? looks like it line 48 is the assertion... unsure if that is needed since you are already checking lots of assertions afterwards?

app/src/tests/components/contents/download/DownloadButton.test.tsx
EmmaLRussell: Yeah, that's what I thought it should do, but I'm sure it's been a problem in the past and only setting up mocks pre-imports fixed it. Will dig into that though, you're right, shouldn't need to do that.

app/src/tests/lib/download.test.ts
EmmaLRussell: I think I wanted to use the spy to check the header had been sent correctly, and then I needed to restore the real fetch for subsequent fetches to invoke msw. But actually I could probably configure msw to only kick in when the correct header is provided, to return the mock blob, and that would confirm that the header had been sent.. I'll give that a try.

app/src/tests/components/contents/download/DownloadButton.test.tsx
EmmaLRussell: It can be downloaded if there's no header and auth is switched off on the server, but you'll get a 401 if auth is switched on and no valid header is provided. But in terms of front end testing, we can just check that we provide the token if we have one.

app/src/tests/components/contents/download/DownloadButton.test.tsx
EmmaLRussell: It was fine! Must have been having other issues when I did this before! 😆

absternator (approved):

PR #42, opened 2023-12-11
Mrc-4722 Packet Pages (Exc Download + Changelogs)

absternator:
This PR Is for the specific packet pages. The following  is included 
- Packet Page layout 
- Metdata + Packet details page
- Unit tests
- Removal of redundant unit tests

Note: the download will be fixed after emma has done the download page

image

image


app/src/app/components/contents/metadata/MetadataListItem.tsx
```diff
@@ -0,0 +1,12 @@
+interface MetadataItemProps {
+  label: string;
+  value: string;
+}
+export const MetadataListItem = ({ label, value }: MetadataItemProps) => {
+  return (
+    <li>
+      <span className="font-semibold mr-2">{label}:</span>
+      <span className="text-muted-foreground">{value}</span>
```
EmmaLRussell, line 9: Feels like the muted and non-muted text are kind of the wrong way round here, like it should be emphasising the values rather than the titles.

app/src/app/components/contents/packets/PacketParameters.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <div>
+      <h4 className="scroll-m-20 text-xl font-semibold tracking-tight">Parameters</h4>
+      <div className="flex flex-wrap gap-1">
+        {Object.keys(parameters)?.length === 0 ? (
+          <div className="italic text-sm">None</div>
+        ) : (
+          Object.entries(parameters).map(([key, val]) => (
+            <div key={key} className="border p-1 rounded-md flex space-x-1 text-sm">
+              <div>{key}</div>
+              <Separator orientation="vertical" />
```
EmmaLRussell, line 17: I guess this should now be consistent with the table view, which shows colon now rather than vertical separator

app/src/types.ts
```diff
@@ -101,8 +101,8 @@ export interface Custom {
 }
 
 interface Description {
-  custom: Record<string, string>;
-  display: string;
+  custom: Record<string, string> | null;
+  display: string | null;
```
EmmaLRussell, line 105: it doesn't look to me like Description is actually being displayed anywhere any more?

EmmaLRussell (approved):
LGTM. A couple of tiny cosmetic comments, feel free to address or ignore. 

Yeah, we'll need to do some more work to get the html file showing, the ticket I've just done only covers the download buttons. Can do that in another ticket.

app/src/app/components/contents/metadata/MetadataListItem.tsx
M-Kusumgar: hmmm wait isn't this the same as the parameter values? idm switching this around but should we not change the parameter value format too if we switch this?

M-Kusumgar (approved):
Looks good to me! I like the custom hooks :) Just one minor appearance comment:

Agree with emma, the separator should be ":" since we use that in the table. However it did get me thinking about the separator. The metadata page has lots of colons, i actually think it looks better without the colons:

image

since we separate by colours anyway, maybe there isnt a need for the separators at all there. I did try this on the parameters and it looked weird so my suggestion is remove the ":" from the metadata page and keep them in the parameters? Just a suggestion though, not super fussed either way

app/src/app/components/contents/metadata/MetadataListItem.tsx
absternator: tried it that way but looked very weird.. and in my opinion makes sense to emphasize title

app/src/types.ts
absternator: yeah its not but will leave in as still part of data structure recieved

app/src/app/components/contents/packets/PacketReports.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface PacketReportsProps {
+  packet: PacketMetadata | undefined;
+}
+// TODO : Fix Auth for files ++ add ability to load multiple reports.
+export const PacketReports = ({ packet }: PacketReportsProps) => {
+  const htmlFilePath = packet ? getHtmlFilePath(packet) : null;
+
+  return (
+    <div>
+      <h4 className="scroll-m-20 text-xl font-semibold tracking-tight">Reports</h4>
```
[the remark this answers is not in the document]
absternator: i havent yet but you are right makes it a bit confusing.... maybe il leave it for this PR and when we fix the auth for this page we can find a suitable name

PR #43, opened 2023-12-13
MRC 4770 - Update Time fields

absternator:
NOTE: requires the merging of https://github.com/mrc-ide/outpack_server/pull/50 first.

This PR fixes the issue of having only import time being used in the application. Thus, start time and import time are added to the packets DB table and returned. This  requires outpack_server work to be merged as above, THe following are done in this PR: 
- Changed Packet data entity (kotlin + DB) and update FE
- Fix unit tests 
- Use start time for ordering

### DIFF
--- a/api/app/src/main/kotlin/packit/model/Packet.kt
+++ b/api/app/src/main/kotlin/packit/model/Packet.kt
@@ -1,16 +1,20 @@
 package packit.model
 
-import jakarta.persistence.*
+import jakarta.persistence.Convert
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
 import packit.helpers.JsonMapConverter
 
 @Entity
 data class Packet(
-        @Id
-        val id: String,
-        val name: String,
-        val displayName: String,
-        @Convert(converter = JsonMapConverter::class)
-        val parameters: Map<String, Any>,
-        val published: Boolean,
-        val time: Long
+    @Id
+    val id: String,
+    val name: String,
+    val displayName: String,
+    @Convert(converter = JsonMapConverter::class)
+    val parameters: Map<String, Any>,
+    val published: Boolean,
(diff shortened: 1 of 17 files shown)


r-ash (approved):
Looks good to me! Though @M-Kusumgar perhaps a better 2nd reviewer than myself

M-Kusumgar (approved):
Works for me! Nice one :)

api/app/src/main/kotlin/packit/model/Packet.kt
```diff
@@ -1,16 +1,20 @@
...
+    @Id
+    val id: String,
+    val name: String,
+    val displayName: String,
+    @Convert(converter = JsonMapConverter::class)
+    val parameters: Map<String, Any>,
+    val published: Boolean,
+    val importTime: Long,
+    val startTime: Double,
+    val endTime: Double
```
EmmaLRussell, line 19: Sorry, late to this party obviously, but is there a reason why we're keeping importTime in a different format than start and end time? Is it because they have a millisecond part? Might be good to be consistent with those, or comment on the distinction.

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
```diff
@@ -13,27 +13,27 @@ interface PacketRepository : JpaRepository<Packet, String>
 {
     @Query("select p.id from Packet p order by p.id asc")
     fun findAllIds(): List<String>
-    fun findTopByOrderByTimeDesc(): Packet?
+    fun findTopByOrderByImportTimeDesc(): Packet?
```
EmmaLRussell, line 16: Looks like this should be `findTopOrderByStartTimeDesc`

EmmaLRussell (commented):
LGTM

PR #44, opened 2023-12-14
mrc-4684 Rename api auth endpoint

EmmaLRussell:
Change the api login endpoint from `/auth/login/github` to `/auth/login/api` as we will be supporting login with montagu credentials, on an instance level in future (so the api endpoint should be the same regardless of auth provider).


absternator (approved):
LGTM nicueee :smile:

PR #46, opened 2024-01-02
Mrc-4772 Redux, axios, login cleanup 

absternator:
The following PR is mainly a cleanup PR, meaning code removal and some files/ logic. The main addition is the replacement of the login state with the new user state. The following is contained within this PR: 
- Remove redux, axios
- Add new Context for user state and app config. This is now used when any user info is needed, or checking if auth is needed
- Extract jwt recieved from Backend to help build user state
- When retrieving bearer add logic to remove user is token expired
- Unit tests 

Dev testing - mainly regression tests: 
- logging in and out works 
- avatar works with details of the person 
- Pages load as expected when logged in


app/src/app/components/header/AccountHeaderDropdown.tsx
```diff
@@ -12,31 +11,31 @@ import {
...
   };
 
   return (
-    <DropdownMenu data-testid="test1">
+    <DropdownMenu>
       <DropdownMenuTrigger asChild>
         <Button variant="ghost" className="relative h-10 w-10 rounded full">
           <Avatar className="h-8 w-8">
-            <AvatarFallback>LA</AvatarFallback>
+            <AvatarFallback>{getInitials(user?.displayName ?? "X X")}</AvatarFallback>
```
M-Kusumgar, line 30: this is probably causing failures
```suggestion
            <AvatarFallback>{getInitials(user?.displayName || "X X")}</AvatarFallback>
```

M-Kusumgar: although thinking about it again, I think the `getInitials` function should be the one setting the default maybe rather than throwing an error?

app/src/lib/string.ts
M-Kusumgar, line 1: general comment about this file, i dont think anything necessarily goes wrong if for example we dont get a firstname and a lastname in that format. while im sure this is standard it could change, i think we should handle this in a softer way than throwing an error, maybe just defaulting to a placeholder or something like that

app/src/app/components/login/UserAuthForm.tsx
```diff
@@ -3,29 +3,29 @@
...
   useEffect(() => {
-    dispatch(actions.fetchAuthConfig());
-  }, []);
-  useEffect(() => {
-    if (isAuthenticated) {
+    if (user?.token) {
       navigate("/");
     }
-  }, [isAuthenticated]);
+  }, [user]);
```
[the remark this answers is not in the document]
absternator: done

app/src/app/components/providers/UserProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { LocalStorageKeys } from "../../../lib/types/LocalStorageKeys";
+import { PacketJwtPayload } from "../../../types";
+import { UserProviderState, UserState } from "./types/UserTypes";
+
+const UserContext = createContext<UserProviderState | undefined>(undefined);
+
+export const useUser = () => {
+  const context = useContext(UserContext);
+
+  if (context === undefined) throw new Error("useUser must be used within a UserProvider");
```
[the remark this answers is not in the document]
absternator: its just a general thing to ensure no one uses it outside provider. THe reason it is not used in auth config is because that component actually fetches data and can be in an error state,, thus cant do this error handling

app/src/lib/string.ts
absternator: upated to return default as suggested

app/src/lib/string.ts
```diff
@@ -0,0 +1,13 @@
+export const getInitials = (name: string) => {
```
M-Kusumgar, line 1: I think it is probably a good idea to change name type to `string | undefined` and then return the default "XX" if undefined as well, this way we dont have to have a mention of "XX" anywhere else except this component and in the `AccountHeaderDropdown` file you can just do `getInitials(user?.displayName)` because right now we have default "XX" being set in the component and defaults being set in this function

M-Kusumgar (commented):
Looks very good! Had a minor comment about that `getInitials` function. Also should probably have a test that tests the remove user/logout logic, think codecov is complaining about it!

Also bug I found that could be related to changes made potentially and that you may want to address at some point:
When you click your avatar for the dropdown, the scroll bar on the home page disappears when the dropdown is open and then comes back when the dropdown is closed

app/src/lib/string.ts
absternator: done

M-Kusumgar (approved):
Looks good to me :)

PR #47, opened 2024-01-04
mrc-4826 Fix display of html resources when user is authenticated

EmmaLRussell:
This ticket allows html inline resources to be displayed, by fetching their data with auth header (already implemented for download). If you run the app with auth enabled, and look at the packet details for a packet with html file, e.g. "test3" you should see that the html file is rendered correctly embedded in the summary page, and you can also view it fullscreen by clicking on the link. 

As described in stand-up, this isn't ideal for View Fullscreen, as you get an odd looking address bar url which isn't persistent - I'm going to rework this so that it uses the router to display fullscreen with a sensible link, and also get redirected properly if login is required. But I think it's worth merging this simple approach as an initial implementation so we can start testing instances next week. 
Tickets for these improvements: 
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4875/Redirect-to-requested-url-on-login
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4876/Make-persistent-links-for-fullscreen-html-files


app/src/app/components/contents/packets/PacketReports.tsx
absternator, line 1: found a tiny bug.. in dark mode cant see in the iframe and text inside(should say TEST)... Maybe a good idea to add border around iframe? 
image

absternator (commented):
Great work !!! just a minor comment and bug on dark mode :smile:

app/src/app/components/contents/packets/PacketReports.tsx
EmmaLRussell: Ha, nice catch! I need to remember to test on the dark side.. :bat:

EmmaLRussell:
Dark mode bug should be fixed now, and I've handled error case. I've also removed 'PacketFile' component which wasn't being used, and just reproduced PacketReports partially. If we need to split this up in future we'll do so in a new component.

absternator (approved):
LGTM

PR #48, opened 2024-01-08
mrc-4875 redirect on login

EmmaLRussell:
As discussed! If you attempt to browse direct to a protected route (e.g. packet group or specific packet page) while not logged in, you should be redirected first to login screen, then after logging in, to the page you originally requested. 
If you explicitly log out from any page, then log in again, you should be taken to the home page, not the page you were on when you logged out.

This uses local storage rather than session storage to keep the `requestedUrl` value as Chrome appeared to be losing sessionStorage values after the redirect from the back end on authenticate. 

A new provider, `RedirectOnLoginProvider` gives access to the `requestedUrl` and also `loggingOut` flag which is used to avoid inadvertently setting current location as requested url on log out (maybe there's a more elegant way to do this but I couldn't think of one!)

Key component changes:
- `ProtectedRoute` - if the user is unauthenticated, before navigating to the login page, this saves `requestedUrl` but only if the user is not explicitly logging out (checks `loggingOut` flag).  
- `Redirect` - this is the entry point when user is redirected back to the front end after authenticating with the back end. Rather than just redirecting to the home page as before, this redirects to `requestedUrl` if it is set. It resets `requestedUrl` to null immediately before doing the actual redirect so stale value doesn't hang around in storage. 
- `AccountHeaderDropdown`: sets `loggingOut` flag to true when user clicks "Log out"
- `UserAuthForm` - resets `loggingOut` flag to false - logout has completed if the auth form is now being rendered

Codecov is moaning, but patch is 100% covered...


absternator (commented):
Works like a charm awesome.. Just some things that got me thinking 
1. Do we need to use local storage? Its very short lived thus the context state thats provided may be sufficient? Have a test without local storage maybe
2. The logout flags? Is that needed? Because To see that you have to be authenticated in the first place? Thus wont be calling func `setRequestedUrl` ever, thus will be empty all the time?

absternator (approved):

PR #49, opened 2024-01-10
mrc-4876 Persistent links for html files

EmmaLRussell:
This replaces temporary file object urls for fullscreen reports with links to report files which are persistent and look sensible. 

The links are of the form `/[packet name]/[packet id]/file/[filename]`.

When you click on a 'View fullscreen' link for a report in a packet summary page you should now be able to copy the link in the navbar, and browse directly to the fullscreen report in another tab or browser window, including being redirected there if you need to log in first. 

I've moved the rendering of the report in the iframe into a new component `PacketReport`. This is contained both by the `PacketReports` component (which shows fullscreen link etc) and by a new `PacketFileFullScreen` component which provides the packet context  when the packet is being rendered fullscreen.
I needed to rearrange the top level component a bit, moving `App` into the router, so that full screen could avoid showing the navbar (which was previously rendered outside the router).


app/src/app/components/contents/packets/PacketReport.tsx
```diff
@@ -0,0 +1,47 @@
...
+    useEffect(() => {
+        const file = getHtmlFileIfExists(packet);
+        if (file?.path != fileName) {
+            // NB Currently we only support displaying the first html file in a packet, but we may support multiple
+            // in future
+            setError(new Error("File name not found"));
+        } else {
+            getHtmlFileObjectUrl(packet);
+        }
+    }, []);
```
absternator, line 38: missing packet prop in depndency array

EmmaLRussell:
> Nice a clean solution and code looks good. My full screen looks like this.. is it okay? image

Should be fixed now.

PR #50, opened 2024-01-12
mrc-4930 Switch off auth by default

EmmaLRussell:
For deployment purposes, we want auth to be switched off by default, so that existing packit deployment configs continue to work.

If auth-related values are not present in a deployment config we will leave this default unchanged.


EmmaLRussell:
Closing this as not required - all config values are provided by the deploy tool

PR #51, opened 2024-01-17
mrc-4922 Set github oauth2 redirect url in application.properties, and do not use Router for github login link

EmmaLRussell:
Use PACKIT_API_ROUTE value injected into app from deploy tool to build Github OAuth2 redirect url in application.properties. 

See [associated packit-deploy PR](https://github.com/mrc-ide/packit-deploy/pull/8) for further explanation (issue 1.)

Also, in the front end, use a plain `<a>` tag for  Github login link rather thab a Router Link - see issue 2. in associated PR.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -10,6 +10,11 @@ Before you can start packit you need to run `./scripts/run-dependencies` from th
 to start database and `outpack_server` instances.
 
 ## Starting App
+The default dev configuration enables Github auth, but you will need to set environment variables GITHUB_CLIENT_ID and
+GITHUB_CLIENT_SECRET with the details of an OAuth app which you can use to log in. You can use the details held in the 
+mrc vault at `VAULT:secret/auth/githubclient/id:value` and `VAULT:secret/auth/githubclient/secret:value`. Alternatively,
+switch off `auth.enabled` in `config.properties`.
+
 Run up app on command line from the /api directory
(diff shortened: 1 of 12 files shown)


app/src/config/appConfig.ts
```diff
@@ -17,3 +17,4 @@ if (process.env.NODE_ENV == "production") {
 }
 
 export default appConfig;
+export const githubAuthEndpoint = (config: AppConfig) => `${config.apiUrl()}/oauth2/authorization/github`;
```
EmmaLRussell, line 20: I added this util when I thought we might have to make a special case to exclude this url from login redirection, so would need to build from separate places. Didn'tdo that in the end, but it's probably ok to leave this as the responsibility of appConfig module.

api/README.md
```diff
@@ -10,8 +10,8 @@ Before you can start packit you need to run `./scripts/run-dependencies` from th
 to start database and `outpack_server` instances.
 
 ## Starting App
-Run up app on command line from the project root directory
```
absternator, line 13: my fault i forgot to add.. but should mention about also adding in GITHUB_CLIENT_ID & GITHUB_CLIENT_SECRET..... otherwise this alll wont work

absternator (approved):
LGTM... just 1 minor comment

richfitz (approved):

PR #52, opened 2024-01-25
Mrc-4989 Remove 'TODO's from packit UI

absternator:
Remove TODO items from UI. Left commented out as it will be easy to find them when we need to implement. 

The below is the things removed: 
Accessibility
Manage access
Publish packets
Runner
Workflow runner
Documentation
Change logs

image
image


EmmaLRussell (approved):
Sorry, forgot to review this before I went on leave! Code looks fine to me. I think I've seen that codecov issue before, it's generally been an intermittent issue.

PR #54, opened 2024-02-28
mrc-5112 API auth integration test

EmmaLRussell:
Adds integration test for the API login endpoint, for both success and failure cases. The gha pulls secrets from the repository for the github app details and a working access token. 

I've also moved custom request factory setup into the base IntegrationTest class (it was previously just being used in PackitExceptionHandlerTest), in order to avoid unwanted errors on 401 responses.


absternator (approved):
Nice and clean integration tests all looks good to me :smile:

PR #53, opened 2024-02-22
mrc-5092 Proof of concept: support basic auth for Montagu

EmmaLRussell:
See ticket for full explanation - we will support Montagu by supporting basic auth. This partial implement just includes a hard-coded test user.

PR #55, opened 2024-03-27
Mrc-5135 Basic Auth initial work 

absternator:
This work is the initial work for basic auth in packet. It works currently for 1 hardcoded user and future work will allow saving of users and checking properly.  Most of the backend work is copied over from emmas POC --> https://github.com/mrc-ide/packit/pull/53. 

Testing: 
 Ensure you can login with hardcoded user and receive back valid token via the API and the frontend - email: test.user@example.com, password: password
Note: to check validity hit the /packets endpoint with the token as a bearer auth

image


api/app/src/test/kotlin/packit/integration/controllers/LoginControllerTest.kt
```diff
@@ -18,9 +22,11 @@ class LoginControllerTest : IntegrationTest()
     }
 
     @Test
-    fun `can login with API`()
+    fun `can login with github API`()
```
EmmaLRussell, line 25: Could add an integration test for hitting the basic endpoint too

api/app/src/test/kotlin/packit/unit/AppConfigTests.kt
```diff
@@ -63,6 +63,6 @@ class AppConfigTests
         assertEquals(sut.authJWTSecret, authSecret)
         assertEquals(sut.authRedirectUri, redirectUrl)
         assertEquals(sut.outpackServerUrl, "outpackServerUrl")
-        assertEquals(sut.authEnableGithubLogin, enableGithubLogin)
+        assertEquals(sut.authEnableGithubLogin, true)
```
EmmaLRussell, line 66: Might as well add tests for basic too.

app/src/app/components/login/BasicUserAuthForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Input } from "../Base/Input";
+import { useUser } from "../providers/UserProvider";
+
+export const BasicUserAuthForm = () => {
+  const navigate = useNavigate();
+  const { setUser } = useUser();
+
+  const formSchema = z.object({
+    email: z.string().email(),
+    password: z.string().min(8)
```
EmmaLRussell, line 20: Is this saying the password needs to be at least 8 characters?
That doesn't seem like something we should be enforcing in the login form, but rather when credentials are created.

app/src/app/components/login/BasicUserAuthForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+        url: `${appConfig.apiUrl()}/auth/login/basic`,
+        body: values,
+        method: "POST",
+        noAuth: true
+      })) as { token: string };
+
+      setUser(data.token);
+      navigate("/");
+    } catch (error) {
+      console.error(error);
```
EmmaLRussell, line 42: Did you meant to leave this in?

app/src/lib/fetch.ts
```diff
@@ -1,25 +1,27 @@
...
       Accept: "application/json",
       "Content-Type": "application/json",
-      ...getAuthHeader()
+      ...(!noAuth && getAuthHeader())
     }
   });
 
   if (!res.ok) {
-    throw new Error("API error");
+    throw new ApiError("API error", res.status);
```
EmmaLRussell, line 24: There is potentially user-readable information coming back in the API errors, so I'm not sure it should always just hardcode the message to "API error" like this. But maybe not something to address now.

app/src/tests/components/login/Redirect.test.tsx
```diff
@@ -25,22 +26,24 @@ jest.mock("../../../app/components/providers/RedirectOnLoginProvider", () => ({
...
+          </MemoryRouter>
+        </UserProvider>
+      </AuthConfigProvider>
     );
   };
 
   beforeEach(() => {
     mockRequestedUrl = null;
-    jest.clearAllMocks();
+    // jest.clearAllMocks();
```
EmmaLRussell, line 46: Why is this commented out?

EmmaLRussell (commented):
Great, thanks for this, looks great.

api/app/config/detekt.yml
```diff
@@ -45,6 +45,8 @@ style:
     active: false
   SerialVersionUIDInSerializableClass:
     active: false
+  ForbiddenComment:
```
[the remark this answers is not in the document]
absternator: yup haha

api/app/src/main/kotlin/packit/controllers/LoginController.kt
```diff
@@ -20,16 +23,35 @@ class LoginController(
         @RequestBody @Validated user: LoginWithToken,
     ): ResponseEntity<Map<String, String>>
     {
+        if (!config.authEnableGithubLogin)
+        {
+            return ResponseEntity.badRequest().body(mapOf("error" to "Github login is not enabled"))
```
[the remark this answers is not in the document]
absternator: fair fair.. but as discovered currently an thrown error just results in a 401 with no body... I've created a ticket for to investigate this next

api/app/src/main/kotlin/packit/controllers/LoginController.kt
```diff
@@ -20,16 +23,35 @@ class LoginController(
...
 
+    @PostMapping("/login/basic")
+    @ResponseBody
+    fun loginBasic(
+        @RequestBody @Validated user: LoginWithPassword
+    ): ResponseEntity<Map<String, String>>
+    {
+        if (!config.authEnableBasicLogin)
+        {
+            return ResponseEntity.badRequest().body(mapOf("error" to "Basic login is not enabled"))
```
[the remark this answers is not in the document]
absternator: as per above

api/app/src/main/kotlin/packit/service/BasicLoginService.kt
```diff
@@ -0,0 +1,37 @@
...
+class BasicLoginService(
+    val jwtIssuer: JwtIssuer,
+    val authenticationManager: AuthenticationManager
+)
+{
+    fun authenticateAndIssueToken(loginRequest: LoginWithPassword): Map<String, String>
+    {
+        if (loginRequest.email.isEmpty() && loginRequest.password.isEmpty())
+        {
+            throw PackitException("Empty user details", HttpStatus.BAD_REQUEST) // TODO: use a key here
```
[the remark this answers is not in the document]
absternator: added as requested

app/src/app/components/login/BasicUserAuthForm.tsx
absternator: its still a form validation so we dont send the request ... ive added at least 8 character password when creating users in the future tickets

app/src/lib/fetch.ts
absternator: yeah currently as issue discuessed above with 401 with no body.. so after that fix can sort

app/src/tests/components/login/Redirect.test.tsx
absternator: by mistake good catch thanks

api/app/src/main/kotlin/packit/controllers/LoginController.kt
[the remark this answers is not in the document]
absternator: found issue and now throwing as reccomended

api/app/src/test/kotlin/packit/integration/controllers/LoginControllerTest.kt
absternator: i couldnt unless im being stupid.. but requires a change in config.properties file.. which is actually not even a spring loaded one.... so I had an idea of moving config properties all into the spring application.properties so they can be controlled in tests

EmmaLRussell: Oh yeah, that's a pain. Let's make a separate ticket to do that so we can have a fuller set of integration tests.

EmmaLRussell (approved):
Looks good. Still not convinced we should be showing the password validation message on login tbh but don't think it's a big issue.

PR #60, opened 2024-04-23
Mrc 5231- Fix Error Responses + update create endpoint

absternator:
This PR does the following 
- Catches 2 exceptions in the handler that are thrown, thus will propagate to frontend... before was just throwing 401 with empty body
- Enable method level security and use that -- will be used for fine grained auth as well

### DIFF
--- a/api/app/src/main/kotlin/packit/controllers/UserController.kt
+++ b/api/app/src/main/kotlin/packit/controllers/UserController.kt
@@ -2,7 +2,10 @@ package packit.controllers
 
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
+import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.security.core.Authentication
+import org.springframework.security.core.context.SecurityContext
+import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.stereotype.Controller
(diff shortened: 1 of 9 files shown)


api/app/src/main/kotlin/packit/controllers/UserController.kt
```diff
@@ -11,23 +11,18 @@ import org.springframework.web.bind.annotation.RequestMapping
...
 import packit.model.CreateBasicUser
-import packit.security.Role
 import packit.service.UserService
 
 @Controller
 @RequestMapping("/user")
 class UserController(private val config: AppConfig, private val userService: UserService)
 {
     @PostMapping("/basic")
+    @PreAuthorize("hasAuthority('ADMIN')")
```
EmmaLRussell, line 21: Starting to feel like roles like "ADMIN" should be coming from a type rather than have hardcoded strings in multiple places.

EmmaLRussell (approved):
Looks good! One tiny comment, but up to you..

api/app/src/main/kotlin/packit/controllers/UserController.kt
absternator: will have to be changes in next pr so have left

PR #64, opened 2024-05-01
Mrc-5298-Role Update Endpoints

absternator:
The work here contains 2 endpoints that encompass role update. They are `addPermissionsToRole` and `removePermissionsToRole`. They respectively add and remove permissions from a role .. 

NOTE: 
For removePermissionsToRole we had to delete the rolepermission rows directly. Hibernate didn't work with just removing the rolePermission from the role. I will create a new ticket that can come back and investigate this further


api/app/src/main/kotlin/packit/service/RoleService.kt
```diff
@@ -52,17 +55,37 @@ class BaseRoleService(
...
+
+        role.rolePermissions.addAll(rolePermissionsToAdd)
+        roleRepository.save(role)
+    }
+
+    override fun removePermissionsFromRole(roleName: String, removeRolePermissions: List<UpdateRolePermission>)
+    {
+        val role = roleRepository.findByName(roleName)
+            ?: throw PackitException("roleNotFound", HttpStatus.BAD_REQUEST)
+        rolePermissionService.removeRolePermissionsFromRole(role, removeRolePermissions)
```
EmmaLRussell, line 86: It seems a bit odd to check that permissions being added don't yet exist here in `RoleService` but that permissions deing deleted do exist, in `RolePermissionService` service. Should both be in the same service, probably `RolePermissionsService`

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val permissionRepository: PermissionRepository,
+    private val packetRepository: PacketRepository,
+    private val packetGroupRepository: PacketGroupRepository,
+    private val tagRepository: TagRepository,
+    private val rolePermissionRepository: RolePermissionRepository
+) : RolePermissionService
+{
+    override fun getRolePermissionsToUpdate(
+        role: Role,
+        addRolePermissions: List<UpdateRolePermission>
```
EmmaLRussell, line 28:
```suggestion
        updateRolePermissions: List<UpdateRolePermission>
```

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -0,0 +1,64 @@
...
+                },
+                packetGroup = addRolePermission.packetGroupId?.let {
+                    packetGroupRepository.findById(it)
+                        .orElseThrow { PackitException("packetGroupNotFound", HttpStatus.BAD_REQUEST) }
+                },
+                tag = addRolePermission.tagId?.let {
+                    tagRepository.findById(it).orElseThrow { PackitException("tagNotFound", HttpStatus.BAD_REQUEST) }
+                }
+            )
+        }
```
EmmaLRussell, line 49: This is quite dense visuall,y but makes sense on closer inspection. 
I think an empty line at l.34 would make it more readable by splitting off the main chunk.

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -0,0 +1,64 @@
...
+    override fun removeRolePermissionsFromRole(role: Role, removeRolePermissions: List<UpdateRolePermission>)
+    {
+        val rolePermissionsToRemove = getRolePermissionsToUpdate(role, removeRolePermissions)
+
+        val matchedRolePermissionsToRemove = rolePermissionsToRemove.map { rolePermissionToRemove ->
+            val matchedPermission = role.rolePermissions.find { rolePermissionToRemove == it }
+                ?: throw PackitException("rolePermissionDoesNotExist", HttpStatus.BAD_REQUEST)
+            matchedPermission
+        }
+
```
EmmaLRussell, line 61: I don't think you really need to do a map here, just a `forEach` then just map the original `rolePermissionsToRemove` to ids to pass to the `deleteAllByIdIn` method

api/app/src/test/kotlin/packit/integration/controllers/RoleControllerTest.kt
```diff
@@ -83,4 +97,50 @@ class RoleControllerTest : IntegrationTest()
         assertSuccess(result)
         assertNull(roleRepository.findByName("testRole"))
     }
+
+    @Test
+    @WithAuthenticatedUser(authorities = ["user.manage"])
```
EmmaLRussell, line 102: Could test for unauthorized request too.

EmmaLRussell (commented):
Could you explain a bit more about the need for equality/hash checking for the role permissions? Is it related to trying to get hibernate to handle the deletion of the rows in the link table? Or was it to support this line in RoleService: 
`if (rolePermissionsToAdd.any { role.rolePermissions.contains(it) })`
...
or in RolePermissionService
`val matchedPermission = role.rolePermissions.find { rolePermissionToRemove == it }`
?
Those would be easier if you were dealing with the data classes (dto's) which would support equality automatically.

absternator:
> Could you explain a bit more about the need for equality/hash checking for the role permissions? Is it related to trying to get hibernate to handle the deletion of the rows in the link table? Or was it to support this line in RoleService: `if (rolePermissionsToAdd.any { role.rolePermissions.contains(it) })` ... or in RolePermissionService `val matchedPermission = role.rolePermissions.find { rolePermissionToRemove == it }` ? Those would be easier if you were dealing with the data classes (dto's) which would support equality automatically.

Yup so pretty much used to compare equality fo the entities.. For delete we need to filter out some out and things just in general for match... In general good to have them for these as can be used for comparison.. Sadly if we were to use dtos we would have to first convert and then compare and then convert back to entity

api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -21,13 +22,35 @@ class RoleController(private val roleService: RoleService)
...
+        @PathVariable roleName: String
     ): ResponseEntity<Map<String, String?>>
     {
-        roleService.deleteRole(name)
+        roleService.deleteRole(roleName)
 
         return ResponseEntity.ok(mapOf("message" to "Role deleted"))
     }
+
+    @PutMapping("/add-permissions/{roleName}")
```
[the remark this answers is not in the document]
absternator: its not really a post or delete because we are not creating/deleting the role. We are updating the role object i.e its a update operation. technically should be patch but spring didn't like patch - said it doesn't accept them

api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -21,13 +22,35 @@ class RoleController(private val roleService: RoleService)
...
+    fun addPermissionsToRole(
+        @RequestBody @Validated addRolePermissions: List<UpdateRolePermission>,
+        @PathVariable roleName: String
+    ): ResponseEntity<Map<String, String>>
+    {
+        roleService.addPermissionsToRole(roleName, addRolePermissions)
+
+        return ResponseEntity.ok(mapOf("message" to "Permissions added"))
```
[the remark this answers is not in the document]
absternator: yup agree have created ticjet fir all these as prev pr comment

api/app/src/main/kotlin/packit/model/dto/UpdateRolePermission.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import org.jetbrains.annotations.NotNull
+
+data class UpdateRolePermission(
+    @field:NotNull
+    val permission: String,
+    val packetId: String? = null,
+    val tagId: Int? = null,
+    val packetGroupId: Int? = null
+)
```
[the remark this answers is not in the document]
absternator: I think for the create the FE most likely wont pass any permissions.. And here update is where it can pass in scoped if needed...The idea was it would be weird adding scoped permissions when creating a role as well... But adding a scoped permission when updating makes sense also....

api/app/src/main/kotlin/packit/model/RolePermission.kt
```diff
@@ -29,3 +29,38 @@ class RolePermission(
...
     var id: Int? = null,
 )
+{
+    init
+    {
+        val nonNullFields = listOf(packet, tag, packetGroup).count { it != null }
+        require(nonNullFields <= 1) {
+            "Either all of packet, tag, packetGroup should be null or only one of them should be not null"
+        }
+    }
```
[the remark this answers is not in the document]
absternator: yeah i thought that too but couldnt find a way to implement once and use both places.. Also they a bit different as one is ids and other is whole objects... And probs don't need but put here to for safety and mirror this constraint on DB and make more visible (as many people don't check db constraints haha)

api/app/src/main/kotlin/packit/model/RolePermission.kt
```diff
@@ -29,3 +29,38 @@ class RolePermission(
...
+    override fun hashCode(): Int
+    {
+        val prime = 31
+        var result = role.name.hashCode()
+        result = prime * result + permission.name.hashCode()
+        result = prime * result + packet?.id.hashCode()
+        result = prime * result + packetGroup?.id.hashCode()
+        result = prime * result + tag?.id.hashCode()
+        return result
+    }
```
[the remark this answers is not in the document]
absternator: yeah i did that initially but lint complained and said I need equals and haschode... and read and found its best to have both

api/app/src/main/kotlin/packit/model/RolePermission.kt
```diff
@@ -29,3 +29,38 @@ class RolePermission(
...
+            this === other -> true
+            other !is RolePermission -> false
+            role.name != other.role.name -> false
+            permission.name != other.permission.name -> false
+            packet?.id != other.packet?.id -> false
+            packetGroup?.id != other.packetGroup?.id -> false
+            tag?.id != other.tag?.id -> false
+            else -> true
+        }
+    }
```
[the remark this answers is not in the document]
absternator: need on entity so can filter out data returned from entities... also mostly services using entites

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: i named like that to differentiate the adding and deleting of role permissions

api/app/src/main/kotlin/packit/model/dto/UpdateRolePermission.kt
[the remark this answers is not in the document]
EmmaLRussell: OK, but you still want to allow global permissions on create to make life easier for the admin cli/setup scripts? 
I'm not really convinced it's a useful distinction. Setup *might* want to make roles with scoped permissions in some cases for example. I think it would be fine to make role create take no perms and require that to be done in a second request.

api/app/src/main/kotlin/packit/model/RolePermission.kt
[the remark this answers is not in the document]
EmmaLRussell: Yeah, pros and cons of both approaches, but as discussed it does seem easier to have the services talking to each other with entities.

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
EmmaLRussell: But you're using it for permissions being removed too: 
https://github.com/mrc-ide/packit/pull/64/files#diff-91662b556b4c3ee7dc5037863b1a55465c6b8218de7a6675c404baec77f0fb66R56

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
EmmaLRussell: Yeah sorry that doesn't work because you don't actually have the ID on the UpdateRolePermissions.

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -0,0 +1,81 @@
...
+            val matchedPermission = role.rolePermissions.find { rolePermissionToRemove == it }
+                ?: throw PackitException("rolePermissionDoesNotExist", HttpStatus.BAD_REQUEST)
+
+            matchedPermission
+        }
+
+        rolePermissionRepository.deleteAllByIdIn(matchedRolePermissionsToRemove.map { it.id!! })
+    }
+
+    override fun getAddRolePermissionsFromRole(
```
EmmaLRussell, line 68: It's quite confusing having getRolePermissionsToUpdate and getAddRolePermissionsFromRole (that sounds like it's getting the permissions *from* the role, but if the permissions are already in the role, you're going to throw an exception!). I wouldn't mind just making the method into a checkPermissionsToAddAreNotOnRole or something, and have the main update method call `getRolePermissionsToUpdate directly, and pass the perms into here to just do the check...

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: oh yup was mistaken have updated in PR --> https://github.com/mrc-ide/packit/pull/68

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: yup agree ended up changing to `getRolePermissionsToAdd` and making the `getRolePermissionsToUpdate` internal so not visible to public

EmmaLRussell (approved):

PR #65, opened 2024-05-02
Mrc-5296  Role read 

absternator:
This PR contains 3 endpoints for Role -- 1. Reading rolenames, 2. Reading roles with relationships, 3. Reading specific roles with relationships.... The latter 2 have enough info so they can be deleted and show view as in orderly web currently


api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -53,4 +55,38 @@ class RoleController(private val roleService: RoleService)
...
+    fun getNonUsernameRoles(): ResponseEntity<List<RoleDto>>
+    {
+        val roles = roleService.getRolesWithRelationships(false)
+        return ResponseEntity.ok(roles.map { it.toDto() })
+    }
+
+    @GetMapping("/{roleName}")
+    fun getRole(@PathVariable roleName: String): ResponseEntity<RoleDto>
+    {
+        val role = roleService.getRole(roleName)
```
EmmaLRussell, line 89: Is this a role "WithRelationships" too? It's returning the same Dto type so I assume so. Maybe drop the "withRelationships" from the plural method in the service for consistency.

api/app/src/main/kotlin/packit/model/dto/BasicPacketDto.kt
```diff
@@ -0,0 +1,6 @@
+package packit.model.dto
+
+data class BasicPacketDto(
+    val name: String,
+    val id: String,
+)
```
EmmaLRussell, line 6: Given that the 'Basic' Packet Dto, the Tag Dto and the PacketGroup Dto are all just ids and names, I think there's an argument for having a single PermissionScope model which has name and id, and scope type. It going to make wrangling these things easier in the front end. But I guess I already lost that argument in the database and POST methods, and you can always put some utils in the front end to work around the awkwardness.

api/app/src/main/kotlin/packit/service/RoleService.kt
```diff
@@ -86,6 +91,27 @@ class BaseRoleService(
...
+    {
+        return roleRepository.findAll().map { it.name }
+    }
+
+    override fun getRolesWithRelationships(): List<Role>
+    {
+        return roleRepository.findAll()
+    }
+
+    override fun getRolesWithRelationships(isUsernames: Boolean): List<Role>
```
EmmaLRussell, line 104: Does "withRelationships" here just mean, 'including permissions'? ...oh, and users?

api/app/src/test/kotlin/packit/integration/controllers/RoleControllerTest.kt
```diff
@@ -143,4 +144,92 @@ class RoleControllerTest : IntegrationTest()
...
+
+        assertEquals(ObjectMapper().writeValueAsString(allUsernameRoles), result.body)
+    }
+
+    @Test
+    @WithAuthenticatedUser(authorities = ["user.manage"])
+    fun `users can get non username roles with relationships`()
+    {
+        roleRepository.save(Role("randomUser", isUsername = true)).toDto()
+        val adminRole = roleRepository.findByName("ADMIN")!!.toDto()
```
EmmaLRussell, line 206: Isn't there a bit of a dangerous assumption in this sort of test that your repository is functioning correctly? It might not be pulling all values out correctly, in which case we'd just be comparing the wrong values from direct repo call with the wrong values from the endpoint and passing the test. We're not testing the repository elsewhere than in these integration tests, so that could happen. It feels like the test objects to compare against should be defined in the test itself. You'll need to pull the ids out of the db, but the rest can be hardcoded.

EmmaLRussell (commented):
Started reviewing this before I looked at some of the intervening branches, hence re-raising issue of dto vs entity classes.

api/app/src/main/kotlin/packit/model/dto/RoleDto.kt
```diff
@@ -0,0 +1,8 @@
+package packit.model.dto
+
+data class RoleDto(
+    var name: String,
+    var rolePermissions: List<RolePermissionDto> = listOf(),
+    var users: List<UserDto> = listOf(),
+    var id: Int
+)
```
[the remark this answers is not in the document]
absternator: as discussed created ticket and can do if needed

api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -53,4 +55,38 @@ class RoleController(private val roleService: RoleService)
...
     }
+
+    @GetMapping
+    fun getRoleNames(): ResponseEntity<List<String>>
+    {
+        return ResponseEntity.ok(roleService.getRoleNames())
+    }
+
+    @GetMapping("/complete")
+    fun getRolesWithRelationships(): ResponseEntity<List<RoleDto>>
```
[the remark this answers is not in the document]
absternator: agreeed

api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -53,4 +55,38 @@ class RoleController(private val roleService: RoleService)
...
+        val roles = roleService.getRolesWithRelationships(true)
+        return ResponseEntity.ok(roles.map { it.toDto() })
+    }
+
+    @GetMapping("/complete/non-usernames")
+    fun getNonUsernameRoles(): ResponseEntity<List<RoleDto>>
+    {
+        val roles = roleService.getRolesWithRelationships(false)
+        return ResponseEntity.ok(roles.map { it.toDto() })
+    }
```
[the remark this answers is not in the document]
absternator: agreed

api/app/src/main/kotlin/packit/model/dto/BasicPacketDto.kt
absternator: the naming is a bit weird i think il try rename... but I don't think scope thing will help, just means one more field to handle.... the FE will be easy because it will just map through them all and display name (2 of them will be null thus nothing will show)

api/app/src/main/kotlin/packit/service/RoleService.kt
absternator: have renamed as per above suggestion

api/app/src/test/kotlin/packit/integration/controllers/RoleControllerTest.kt
absternator: as discussed have created ticket for repository tests

EmmaLRussell (approved):

PR #67, opened 2024-05-03
feat: delete user

absternator:
User delete endpoit... deletes user and username role


api/app/src/main/kotlin/packit/controllers/UserController.kt
```diff
@@ -52,4 +52,14 @@ class UserController(private val config: AppConfig, private val userService: Use
...
     }
+
+    @DeleteMapping("/{username}")
+    fun deleteUser(
+        @PathVariable username: String
+    ): ResponseEntity<Map<String, String?>>
+    {
+        userService.deleteUser(username)
+
+        return ResponseEntity.ok(mapOf("message" to "User deleted"))
```
EmmaLRussell, line 63: Ditto earlier branches, maybe just empty 200

api/app/src/test/kotlin/packit/integration/controllers/UserControllerTest.kt
```diff
@@ -132,4 +132,32 @@ class UserControllerTest : IntegrationTest()
...
+        val result = restTemplate.exchange(
+            "/user/$username",
+            HttpMethod.DELETE,
+            getTokenizedHttpEntity(),
+            String::class.java
+        )
+
+        assertSuccess(result)
+        assertEquals(userRepository.findByUsername(testUser.username), null)
+        assertEquals(roleRepository.findByName(username), null)
```
EmmaLRussell, line 161: Can we test that the cascades also worked to delete rows from the link tables?

EmmaLRussell (commented):
Right, I think that's the last one. Epic work!

Couple of themes to chat through tomorrow, but looks pretty solid to me generally...

api/app/src/main/kotlin/packit/controllers/UserController.kt
absternator: yup added ticket fir al these

api/app/src/test/kotlin/packit/integration/controllers/UserControllerTest.kt
absternator: will test all these in repostory tests as talked aboit

api/app/src/main/kotlin/packit/service/RoleService.kt
```diff
@@ -60,7 +60,7 @@ class BaseRoleService(
 
     override fun deleteRole(roleName: String)
     {
-        if (!roleRepository.existsByName(roleName))
+        if (!roleRepository.existsByName(roleName) || roleName == "ADMIN")
```
EmmaLRussell, line 63: Should probably be a different error if they've tried to delete ADMIN

EmmaLRussell: ...And should it also be an error if an attempt is made to delete a username role instead of deleting the user? (shouldn't be available in the FE, but might happen via the API... )

absternator: true admin one different error... 2nd comment that is done already as part of user-role update

EmmaLRussell (approved):

PR #69, opened 2024-05-16
Mrc 5359-Update Password 

absternator:
This work is for password update... The general flow is on login if user has not logged in they will get redirected to update password page. If that is successful then they will get redirected back to login and then good to go.

There is also a tiny update to some Kotlin integration tests data deletion. 

Dev Testing:
*Note: feel free to message me for setup help* 
PreReq:
1. Ensure you are running in basic auth by going to `application.properties` and changing `auth.method=basic` or setting `PACKIT_AUTH_METHOD` env variable to *basic*. Then run app and setup admin user by running `./scripts/basic-create-super-user`. (this info is in readme also :smile: )
2. Following step best through postman. 
i. login with user by hitting `http://localhost:8080/auth/login/basic` and put `{email, password}`(get from vault) in body. Copy the token you get back  
ii. use that token to create a new basic user by hitting POST `http://localhost:8080/user/basic` with body like `{
    "email": "d@gmail.com",
    "password": "password",
    "displayName": "random pussio"
}`

Actual Testing : 
1. Try logging in with new user created... should get redirected to update the password 
2. update this and you should be redirected back to login page and login and get into app


M-Kusumgar (approved):
Nice one, looks good to me bro!

api/app/src/main/kotlin/packit/service/UserService.kt
```diff
@@ -110,6 +117,20 @@ class BaseUserService(
...
     }
 
+    override fun updatePassword(username: String, updatePassword: UpdatePassword)
+    {
+        val user = userRepository.findByUsername(username)
+            ?: throw PackitException("userNotFound", HttpStatus.NOT_FOUND)
+
+        if (!passwordEncoder.matches(updatePassword.currentPassword, user.password))
+        {
+            throw PackitException("invalidPassword", HttpStatus.BAD_REQUEST)
```
EmmaLRussell, line 127: Feels like this should maybe be a 401 since it's a wrong password..? But it is embedded the the update password request body, rather than like a normal login, so maybe a 400 is ok..

Also - I'm not sure we should re-use this for a normally logged in user choosing to update their password, if that was the idea. In that case, they would already be logged in and should hit an endpoint secured with a token, rather than posting their password in the body again.

app/src/app/components/login/BasicUserAuthForm.tsx
```diff
@@ -40,48 +43,54 @@ export const BasicUserAuthForm = () => {
       navigate("/");
     } catch (error) {
       console.error(error);
-      if (error instanceof ApiError) {
-        if (error.status === 401) {
-          form.setError("email", { message: "Invalid email or password" });
-          form.setError("password", { message: "Invalid email or password" });
+      if (error instanceof ApiError && error.status === HttpStatus.Unauthorized) {
+        if (error.message.includes("must change your password")) {
```
EmmaLRussell, line 47: Relying on partial message strings feels a bit icky. we could use error codes.

app/src/app/components/login/BasicUserAuthForm.tsx
```diff
@@ -40,48 +43,54 @@ export const BasicUserAuthForm = () => {
...
+          return navigate(`/update-password?email=${values.email}&error=${error.message}`);
```
EmmaLRussell, line 48: So a non-new user *could* navigate here and change their password that way.?

api/app/src/main/kotlin/packit/service/UserService.kt
```diff
@@ -80,7 +83,11 @@ class BaseUserService(
     override fun getUserForLogin(username: String): User
     {
         val user =
-            userRepository.findByUsername(username) ?: throw PackitException("userNotFound", HttpStatus.UNAUTHORIZED)
+            userRepository.findByUsername(username) ?: throw AuthenticationException()
+        if (user.lastLoggedIn == null)
+        {
+            throw AuthenticationException("You must change your password before logging in.")
+        }
```
EmmaLRussell, line 90: Sorry I haven't actually run this up, but I think this means that even if the user puts in the wrong password on the initial login screen, they'll still get redirected to the change password page here, because the password checking happens after this is called (?). That's not catastrophic as they will still need to get their password correct in order to change it, but it does feel like a wrong password should be flagged immediately before getting handed to the next screen. User might interpret this behaviour as meaning that they did get their pw right the first time, and be confused when they can't complete the update password form with the same pw...

api/app/src/main/kotlin/packit/service/UserService.kt
absternator: yeah i initially had in user controller, but then realised they don't need auth.. technically yeah you are right but fundamentally we can resuse this same one... as they need they need their email and password  (same as login)... unless we make another update one that only requires new password

app/src/app/components/login/BasicUserAuthForm.tsx
absternator: yeah felt weird too but had to before when they both were 401.. but have updated so I can throw a 403 as per change suggested

app/src/app/components/login/BasicUserAuthForm.tsx
absternator: yeah if they want

api/app/src/main/kotlin/packit/service/UserService.kt
absternator: yip 100% right... initially was because behind spring auth.. but I just moved after the auth check..

api/app/src/main/kotlin/packit/service/UserService.kt
EmmaLRussell: I guess it's good practice to ask for old password again when they want to change it!

EmmaLRussell (approved):
lgtm 👍

PR #76, opened 2024-05-17
Mrc-5367- Initial setup of manage access

absternator:
This is the initial work manage access pages. It includes the following : 
- Add authorities to user from token 
- setup routes for update-roles and update-users pages 
- side bar and manage access button to get to this page (only if has *user.manage*) access

Dev testing:
- Login with super user and check you can see *manage access* button and can navigate. Also the sidebar afterwards to switch b/w update roles and update users. 
- Try deleting *user.manage* authority from local storage and refresh... the manage users button should not visible and you should not be able to go directly to /maange-roles or /manage-users

figma: https://www.figma.com/design/NMR5XdmWYfphrrRYgIV2US/Packit-User-manage?node-id=22-285&t=llDh1pAzXzdMAunV-0
image


app/src/app/components/contents/common/Unauthorized.tsx
```diff
@@ -0,0 +1,10 @@
+export const Unauthorized = () => {
+  return (
+    <div className="container h-[800px] flex items-center justify-center m-auto">
+      <div className="flex flex-col space-y-2 text-center">
+        <h1 className="text-2xl font-semibold tracking-tight text-red-500">401 Unauthorized</h1>
```
EmmaLRussell, line 5: Putting the 401 here might confuse us, since you haven't necessarily had a 401 back from the API to make this show! And probably wouldn't make much sense to non-tech users. :laughing:  But don't mind!

EmmaLRussell (commented):
Nice! Good to extract our the generic sidebar. 

I think the visible breadcrumb should be "human readable" i.e. "Manager Users" rather than "manage-users" - as is displayed in the sidebar. 

Someone unauthorized *could* hack their local storage into making the app show the link to the page, but they'd get nothing back from the API so there isn't a security issue with that..(?) 

But I do wonder if there's an issue with keeping roles and permissions together in the same "authorities" array. It's unlikely, but an unhinged admin could create a role with the same name as a permission (but which doesn't actually have that permission), which could lead to confusion in the UI.... and actually would that not potentially wrongly grant that permission to any user with the badly named role, since aren't the token 'authorities' used by Spring to apply authorization?

We still don't have e2e tests in packit, right? Getting some in to test e2e auth might not be a bad idea!

absternator:
> I think the visible breadcrumb should be "human readable" i.e. "Manager Users" rather than "manage-users" - as is displayed in the sidebar.
> 
> Someone unauthorized _could_ hack their local storage into making the app show the link to the page, but they'd get nothing back from the API so there isn't a security issue with that..(?)
> 
> But I do wonder if there's an issue with keeping roles and permissions together in the same "authorities" array. It's unlikely, but an unhinged admin could create a role with the same name as a permission (but which doesn't actually have that permission), which could lead to confusion in the UI.... and actually would that not potentially wrongly grant that permission to any user with the badly named role, since aren't the token 'authorities' used by Spring to apply authorization?
> 
> We still don't have e2e tests in packit, right? Getting some in to test e2e auth might not be a bad idea!


1. have made breadcrumb human readble 
2. Yup they can that's why we have locked down our api
3. Authotities is what we get back... if a a admin was to go ahead and name a role 'user.manage' then pretty sure they would want that permission.. (this is something stupid anyways and don't think anyone would)... also those would never get shown in the UI its just the ones extracted from the token 
4. Yup good idea I have made ticket for this -- will do straight after the manage access pages are done

EmmaLRussell:
As discussed, let's take roles out of the token authorities (since we only use permissions as authorities).

EmmaLRussell (approved):
Looks good - I think it would also be good to capitalise "Manage Users" (and roles) - to in the breadcrumb, to match the casing on the sidebar and the breadcrumb for "Home". Not a showstopper though..

PR #77, opened 2024-05-22
Mrc 5374-manage roles table

absternator:
This PR works for the roles table... All the data is showing and filtering and pagination is working.... No actions are implemented yet (any buttons) these will be in subsequent PRs... Also have added @JsonProperty to all dtos as this was needed by integration tests so can map result.body to exact types.. eg can do  `val userResult = ObjectMapper().readValue(result.body, object : TypeReference<UserDto>() {})`

image

Note: Have changed the Figma mockups part when displaying the permissions and users table.  I have made it a scroll... found it a lot nicer than the accordion thing as it would blow up the page.... also makes searching easier on the page with the scroll as all elements are showing
image


app/src/app/components/contents/common/DataTable.tsx
```diff
@@ -1,47 +1,74 @@
...
-  const table = useReactTable({ columns, data, getCoreRowModel: getCoreRowModel() });
+export const DataTable = <TData,>({ data, columns, enablePagination }: DataTableProps<TData>) => {
+  const table = useReactTable({
+    columns,
+    data,
+    getCoreRowModel: getCoreRowModel(),
+    ...(enablePagination && { getPaginationRowModel: getPaginationRowModel() }),
+    initialState: {
+      pagination: {
+        pageSize: PAGE_SIZE
```
EmmaLRussell, line 20: I think we might eventually want to support different page sizes for different entities - 50 packets might be about right, but might be too many for roles. But don't need to do anything about that now..

app/src/app/components/contents/explorer/PacketGroupSummaryList.tsx
```diff
@@ -24,7 +24,7 @@ export const PacketGroupSummaryList = ({
   if (isLoading)
     return (
       <ul className="flex flex-col border rounded-md">
-        {[...Array(8)].map((val, index) => (
+        {[...Array(2)].map((_, index) => (
```
EmmaLRussell, line 27: This is  just reducing the number of skeleton placeholders?

app/src/app/components/contents/common/FilterByName.tsx
```diff
@@ -0,0 +1,53 @@
...
+  setFilterByName: Dispatch<SetStateAction<string>>;
+  setPageNumber?: Dispatch<SetStateAction<number>>;
+}
+export const FilterByName = ({ setFilterByName, setPageNumber }: FilterByNameProps) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleSetNameFilter = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setFilterByName(event.target.value);
+    if (setPageNumber) {
+      setPageNumber(0);
```
EmmaLRussell, line 17: OK, so when the filter changes it resets the page number to 0. I'm not sure it's really the responsibility of the filter component to worry about paging. Perhaps the paging component should do the reset when its item list changes, But don't feel too strongly about that.

app/src/app/components/contents/common/FilterByName.tsx
```diff
@@ -0,0 +1,53 @@
+import debounce from "lodash.debounce";
+import { Dispatch, SetStateAction, useCallback, useEffect, useRef } from "react";
+import { Input } from "../../Base/Input";
+import { Button } from "../../Base/Button";
+import { X } from "lucide-react";
+
+interface FilterByNameProps {
+  setFilterByName: Dispatch<SetStateAction<string>>;
+  setPageNumber?: Dispatch<SetStateAction<number>>;
+}
```
EmmaLRussell, line 10: Actually, this is more generic than `FilterByName` - doesn't have to be name that you're filtering by, it could be anything, as interpreted by the action. Placeholder could be a prop, and this could just be `FilterByUserInput` or something. That's almost certainly going to happen when we introduce more complex filtering.

app/src/app/components/contents/manageAccess/manageRolesColumns.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <ScrollArea className="h-12">
+          <div className="flex flex-wrap italic gap-0.5  text-xs pl-0.5">
+            {permissions?.length === 0
+              ? "None"
+              : permissions.map((permission, index) => (
+                  <div key={permission.id}>
+                    {index === permissions.length - 1
+                      ? constructPermissionName(permission)
+                      : constructPermissionName(permission) + ","}
+                  </div>
```
EmmaLRussell, line 32: Any particular reason to make these separate divs? could just do a map and join on the permissions and put that string into a single element.

EmmaLRussell (commented):
Looks good! Looks like you've separated global from scoped permission so the former come first? I think that's good, and would maybe also be good (in another ticket) to order by scope type (e.g. global, then by tag, then by packet group, then by packet), and put the permissions in each group in alphabetical order. I think it would also help break things up visually just to put a little vertical space between the global and scoped perms groupings. 

> Also have added @JsonProperty to all dtos as this was needed by integration tests so can map result.body to exact types.. eg can do val userResult = ObjectMapper().readValue(result.body, object : TypeReference<UserDto>() {})

Are you sure you need to do this? I thought the default deserialisation would cope pretty happily without annotations if you give it a class reference. https://www.baeldung.com/jackson-object-mapper-tutorial#2-json-to-java-object

absternator:
> Looks good! Looks like you've separated global from scoped permission so the former come first? I think that's good, and would maybe also be good (in another ticket) to order by scope type (e.g. global, then by tag, then by packet group, then by packet), and put the permissions in each group in alphabetical order. I think it would also help break things up visually just to put a little vertical space between the global and scoped perms groupings.
> 
> > Also have added @JsonProperty to all dtos as this was needed by integration tests so can map result.body to exact types.. eg can do val userResult = ObjectMapper().readValue(result.body, object : TypeReference() {})
> 
> Are you sure you need to do this? I thought the default deserialisation would cope pretty happily without annotations if you give it a class reference. https://www.baeldung.com/jackson-object-mapper-tutorial#2-json-to-java-object

1. Sounds good have made ticket for that one
2. Initially found it weird to but its because of kotlin data classes it messes up with serialization... but I found have to use `jacksonObjectMapper()` instead because it registers kotlin module :smile:

app/src/app/components/contents/common/DataTable.tsx
absternator: you are right i have made it so we pass in the page size here now :smile:

app/src/app/components/contents/common/FilterByName.tsx
absternator: yup fully agreee with that too.. have now passed in postfilter action that parent is responsible for

app/src/app/components/contents/common/FilterByName.tsx
absternator: great idea!! have passed in placeholder text and named it `FilterInput`

app/src/app/components/contents/manageAccess/manageRolesColumns.tsx
absternator: the only reason was to control the spacing b/w elements ... especially vertial

EmmaLRussell (approved):
Cool, thanks!

PR #78, opened 2024-05-23
Mrc-5402 Users table

absternator:
This PR works for the users table... All the data is showing and filtering and pagination is working.... No actions are implemented yet (any buttons) these will be in subsequent PRs...

figma--> https://www.figma.com/design/NMR5XdmWYfphrrRYgIV2US/Packit-User-manage?node-id=91-3030&t=j94obNpJPd4DyqpM-0

image

### DIFF
--- a/app/src/app/components/contents/manageAccess/ManageAccessLayout.tsx
+++ b/app/src/app/components/contents/manageAccess/ManageAccessLayout.tsx
@@ -1,8 +1,8 @@
-import { Outlet } from "react-router-dom";
 import { Sidebar } from "../common/Sidebar";
 import { SidebarItem } from "../../../../lib/types/SidebarItem";
 import { useUser } from "../../providers/UserProvider";
 import { Unauthorized } from "../common/Unauthorized";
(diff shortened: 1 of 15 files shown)


M-Kusumgar (approved):
Looks good to me, so just for my understanding, we treat user specific permissions as its own "role" in the backend that only applies to a user, i think it would be nice to have a toggle that just shows all the permissions that user has, because right now you would have to remember what permissions the role "A" has and combine it with user specific permissions if you want the total collections of permissions a user has, obviously not in this PR if you choose to do it but just an idea

Otherwise everything looks good

absternator:
> Looks good to me, so just for my understanding, we treat user specific permissions as its own "role" in the backend that only applies to a user, i think it would be nice to have a toggle that just shows all the permissions that user has, because right now you would have to remember what permissions the role "A" has and combine it with user specific permissions if you want the total collections of permissions a user has, obviously not in this PR if you choose to do it but just an idea
> 
> Otherwise everything looks good

Yup spot on... Yeah that makes so much sense I have made a ticket for that good call

PR #79, opened 2024-05-24
Mrc-5368 Add Role 

absternator:
This PR works for the add role functionality.. Work done 
- Add dialog popup with form to add role (`addRoleButton.tsx`)
- Moved `manageUserColumns.tsx and manageRolesColumns.tsx` to `utils` inside manageAccess folder --- THUS Do not have to review this code

Note: Added components to `Base` are copy pastes form shadcn so don't need review

Testing:
With a user that has user.manage permission... go to mange roles and click add role... The dialog should popup and allow you to create role.... error states: empty name field, duplicate role name
image

### DIFF
--- a/app/src/app/components/contents/manageAccess/AddRoleForm.tsx
+++ b/app/src/app/components/contents/manageAccess/AddRoleForm.tsx
@@ -26,7 +26,10 @@ export const AddRoleForm = ({ mutate, setOpen }: AddRoleFormProps) => {
     name: z
       .string()
       .min(1)
-      .regex(/^\S.*\S$/, "Name must not start or end with whitespace"),
+      .regex(
+        /^[a-zA-Z0-9]+(\s[a-zA-Z0-9]+)*$/,
+        "Name must only contain alphanumeric characters & no leading/trailing spaces."
+      ),
     permissionNames: z.array(z.string())
   });
   const form = useForm<z.infer<typeof formSchema>>({
--- a/app/src/app/components/contents/manageAccess/AddRoleForm.tsx
+++ b/app/src/app/components/contents/manageAccess/AddRoleForm.tsx
@@ -1,0 +1,122 @@
+import { zodResolver } from "@hookform/resolvers/zod";
+import { Dispatch, SetStateAction, useState } from "react";
+import { useForm } from "react-hook-form";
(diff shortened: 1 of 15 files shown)


app/src/lib/constants.ts
```diff
@@ -1 +1,3 @@
 export const PAGE_SIZE = 50;
+
+export const GLOBAL_PERMISSIONS = ["packet.push", "packet.read", "packet.run", "user.manage"] as const;
```
EmmaLRussell, line 3: I don't really think we should make these constants - if we add new permissions, we'd want the front end to pick them up automatically. So we should make an endpoint for those. One for another ticket though, 

For future functionality - when we support adding scoped permissions to users and roles - we should also save and return with the permission definition whether they are global only (like user.manage) or scope-able (like packet.push).

EmmaLRussell (commented):
Looks great!

I think you'd better trim the user-provided names - it let me make a role called "ADMIN " which looks identical to ADMIN in the list view...

app/src/lib/constants.ts
absternator: yup 100%.. have created ticket for that one

M-Kusumgar (approved):
Nice, works like a charm

M-Kusumgar (changes requested):
Just realised while playing around with the adding user PR, we should not allow commas in the role names, otherwise it gets confusing as we use commas to separate out the roles

app/src/app/components/contents/manageAccess/AddRoleForm.tsx
```diff
@@ -23,7 +23,13 @@ export const AddRoleForm = ({ mutate, setOpen }: AddRoleFormProps) => {
...
 
   const formSchema = z.object({
-    name: z.string().min(1),
+    name: z
+      .string()
+      .min(1)
+      .regex(
+        /^[a-zA-Z0-9]+(\s[a-zA-Z0-9]+)*$/,
+        "Name must only contain alphanumeric characters & no leading/trailing spaces."
+      ),
```
EmmaLRussell, line 32: This might be too restrictive? But I think it's fair to encourage sensible role names!

EmmaLRussell (approved):

PR #80, opened 2024-05-29
Mrc 5369 Add basic user

absternator:
This PR contains functionality to add basic user... Remeber code in `/Base/..` is copy pasted so no need to review.... The main thing to review Is the actual add user button and dialog.

Testing: 
1. create some roles with Add role button on Manage Roles page
2. Try create some users by clicking add user button and fill out form... play around with invalid states (email,password,display name validation)

image
Note: 1 change from figma is the role adding... using a multi-select as more intuitative
Figma: https://www.figma.com/design/NMR5XdmWYfphrrRYgIV2US/Packit-User-manage?node-id=91-3030&t=WyF73xO0Xnhs18rM-0

### DIFF
--- a/app/src/app/components/contents/manageAccess/AddUserForm.tsx
+++ b/app/src/app/components/contents/manageAccess/AddUserForm.tsx
@@ -1,0 +1,143 @@
+import { zodResolver } from "@hookform/resolvers/zod";
+import { Dispatch, SetStateAction, useState } from "react";
+import { useForm } from "react-hook-form";
+import { KeyedMutator } from "swr";
+import { z } from "zod";
+import appConfig from "../../../../config/appConfig";
+import { ApiError } from "../../../../lib/errors";
+import { fetcher } from "../../../../lib/fetch";
+import { Button } from "../../Base/Button";
+import { DialogClose, DialogFooter } from "../../Base/Dialog";
+import { Form, FormControl, FormDescription, FormField, FormItem, FormLabel, FormMessage } from "../../Base/Form";
+import { Input } from "../../Base/Input";
+import { RoleWithRelationships } from "./types/RoleWithRelationships";
+import {
+  MultiSelector,
+  MultiSelectorContent,
+  MultiSelectorInput,
+  MultiSelectorItem,
+  MultiSelectorList,
+  MultiSelectorTrigger
+} from "../../Base/MultiSelect";
+
+interface AddUserFormProps {
+  mutate: KeyedMutator<RoleWithRelationships[]>;
+  setOpen: Dispatch<SetStateAction<boolean>>;
+  roleNames: string[];
+}
+export const AddUserForm = ({ mutate, setOpen, roleNames }: AddUserFormProps) => {
+  const [fetchError, setFetchError] = useState("");
+
+  const formSchema = z.object({
+    email: z.string().email(),
+    displayName: z.string().min(1),
+    password: z.string().min(8),
+    userRoles: z.array(z.string())
+  });
+  const form = useForm<z.infer<typeof formSchema>>({
+    resolver: zodResolver(formSchema),
+    defaultValues: {
+      email: "",
+      displayName: "",
+      password: "",
+      userRoles: []
+    }
+  });
+  const onSubmit = async (values: z.infer<typeof formSchema>) => {
+    try {
+      await fetcher({
+        url: `${appConfig.apiUrl()}/user/basic`,
+        body: values,
+        method: "POST"
+      });
+      form.reset();
+      setOpen(false);
+      mutate();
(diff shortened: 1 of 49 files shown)


app/src/app/components/contents/manageAccess/AddUserForm.tsx
```diff
@@ -0,0 +1,144 @@
...
+        body: values,
+        method: "POST"
+      });
+      form.reset();
+      setOpen(false);
+      mutate();
+    } catch (error) {
+      console.error(error);
+      if (error instanceof ApiError && error.status === HttpStatus.BadRequest) {
+        return form.setError("email", { message: error.message });
```
EmmaLRussell, line 60: Is this assuming that if you get a BadRequest resons it's because something was wrong with the email address (presumably because it already exists?)

EmmaLRussell (commented):
Looks really good. Yeah, being able to add roles from the create user dialog is nice!

It does make me wonder if you should also be able to add permissions too - but I think that would be too complicated, as that's going to get quite fiddly with the fine grained perms. But might be a bit confusing to the user that they can add roles here but nor perms. Maybe include a little note on the dialog something like "You can add user-specific permissions after the user has been created."

Tiny point about consistency - for both Role and User, the buttons say "Add.." but the dialogs say "Create new.. " (and Create on the button). Could probably just use "Add.." for both .

I've also just noticed that for both dialogs, if you click outside on the overlay, it cancels out of the dialog, which is potentially annoying!

Also, it feels like clicking on the "Select roles" area should close the multi-select. And maybe the Enter key as well?  Because the multi-select list covers up the Create button, the only way to actually create the user having selected roles is to click on some other part of the dialog (being careful not to click outside on the overlay and accidentally cancel) to re-expose the button... 

Not sure how hard that would be to fix in the imported components though!

app/src/app/components/contents/manageAccess/AddUserForm.tsx
absternator: yup its the only way to get bad request from this form

absternator:
> Looks really good. Yeah, being able to add roles from the create user dialog is nice!
> 
> It does make me wonder if you should also be able to add permissions too - but I think that would be too complicated, as that's going to get quite fiddly with the fine grained perms. But might be a bit confusing to the user that they can add roles here but not perms. Maybe include a little note on the dialog something like "You can add user-specific permissions after the user has been created."
> 
> Tiny point about consistency - for both Role and User, the buttons say "Add.." but the dialogs say "Create new.. " (and Create on the button). Could probably just use "Add.." for both .
> 
> I've also just noticed that for both dialogs, if you click outside on the overlay, it cancels out of the dialog, which is potentially annoying!
> 
> Also, it feels like clicking on the "Select roles" area should close the multi-select. And maybe the Enter key as well? Because the multi-select list covers up the Create button, the only way to actually create the user having selected roles is to click on some other part of the dialog (being careful not to click outside on the overlay and accidentally cancel) to re-expose the button...
> 
> Not sure how hard that would be to fix in the imported components though!

1. have added that help text for adding permissions later
2. made it so click outside doesn't close dialog.... This should help with the problem of closing the multi select too.... Yeah tried but its just an imported component and would be too hard ... i can try make ticket if needed to dig more

M-Kusumgar (commented):
Works great, liking the look of it too, i think there should be the same check as the previous PR with whitespace at beginning or end of the display name, currently can give someone a display name with just a space

EmmaLRussell (commented):
Thanks, that's a lot better without the peril of accidentally cancelling. 

We shouldn't limit to a-z alphanumeric for display names as lots of people have accented letters in their names, e.g. François or Seán

Sorry, I just noticed another thing, which is that the user id is called "Username" in the table and "Email" in the dialog - should probably be consistent there. Also, shouldn't we show Display Name in the table too?

Still have the Create vs Add inconsistency too.

absternator:
> Thanks, that's a lot better without the peril of accidentally cancelling.
> 
> We shouldn't limit to a-z alphanumeric for display names as lots of people have accented letters in their names, e.g. François or Seán
> 
> Sorry, I just noticed another thing, which is that the user id is called "Username" in the table and "Email" in the dialog - should probably be consistent there. Also, shouldn't we show Display Name in the table too?
> 
> Still have the Create vs Add inconsistency too.

1. agree and have reverted this and just fixed leading/trailing spaces
2. Its intended as the table can be github users as well... but for the dialog its only for basic and we email... Also display name is not a required field (github users may not have) but I can add later?
3. fixed the Create vs add consitency

EmmaLRussell (approved):
Yeah fair point re username vs email - I guess most people won't really notice!

PR #81, opened 2024-05-29
Mrc-5370 Delete user and role functionality

absternator:
The following add functionality to delete users and roles.

Testing: 
- click on trash icon and confirm the popup and ensure deleting of roles
- - click on trash icon and confirm the popup and ensure deleting of users

image

image

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@hookform/resolvers": "^3.3.4",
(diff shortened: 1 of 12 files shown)


absternator:
> Think this works well, and the messages are clear. A weird thing was when I tried to delete myself (super user) from the users and then nothing happened, either we should sign them out or we should disable deleting yourself from the manage users page i think

good catch didn't even think about that.. I've made it so user cannot delete themselves, seems like most appropriate behaviour

M-Kusumgar (approved):
Looks good to me!

PR #82, opened 2024-05-30
Mrc 5371-  manage roles update users

absternator:
This PR adds functionality to add users to specific roles. On manage roles page, the 3 dots can be clicked and add user option selected. This will show a form that allows deleting and adding of users... 

Also made the read of roles sort the list of users for each role... this makes it easier to search and find particular users

Testing:
Try add/remove users from roles and ensure correct users as a result on role

Note: as per role adding, have used a multi select for cleanest approach over initial figma.

https://www.figma.com/design/NMR5XdmWYfphrrRYgIV2US/Packit-User-manage?node-id=0-1&t=6bOarH6k4cZX3I7E-0

image

image


M-Kusumgar (commented):
All looks good for this one. One comment about the styling is that when there are too many permissions or user emails, the table implements a scroll but only on hover. Sometimes it isnt immediately obvious that there is more (if the cut off is aligned with the bottom of the text) perhaps some kind of indication that there is more if you scroll on the table? maybe the scrollbar is always visible but more subtle if you dont hover, something like that

A bigger problem i realised that you can fix in another PR maybe or these ones is we need stricter validation on certain things like the role name. There are a couple places in the app where you use `useSWR` and fetch stuff and use a variable the user can freely define in the url string, if they put a `/` in the role name they will get an error as they are trying to access another path. If they put a `;` in the role name, they will also get an error but this time a CORS error. we should probably limit them to very specific special characters if we even allow special characters! There was another url with `packetName` in it, not sure if that can also be general and contain `/` or `;` but yh would be good to have a look at this for all the urls we create

absternator:
> All looks good for this one. One comment about the styling is that when there are too many permissions or user emails, the table implements a scroll but only on hover. Sometimes it isnt immediately obvious that there is more (if the cut off is aligned with the bottom of the text) perhaps some kind of indication that there is more if you scroll on the table? maybe the scrollbar is always visible but more subtle if you dont hover, something like that
> 
> A bigger problem i realised that you can fix in another PR maybe or these ones is we need stricter validation on certain things like the role name. There are a couple places in the app where you use `useSWR` and fetch stuff and use a variable the user can freely define in the url string, if they put a `/` in the role name they will get an error as they are trying to access another path. If they put a `;` in the role name, they will also get an error but this time a CORS error. we should probably limit them to very specific special characters if we even allow special characters! There was another url with `packetName` in it, not sure if that can also be general and contain `/` or `;` but yh would be good to have a look at this for all the urls we create
> 
> Not as fussed about this last point but i think there is still value in showing the user the list of users in a particular role in addition to the add and remove, instead of it just filtering users in add and remove depending on whether they are in the role or not

1. good idea I tried the cutoff thing but was hard because of size of box... so I've made scroll bar always visible if overflow 
2. Yup have made it not allow special characters (Role name)... and if users are going into the URL and messing with it they shouldn't complain about unexpected behaviour as that is outside of how the app should be used
3. Im not sure because you can already see it in the table... thus the extra info on the add/remove will make it more clouded and not be concise

EmmaLRussell (commented):
Multi-select seems like a good interface to me

We should protect the permissions on the Admin role, so "Update Permissions" shouldn't be visible, or should be disabled, for that role...but can do that in the next ticket. 

I'm not sure the "Actions" heading in the dropdown from the dots is useful, think we could probably do without that. 

Do we have a ticket for preventing the admin user from removing themselves from the admin role? I think we talked about it and maybe said we'd allow it but I think we should at least warn them!

absternator:
> Multi-select seems like a good interface to me
> 
> We should protect the permissions on the Admin role, so "Update Permissions" shouldn't be visible, or should be disabled, for that role...but can do that in the next ticket.
> 
> I'm not sure the "Actions" heading in the dropdown from the dots is useful, think we could probably do without that.
> 
> Do we have a ticket for preventing the admin user from removing themselves from the admin role? I think we talked about it and maybe said we'd allow it but I think we should at least warn them!

1. yup already done update permissions pr 
2. removed action heading 
3. yup will make ticket for those

EmmaLRussell (approved):

PR #83, opened 2024-05-31
mrc-5428 add dev-start script

M-Kusumgar:
The aim of this was just to streamline setup for packit, felt like it could've just been one command so that is what I have done here

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,32 +3,50 @@
 Web app for serving outpack and/or orderly2 metadata.
 
 ## Pre-requisites
-Ensure you have logged into vault. If not,  Run the following command to login with github PAT token:
+
+Ensure you have logged into vault. If not, Run the following command to login with github PAT token:
+
 ```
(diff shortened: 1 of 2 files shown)


absternator (approved):
legend lgtm

EmmaLRussell (approved):
This is cool - I guess this is going to be most useful when you're revieiwing, or when working on the front end, as it isn't very conducive to tearing down and rerunning the backend container. But if you were doing that you'd probably just be running the backend with gradle anyway.

EmmaLRussell (commented):
Might also be nice to include the auth method as as option!

M-Kusumgar:
> Might also be nice to include the auth method as as option!

will add this in this ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5435/Add-auth-method-in-dev-start-script mainly because im not actually sure what possible values for the auth methods there are and also if i have to change anything other than the auth in the app properties

PR #85, opened 2024-05-31
Mrc 5430 better GitHub actions

M-Kusumgar:
github actions kept failing every first push because building the db image is one of the actions and the backend tests require that to run `./scripts/run-dependencies` so I am just building the db, similarly the api build is required in frontend tests but the build only  gets pushed when the backend tests finish so am building api in frontend tests too

### DIFF
--- a/.github/actions/build-api/action.yml
+++ b/.github/actions/build-api/action.yml
@@ -4,5 +4,6 @@ runs:
   using: "composite"
   steps:
     - name: Build API
+      shell: bash
       run: |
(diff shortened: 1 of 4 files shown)


absternator (approved):
THANKS FOR SAVING THIS!!!

EmmaLRussell (approved):
Nice one. I suppose you could push the image early, which would save building it multiple times. But maybe want to avoid pushing until tests are passing.

PR #86, opened 2024-05-31
Mrc 5372-manage users update roles

absternator:
This PR is similar to #82 but is on the manage roles page and is the option to update roles on users... It also refactors some code as there was duplicate code that was exactracted

Testing:
On manage users page, click on 3 dots on row and try click update roles and try update roles. should work as expected
image

### DIFF
--- a/.github/actions/build-api/action.yml
+++ b/.github/actions/build-api/action.yml
@@ -1,0 +1,9 @@
+name: Build API
+
+runs:
+  using: "composite"
+  steps:
+    - name: Build API
+      shell: bash
+      run: |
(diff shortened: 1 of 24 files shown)


M-Kusumgar (commented):
I think this works as expected but one weird thing was:
* As the super user, I created a role without user manage permission
* I updated my role to that one and removed admin role
* I can still manage users and add roles and so on after that

Perhaps we should keep local storage up to date (I think that is how you're checking stuff) and do a page refresh if someting like that happens. Either way I think this should be reactive

PR #88, opened 2024-06-12
Mrc 5364- Global + single resource auth

absternator:
This ticket 'enables' auht on packet controller and fine grained on single resource ones. The following is done 
- Show unauthorized if no packet.read permission. This is for all packet groups page, and packets for each packet group
- Allow specific permission to view specific packets and their reports and download files

Test urls: 
1. http://localhost:3000/
2. http://localhost:3000/parameters/
3. http://localhost:3000/parameters/20231205-073715-6635e044/ 

Testing: 
1. Create user with no packet.read... then login and ensure cant see unauthorized in above urls
2. Create user with packet.read... then login and ensure can see all URLs and be able to see report and download file 
3. As admin Give the user the permission `packet.read:packetGroup:paramaters` (via the update permissions form in user manage).. Then login as user and ensure can see http://localhost:3000/parameters/20231205-073715-6635e044/ only and also be able to see report and download files 
4. Do same as above but instead remove the above permission and add `packet.read:packet:20231205-073715-6635e044`

### DIFF
--- a/api/app/src/main/kotlin/packit/controllers/PacketController.kt
+++ b/api/app/src/main/kotlin/packit/controllers/PacketController.kt
@@ -56,17 +56,18 @@ class PacketController(private val packetService: PacketService)
     }
 
     @GetMapping("/metadata/{id}")
-    @PreAuthorize("hasAuthority('packet.read')")
+    @PreAuthorize("hasAuthority('packet.read') or @authz.canReadPacketMetadata(#root, #id)")
     fun findPacketMetadata(@PathVariable id: String): ResponseEntity<PacketMetadata>
     {
         return ResponseEntity.ok(packetService.getMetadataBy(id))
     }
 
-    @GetMapping("/file/{hash}")
-    @PreAuthorize("hasAuthority('packet.read')")
+    @GetMapping("/file/{id}")
+    @PreAuthorize("hasAuthority('packet.read') or @authz.canReadPacketMetadata(#root, #id)")
     @ResponseBody
     fun findFile(
-        @PathVariable hash: String,
+        @PathVariable id: String,
+        @RequestParam hash: String,
         @RequestParam inline: Boolean = false,
         @RequestParam filename: String,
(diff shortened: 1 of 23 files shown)


M-Kusumgar (approved):
LGTM! thanks for the explanations for spring security!

EmmaLRussell (commented):
Just a super-quick look at this as I ran out of time on Friday! Looks really tidy, so go ahead and merge. Just making note of a couple of questions we can talk about once I'm back if that's ok. 

So, it looks to me like this means that you currently need to have global read scope to list packets, is that right? That's what `hasAuthority` in spring's built-in `Pre-Authorize` gives you?  But for packit to work we're going to need to get the list of packets that a user has access to in order to use the app, even if they don't have global read perm, So we are going to have return a listed filtered by the user's actual scoped perms. I guess that's happening in another ticket? Probably goes for a lot of the other endpoint too, e.g. packetGroupSummary - though to be honest I can't exactly remember which endpoint is used where - but I think most if not all of them are going to need to work for non-global scope users. 

Also assuming protection of manage access endpoints happening in another ticket.

absternator:
> Just a super-quick look at this as I ran out of time on Friday! Looks really tidy, so go ahead and merge. Just making note of a couple of questions we can talk about once I'm back if that's ok.
> 
> So, it looks to me like this means that you currently need to have global read scope to list packets, is that right? That's what `hasAuthority` in spring's built-in `Pre-Authorize` gives you? But for packit to work we're going to need to get the list of packets that a user has access to in order to use the app, even if they don't have global read perm, So we are going to have return a listed filtered by the user's actual scoped perms. I guess that's happening in another ticket? Probably goes for a lot of the other endpoint too, e.g. packetGroupSummary - though to be honest I can't exactly remember which endpoint is used where - but I think most if not all of them are going to need to work for non-global scope users.
> 
> Also assuming protection of manage access endpoints happening in another ticket.

yup sorry the list endpoints are done in the following PR :smile: ... Also manage access already protected

PR #89, opened 2024-06-12
Mrc-5365 - Enable auth for packet read lists

absternator:
This ticket is focused on enabling auth for packet lists + some other updates. Below is the work done 
- update packet permission scope to add name and for update permission for packet scope filter by name & id
- Filter out lists to ensure users only see what they are allowed. These rules are in `AuthorizationLogic.kt` and the filtering annotation is in `PacketRepository.kt` and `PacketGroupRepository.kt`
- As per #87 comment.. have moved /packetGroup endpoint in PacketController to its own PacketGroupController and Service 

Testing Urls: 
1. http://localhost:3000/
2. http://localhost:3000/parameters/
3.  http://localhost:3000/parameters/20231205-073715-6635e044/

Test cases : 
1. First URL, only be able to see packet groups user has permission to see... (if has packetGroup name or packet with that name) or see all if packet.read 
2. Second Url try give packetGroup:parameters permission.. Should see all parameters... Then try just parameters:20231205-073715-6635e044, then will only be able to see that packet in the list 
3. When editing permissions ensure same rules apply as above when adding scoped permission (lists for packet and packet group)


M-Kusumgar (approved):
LGTM, just noting a discussion had in person, used can call their packet groups whatever they want and this can mess up the urls (e.g. if they include a `;` in their packet group names) so we should use unicode like google does for search results to avoid this problem

PR #91, opened 2024-07-09
Allow API URL to be overriden at build time.

plietar:
The API URL is now defined using a `REACT_APP_PACKIT_API_URL` environment variable. The variable defaults to "http://localhost:8080" for development or "$PUBLIC_URL/packit/api" for production build.  The production build definition enables the common use case of placing both the frontend and API under the same path.

The default values are set in the dotenv files. These can be overriden locally by creating a `.env.local`, or simply setting an environment variable during the build.

This flexibility will allow us to run Packit at different locations, for a multi-tenant deployment.


absternator (approved):
LGTM

PR #92, opened 2024-07-09
Set the router's basepath to PUBLIC_URL.

plietar:
This is needed in order to host Packit on a sub-path. This fixes both the routing, allowing the right pages to be rendered, as well as all links within the application. React Router will automatically add the basepath as prefixes to any `Link` used in the application.


absternator (approved):
LGTM

PR #93, opened 2024-07-09
Remove practice submodule.

plietar:
It's not clear to me why it exists, and it has no defined URL. It causes issues with tooling that tries to init and update submodules by default.


absternator (commented):
LGTM

M-Kusumgar (approved):

PR #95, opened 2024-07-24
Mrc 5513 namespace local storage

M-Kusumgar:
Namespacing local storage keys so that we can have multi tenant packit instances. I have not namespaced the ui theme because i think this should be shared amongst all instances

Note: Opportunistic update of node version when running ci from 16 to 20


app/src/tests/appConfig.test.ts
```diff
@@ -28,5 +28,21 @@ describe("api service", () => {
         /* eslint-enable */
         expect(appConfig.apiUrl).toThrow();
     });
+
+    test("uses PACKET NAMESPACE from environment", () => {
```
plietar, line 32:
```suggestion
    test("uses PACKIT NAMESPACE from environment", () => {
```

app/src/lib/types/LocalStorageKeys.ts
```diff
@@ -1,6 +1,13 @@
+import appConfig from "../../config/appConfig";
+
+const getLocalStorageKey = (key: string) => {
+  const ns = appConfig.appNamespace();
+  return ns ? `${ns}-${key}` : key;
```
plietar, line 5: The ns and the key are likely to contain some themselves, and then we might end up with something like `priority-pathogen-authConfig`. It wouldn’t call any problems per se, but best to avoid I think.

A dot or colon or slash would be better as a separator I think.

EmmaLRussell: We should be using camelCase for the keys, but agree that dot is probably better as ns might include dashes, if they're going to match the url part.

app/src/tests/appConfig.test.ts
```diff
@@ -28,5 +28,21 @@ describe("api service", () => {
...
+        process.env.REACT_APP_PACKIT_NAMESPACE = "my-repo";
+        /* eslint-disable */
+        const appConfig = require("../config/appConfig").default;
+        /* eslint-enable */
+        expect(appConfig.appNamespace()).toBe("my-repo");
+    });
+
+    test("returns null when PACKET NAMESPACE is missing", () => {
```
EmmaLRussell, line 40:
```suggestion
    test("returns null when PACKIT NAMESPACE is missing", () => {
```
This one too I guess!

app/src/lib/types/LocalStorageKeys.ts
M-Kusumgar: good point, have gone with . instead

EmmaLRussell (approved):

plietar (approved):

PR #97, opened 2024-07-31
Mrc 5564 outpack update metadata

M-Kusumgar:
This just updates the metadata data classes to more closely resemble the [outpack json schema](https://github.com/mrc-ide/outpack/blob/main/schema/metadata.json). You can go on the link to the schema and compare with the kotlin data classes. I am actually not sure where this is tested. I don't think this actually is as far as I can tell which is fair enough as this is just the data class definitions

This update was required because you would get the error below if you tried to look at the summary of the packet that was run with a detached head (so not on any branch and so the git.branch would be `null` in the metadata) and the frontend would receive this error:
```json
{
    "error": {
        "detail": "Error while extracting response for type [packit.model.OutpackResponse<packit.model.PacketMetadata>] and content type [application/json]",
        "error": "OTHER_ERROR"
    },
    "data": {},
    "status": "failure"
}
```
i don't think i can test the case when we have git.branch as `null` because that requires me to do a detached checkout in packit which sounds messy, we will know pretty quickly if this works to solve the error if we deploy to packit.dide.ic.ac.uk and the error goes away


EmmaLRussell (approved):
> i don't think i can test the case when we have git.branch as null because that requires me to do a detached checkout in packit which sounds messy

yeah, I'm not sure what it would take to persuade orderly that one of the demo packets was detached!

Updated types look fine to me. Bit confused about the demo report change. It looks like there was already a demo report which specified a dependency - but that was for "orderly3"??

M-Kusumgar:
> yeah, I'm not sure what it would take to persuade orderly that one of the demo packets was detached!

thinking a bit more about it, we can test it manually, would have to exec into docker with outpack running, initialise `/outpack` in the docker container as a git repo, create a commit and check it out in detached mode, but yh would be super messy to do in code

> Updated types look fine to me. Bit confused about the demo report change. It looks like there was already a demo report which specified a dependency - but that was for "orderly3"??

My suspicion is that someone generated the packets and then maybe accidentally changed it to orderly3 for some reason, i have no idea what happened there :laughing:

absternator (approved):
LGTM

PR #98, opened 2024-08-12
mrc-5601 wire up orderly.runner to kotlin api

M-Kusumgar:
This PR hooks up the runner to the packit api. To test you can run up the app as normal and fetch `/runner/version` in the app. Main changes include:
1. `autoCorrect` in detekt task so that we can easily correct linting errors
2. `orderlyRunnerUrl` env var
3. controller, service, client for runner
4. scripts to setup all the docker container (this bit might be meaty without context so we can have a chat about it)

Note to self: remove any instance of `mrc-5146-container-gh-actions` once the PR has been merged to main


api/app/build.gradle.kts
```diff
@@ -89,6 +89,7 @@ detekt {
     config =
         files("$projectDir/config/detekt.yml") // point to your custom config defining rules to run, overwriting default behavior
     baseline = file("$projectDir/config/baseline.xml") // a way of suppressing issues before introducing detekt
+    autoCorrect = true
```
M-Kusumgar, line 92: this should be alright to leave in CI as well. The task does fail even if it auto corrects and if we run this, it'll fix linting issues wherever it can

scripts/run-dependencies
```diff
@@ -9,9 +9,59 @@ HERE=$(realpath "$(dirname $0)")
 
 docker exec packit-db wait-for-db
 
-# Start outpack_server
-docker pull $OUTPACK_SERVER_TAG
-docker run --name outpack_server \
-  -v $HERE/../demos/orderly:/outpack \
-  -p 8000:8000 -d --rm \
-  $OUTPACK_SERVER_TAG
+docker network create $NETWORK
```
absternator, line 12: could put all this into a orderly.runner script that gets called... can then also take all env vars out of common and use thre also

scripts/common
```diff
@@ -1,6 +1,24 @@
...
 
 # Export env vars needed for running test dependencies
 export ORG=mrcide
-export OUTPACK_SERVER_TAG=$ORG/outpack_server:main
+export OUTPACK_SERVER_TAG=$ORG/outpack_server:497d8b1
+
+export CONTAINER_NAMESPACE=orderly.runner
+
+export NETWORK=$CONTAINER_NAMESPACE-network
+export REDIS=$CONTAINER_NAMESPACE-redis
```
absternator, line 11: this isnt really common right its specific to runner?...as per comment on `run-dependencies` this can be moved to script that spins up orderly.runner stuff....

scripts/clear-docker
```diff
@@ -1,6 +1,11 @@
...
 set -euxo pipefail
 
+HERE=$(realpath "$(dirname $0)")
+. "$HERE"/common
+
 docker rm --force $(docker ps --all --quiet) || true
 docker network prune --force
 docker volume prune --force
+
+docker volume rm $ORDERLY_VOLUME $ORDERLY_REMOTE_VOLUME
```
absternator, line 11: probably best not to remove every docker container and network and volume.... if people are working on other stuff its gets removed too ahha.. i learnt this the hard way by running this hahaha

absternator (commented):
Nice work and works like a charm!!! just left some commnets

scripts/run-dependencies
absternator: can we use just the default bridge network?

api/app/src/main/kotlin/packit/controllers/RunnerController.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+import org.springframework.web.bind.annotation.RestController
+import packit.model.dto.OrderlyRunnerVersion
+import packit.service.RunnerService
+
+@RestController
+@RequestMapping("/runner")
```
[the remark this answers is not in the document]
absternator: ignore ive done in consequent pr

api/app/src/main/kotlin/packit/service/OrderlyRunnerClient.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    override fun getVersion(): OrderlyRunnerVersion
+    {
+        return getEndpoint("/")
+    }
+
+    private inline fun <reified T> getEndpoint(urlFragment: String): T
+    {
+        val url = "$baseUrl/$urlFragment"
+        log.debug("Fetching {}", url)
```
EmmaLRussell, line 32: Rogue debug?

api/app/src/main/kotlin/packit/service/RunnerService.kt
```diff
@@ -0,0 +1,20 @@
...
+import packit.model.dto.OrderlyRunnerVersion
+
+interface RunnerService
+{
+    fun getVersion(): OrderlyRunnerVersion
+}
+
+@Service
+class BaseRunnerService(
+    private val orderlyRunnerClient: OrderlyRunnerClient
```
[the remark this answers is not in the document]
absternator: ignore will resolve

api/app/src/main/kotlin/packit/service/OrderlyRunnerClient.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.core.ParameterizedTypeReference
+import org.springframework.http.*
+import org.springframework.stereotype.Service
+import org.springframework.web.client.RestTemplate
+import packit.AppConfig
+import packit.exceptions.PackitException
+import packit.model.ServerResponse
+import packit.model.dto.OrderlyRunnerVersion
+
```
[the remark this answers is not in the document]
absternator: ignore will resolve

api/app/src/main/kotlin/packit/service/OrderlyRunnerClient.kt
```diff
@@ -0,0 +1,63 @@
...
+interface OrderlyRunner
+{
+    fun getVersion(): OrderlyRunnerVersion
+}
+
+@Service
+class OrderlyRunnerClient(appConfig: AppConfig) : OrderlyRunner
```
EmmaLRussell, line 19: There's a lot of shared code between this client and the outpack client isn't there? in terms of Anmol's comment about error handling etc, and dealing with standard Porcelain Server responses (which we should be getting from both Outpack and Orderly), it feels like that could be delegated to a generic client class, and then the orderly client and outpack clients could provide methods for their various offerings (e.g. `getVersion`) which would invoke the correct endpoint via the generic client.

scripts/run-dependencies
```diff
@@ -9,9 +9,59 @@ HERE=$(realpath "$(dirname $0)")
...
+
+docker run --rm -d --pull=always \
+       --net=$NETWORK \
+       --name=$WORKER \
+       --entrypoint="/usr/local/bin/orderly.runner.worker" \
+       --env=ORDERLY_RUNNER_QUEUE_ID=$ORDERLY_RUNNER_QUEUE_ID \
+       --env=REDIS_CONTAINER_NAME=$REDIS \
+       -v $ORDERLY_VOLUME:$CONTAINER_ORDERLY_ROOT_PATH \
+       $ORDERLY_RUNNER_IMAGE \
+       $CONTAINER_ORDERLY_ROOT_PATH
```
EmmaLRussell, line 67: I'm not too bothered about whether this goes into a separate script or not, but I think it would be really nice to put in a few comments about what all these various docker interactions are doing!

scripts/common
```diff
@@ -1,6 +1,24 @@
...
+export DEBUG=$CONTAINER_NAMESPACE-debug
```
EmmaLRussell, line 12: So this is the docker container where you're going to do your git setup? I find `DEBUG` quite a confusing suffix for that!
```suggestion
export GIT_SETUP=$CONTAINER_NAMESPACE-git-setup
```

EmmaLRussell (commented):
Looks good! I agree with everything Anmol has said, and also...

api/app/src/main/kotlin/packit/controllers/RunnerController.kt
```diff
@@ -0,0 +1,21 @@
...
+class RunnerController(private val runnerService: RunnerService)
+{
+    @GetMapping("/version")
+    @ResponseBody
```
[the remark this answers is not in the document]
M-Kusumgar: leaving this for you in the next PR

scripts/run-dependencies
M-Kusumgar: i would prefer to keep it in the run dependencies script, i think its okay if theres a couple more containers being run in the same script, i think its more annoying and time consuming to have lots of referencing of other scripts to work out whats going on

default bridge network doesnt work, the server and workers cannot find the redis connection, i think it has to be a separate network

scripts/common
M-Kusumgar: similar to above i would say its nice to have one place for these scripts, like i can just look here and see this is all the names of the containers or something like that, its not a long script, i think whitespace could be a sufficient separator for these concepts, i also would vote for just moving the db/scripts/run into here, it seems annoying to navigate 3 folder levels away to find a single script that just does a docker run, i think we would be at a risk of over abstracting where its not needed

scripts/clear-docker
M-Kusumgar: yep fair enough, can modify it so it removes just the containers we spin up

api/app/src/main/kotlin/packit/service/OrderlyRunnerClient.kt
M-Kusumgar: just gonna tag anmol in the kotlin related comments so he doesnt miss them, @absternator

api/app/src/main/kotlin/packit/service/OrderlyRunnerClient.kt
M-Kusumgar: @absternator

scripts/common
M-Kusumgar: done

scripts/run-dependencies
M-Kusumgar: done

absternator (approved):
LGTM!!! il sort the kotlin stuff in another PR..... also may want to get outpack server and orderly merged first then can update `common` to point to main branches .....

PR #103, opened 2024-08-20
Mrc 5627 runner table

M-Kusumgar:


api/app/src/main/kotlin/packit/model/RunInfo.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Id
+    var taskId: String,
+
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "packet_group_id", nullable = false)
+    var packetGroup: PacketGroup,
+
+    var status: String? = null,
+    var commitHash: String? = null,
+    var branch: String? = null,
```
absternator, line 21: should these be not null ?

api/app/src/main/resources/db/migration/V20240820__initial_run_info.sql
absternator, line 1: run_info_table or create_run_info maybe more suitble names.... initial is a bit off putting here

api/app/src/main/resources/db/migration/V20240820__initial_run_info.sql
```diff
@@ -0,0 +1,16 @@
+-- Create run info table
+CREATE TABLE IF NOT EXISTS "run_info"
+(
+    "task_id"             TEXT PRIMARY KEY NOT NULL,
+    "status"              TEXT,
+    "commit_hash"         TEXT,
+    "branch"              TEXT,
+    "logs"                TEXT,
```
absternator, line 8: should this be json as well? since its stored that in in the entity?

api/app/src/main/kotlin/packit/model/dto/RunInfoDto.kt
```diff
@@ -0,0 +1,15 @@
+package packit.model.dto
+
+data class RunInfoDto(
+    val taskId: String,
+    val packetGroup: PacketGroupDto,
+    val status: String? = null,
+    val commitHash: String? = null,
+    val branch: String? = null,
```
absternator, line 8: same comment are these non nullable?

api/app/src/main/kotlin/packit/model/dto/BasicRunInfoDto.kt
absternator, line 1: wait are BasicRunInfoDto and RunInfoDto the same? can delete one if thats the case

absternator (commented):
Nice work looks really good just a couple comments / questions.. also probs need to pull main back in

1. how does it work when you first create a packetGroup/report in orderly? Do they have to run a packet first? Leading on from that can there be a scenario where we can do the first run of a packet from the web? in that case packetGroup in the table wont exist as its derived from packets we receive....... If that is the case maybe best to unlink the reference to packetGroup.
2.

M-Kusumgar:
> Nice work looks really good just a couple comments / questions.. also probs need to pull main back in
> 
> 1. how does it work when you first create a packetGroup/report in orderly? Do they have to run a packet first? Leading on from that can there be a scenario where we can do the first run of a packet from the web? in that case packetGroup in the table wont exist as its derived from packets we receive....... If that is the case maybe best to unlink the reference to packetGroup.

i thought it would just initialise it as an empty mutable list no? but the only reason i put it there is cos i thought it was the proper way to do it, but im more than happy to unlink

api/app/src/main/kotlin/packit/model/RunInfo.kt
M-Kusumgar: yep youre right, done

api/app/src/main/resources/db/migration/V20240820__initial_run_info.sql
M-Kusumgar: yep changed to run info table

api/app/src/main/resources/db/migration/V20240820__initial_run_info.sql
M-Kusumgar: yep, good spot, thanks

api/app/src/main/kotlin/packit/model/dto/RunInfoDto.kt
M-Kusumgar: done

api/app/src/main/kotlin/packit/model/dto/BasicRunInfoDto.kt
M-Kusumgar: basic one doesnt have logs

M-Kusumgar:
I have addressed the comments in the new pr #109

PR #104, opened 2024-08-20
Handle expired tokens from the ProtectedRoute component.

plietar:
Previously, ProtectedRoute would load the page even if the local storage contained an expired token. When the nested component tried to make a request to the API, the `getBearerToken` function would check the expiry time and would cause a redirection to the login page by assigning to `window.location.href`.

Assigning to `window.location.href` directly does not play well with the rest of the React router. In particular it does not take the PUBLIC_URL environment variable into account. This means that on an expired token, the user was always redirected to `/login` even when it should have been `$PUBLIC_URL/login`.

Using the `navigate(...)` function works much better. However it cannot be called directly from `getBearerToken()`, which is not a React component. Instead it can be done from the `ProtectedRoute`. In fact this is already implemented for the case where the user has no token at all. All that was required was a stricter definition of the `isAuthenticated()` function.

This change also makes sure the URL the user was trying to load with the expired token is saved through the existing `RedirectOnLoginProvider`. This was correctly implemented for GitHub authentication but not for the `BasicUserAuthForm`.

Steps to reproduce:
1. Start the frontend at a subdirectory, with `PUBLIC_URL=/app npm start`
2. Log in to the application with the super-user account.
3. Manually cause the token in the local storage to be expired, via the DevTools console:

    ```js
    localStorage.setItem("user", JSON.stringify({
      ...JSON.parse(localStorage.getItem("user")),
      exp: 0,
    }));
    ```

4. Navigate to a protected URL, such as `/app/manage-roles`.

Step 4 should direct the user to `/app/login`. After a successful login, the app should redirect to the `/app/manage-roles` page.

Without this change, the user is redirected to `/login`, which is not a valid route.

In the long term I think the `getBearerToken()` function is a bit of a code smell. It works completely outside of the rest of the application structure, and there is no reason to load from the local storage multiple times. Instead I think we should retrieve the token from the `UserProvider`. This does however entail a much larger refactor than this simple fix.

### DIFF
--- a/app/src/app/components/login/BasicUserAuthForm.tsx
+++ b/app/src/app/components/login/BasicUserAuthForm.tsx
@@ -1,4 +1,5 @@
 import { zodResolver } from "@hookform/resolvers/zod";
+import { useRedirectOnLogin } from "../providers/RedirectOnLoginProvider";
 import { Mail } from "lucide-react";
 import { useForm } from "react-hook-form";
 import { useNavigate, useSearchParams } from "react-router-dom";
@@ -17,6 +18,7 @@ export const BasicUserAuthForm = () => {
   const { setUser } = useUser();
   const [searchParams] = useSearchParams();
   const updatePasswordSuccess = searchParams.get("success");
+  const { requestedUrl, setRequestedUrl } = useRedirectOnLogin();
 
   const formSchema = z.object({
     email: z.string().email(),
@@ -40,7 +42,10 @@ export const BasicUserAuthForm = () => {
       });
 
       setUser(data.token);
-      navigate("/");
+
+      const url = requestedUrl || "/";
+      setRequestedUrl(null); // reset requested url before redirecting
+      navigate(url);
     } catch (error) {
       console.error(error);
       if (error instanceof ApiError) {
--- a/app/src/app/components/login/UserAuthForm.tsx
+++ b/app/src/app/components/login/UserAuthForm.tsx
@@ -3,6 +3,7 @@
 import { useEffect } from "react";
 import { useNavigate } from "react-router-dom";
 import { cn } from "../../../lib/cn";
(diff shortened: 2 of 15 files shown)


plietar:
There's still a test failure but I don't understand what is going on. Locally the test passes:

```
Summary of all failing tests
FAIL src/tests/components/contents/manageAccess/updatePermissions/PermissionScopeCommandList.test.tsx
  ● PermissionCommandList › should call setScopeResource, setFilterName and setOpen on select

    TypeError: Cannot read properties of null (reading 'visibilityState')

      at Object.isVisible (node_modules/swr/_internal/dist/index.js:140:59)
      at isActive (node_modules/swr/core/dist/index.js:55:38)
      at node_modules/swr/core/dist/index.js:330:29
```

The line in question from `swr` is:
```js
const isVisible = ()=>{
     const visibilityState = isDocumentDefined && document.visibilityState;
     return isUndefined(visibilityState) || visibilityState !== 'hidden';
};
```

where
```js
const STR_UNDEFINED = 'undefined';
const isDocumentDefined = typeof document != STR_UNDEFINED;
```

so somehow in the tests `document` is null instead of undefined?

What does this have to do with this PR?

absternator (approved):
Nice work code looks great!!!!

- SWR test comment.. yeah its an unmounting issue for sorts sometimes.... like component unmounts but swr is trying to make a call... i think
- getBearerToken was purposefully there to get from local storage so users can close/refresh and maintain their valid jwt, also allows the generic fetch to get the token... as userprovider reloads on refresh and we cant get token into fetch function unless we pass down with every SWR call

absternator (changes requested):
Actually after testing it didnt work for me... it only logged me out when I refreshed the page

absternator (approved):
dw should be sweet .. it was just the testing method you provided... as changing localstorage doesn't update the user state, thus it still stored old exp time thinking it was fine....

PR #105, opened 2024-08-20
Mrc-5677 Refactor orderly + outpack clients

absternator:
THis PR refactors the clients as per discussions from pr #98


M-Kusumgar (approved):
LGTM, pending lint fix

api/app/src/main/kotlin/packit/service/OrderlyRunnerClient.kt
```diff
@@ -25,55 +17,23 @@ interface OrderlyRunner
 class OrderlyRunnerClient(appConfig: AppConfig) : OrderlyRunner
 {
     val baseUrl: String = appConfig.orderlyRunnerUrl
-    private val restTemplate = RestTemplate()
-
     override fun getVersion(): OrderlyRunnerVersion
     {
-        return getEndpoint("/")
+        return GenericClient.get(constructUrl("/"))
```
EmmaLRussell, line 22: I guess one advantage of making `GenericClient` a base class or internal object (not static) of the specialised clients would be that it could keep the base url and you wouldn't have to constuct the url in the specialised client for every call. But not a showstopper.

EmmaLRussell (approved):
Works for me!

PR #106, opened 2024-08-22
Mrc-5659 Runner Layout

absternator:
The following PR creates the layout for the runner pages. This includes the logs and run page, and users only have access to this if they have packet.run permission... below is screenshot of how it looks 

Testing:
1. With admin user ensure can see link showing Runner page. When navigating ensure see layout as shown by screen shot. also try navigating b/w sidebar (logs and run) to see different page and URLs change.
2. Create user without packet.run permission and login. You should no be able to see Runner link... Also try head to URL /runner or /runner/logs and you should get unauthorized page 
image


M-Kusumgar (approved):
LGTM!

PR #107, opened 2024-08-22
Stream request bodies when proxying to outpack server.

plietar:
If a client tries to upload a file to Packit using the outpack API, the Packit server would try to buffer the entire request in memory before sending it out to the outpack server. This wasn't obvious from the code but an implementation detail of the HTTP libraries being used.

This caused issues when clients upload very large files - too large to fit in memory. The Packit server would run out of heap space and cause an exception.

Thankfully the issue is simple to fix, and the HTTP library only needs to be configured with `setBufferRequestBody(false)`. In fact, [according to the docs][setBufferRequestBody], the latest version of Spring has these flags set by default. We don't, however, use this version of Spring yet.

Steps to reproduce:
1. Create a 1GB `data` file:

    ```
    dd if=/dev/random of=data bs=1M count=1024
    ```

2. Start the Packit server with a 128MB maximum heap size:

    ```
    java -Xmx128m -jar app/build/libs/app.jar
    ```

3. Calculate the hash of the file and upload it to the server:

    ```
    hash=$(sha256sum data | awk '{ print $1 }')
    curl -X POST -T data \
      -H "Authorization: Bearer $TOKEN" \
      http://127.0.0.1:8080/outpack/file/sha256:$hash
    ```

Without this change, step 3 results in a `OutOfMemoryError` exception.

[setBufferRequestBody]: https://docs.spring.io/spring-framework/docs/6.1.0/javadoc-api/org/springframework/http/client/SimpleClientHttpRequestFactory.html#setBufferRequestBody(boolean)

### DIFF
--- a/api/app/src/main/kotlin/packit/controllers/OutpackServerController.kt
+++ b/api/app/src/main/kotlin/packit/controllers/OutpackServerController.kt
@@ -19,20 +19,23 @@ class OutpackServerController(private val outpackServerClient: OutpackServerClie
     @GetMapping("/**")
     fun get(request: HttpServletRequest, response: HttpServletResponse)
     {
-        proxyRequest(request, response)
+        proxyRequest(request, response, copyRequestBody = false)
     }
 
     @PostMapping("/**")
     @PreAuthorize("hasAuthority('outpack.write')")
     fun post(request: HttpServletRequest, response: HttpServletResponse)
     {
-        proxyRequest(request, response)
+        proxyRequest(request, response, copyRequestBody = true)
     }
 
-    private fun proxyRequest(request: HttpServletRequest, response: HttpServletResponse)
+    private fun proxyRequest(
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        copyRequestBody: Boolean)
     {
         val url = getUrlFragment(request)
-        outpackServerClient.proxyRequest(url, request, response)
+        outpackServerClient.proxyRequest(url, request, response, copyRequestBody)
     }
 
     private fun getUrlFragment(request: HttpServletRequest): String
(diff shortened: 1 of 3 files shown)


plietar:
I spent way too long trying to write integration tests for this but got no where.

I had code that created a 1GB `InputStream` full of zero bytes, that could be sent to the API. Naturally, the http client used by the integration was also buffering the request, and blew the heap. To fix that I tried to set `setBufferRequestBody(false)` on the test client as well, just like this PR does for the server's client, however the test client already has [`setOutputStreaming(false)
`](https://github.com/mrc-ide/packit/blob/e04be1e0cbf298d9513950c617c99ea1751898f4/api/app/src/test/kotlin/packit/integration/IntegrationTest.kt#L27).

With `setOutputStreaming(false)`, the ["the setFixedLengthStreamingMode and setChunkedStreamingMode methods of the underlying connection will never be called"](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/SimpleClientHttpRequestFactory.html#setOutputStreaming(boolean)), which means the request ends up being buffered one layer further down the stack (I don't know why it can't call those methods, but here we are).

I tried removing `setOutputStreaming(false)` from the test client, but then any test that is supposed to yield a 401 error produces an `HttpRetryException` instead. Turns out the underlying URLConnection does weird retry stuff but it can't if streaming is enabled.

[The docs for `TestRestClient`](https://docs.spring.io/spring-boot/api/java/org/springframework/boot/test/web/client/TestRestTemplate.html) suggest using "Apache Http Client 4.3.2" instead of URLConnection as the request factory, so I did that. It fixes the issues with the test client and now errors are correctly reported instead of raising `HttpRetryException`, but just having this in the classpath changes the behaviour of the server, and now the `IOUtils.copy(request.inputStream, serverRequest.body)` line from the `GenericClient` doesn't work because [there's no `getBody` implementation](https://github.com/spring-projects/spring-framework/blob/6.0.x/spring-web/src/main/java/org/springframework/http/client/HttpComponentsStreamingClientHttpRequest.java#L93-L95)

This is the point where I gave up.

plietar:
Re-opened as #108. I don't understand what GItHub did here and it won't let me re-open.

PR #109, opened 2024-08-22
Mrc 5627 runner table

M-Kusumgar:
sets up table for the runner


M-Kusumgar:
update, have made status nullable in #110

api/app/src/main/kotlin/packit/model/RunInfo.kt
```diff
@@ -0,0 +1,40 @@
...
+    @JdbcTypeCode(SqlTypes.JSON)
+    var parameters: Map<String, Any>? = null
+)
+
+fun RunInfo.toDto() = RunInfoDto(
+    taskId, packetGroupName, status, commitHash, branch, logs,
+    timeStarted, timeCompleted, timeQueued, packetId, parameters
+)
+
+fun RunInfo.toBasicDto() = BasicRunInfoDto(
```
absternator, line 37: feel like this is a bit overkill as it just doesn't return the logs... since logs is nullable anyways can probs just remove this BasicDto imo

absternator (approved):
LGTM nice work!!!  just had 1 more tiny comment.. feel free to ignore If ya disagree

api/app/src/main/kotlin/packit/model/RunInfo.kt
M-Kusumgar: i was thinking about json serialisation, basically we'll have a bunch of objects with logs as nulls, but now that i think about it, it doesnt really matter, i think youre rights, no need for basic dto as well

PR #110, opened 2024-08-22
Mrc 5603 submit endpoint

M-Kusumgar:
Can finally do the submit endpoint here. The main strategy is:
1. Hit orderly endpoint that submits the job
2. Get job id and save `RunInfo` entry in `run_info` database (this was created in #109)
3. return task id to the frontend so that it can redirect the user to the status logs page


api/app/src/main/kotlin/packit/model/dto/RunInfoDto.kt
```diff
@@ -3,7 +3,7 @@ package packit.model.dto
 data class RunInfoDto(
     val taskId: String,
     val packetGroupName: String,
-    val status: String,
+    val status: String? = null,
```
EmmaLRussell, line 6: The `RunInfoDto` is just the `BasicRunInfoDto` with logs? And we don't use inheritance because data classes don't get on with inheritance?

I guess status could arguably be an enum.

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
```diff
@@ -0,0 +1,8 @@
+package packit.model.dto
+
+data class SubmitRunInfo(
+    val name: String,
+    val branch: String,
+    val hash: String,
```
EmmaLRussell, line 6: Should branch and hash be nullable or do you always require both?

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
```diff
@@ -0,0 +1,8 @@
...
+    val parameters: Map<String, Any>?
```
EmmaLRussell, line 7: What are the possible types of parameters? Is it just string or number? Or even just strings (I think orderly parameters originally all got passed around as strings, but I might be misremembering!) But perhaps it's all more flexible now.

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
EmmaLRussell, line 6:
```suggestion
    val commitHash: String,
```
Maybe a bit clearer, and since this is what it's called in the db...

EmmaLRussell (commented):
Looks good! Just a couple of queries.

api/app/src/main/kotlin/packit/model/RunInfo.kt
```diff
@@ -13,7 +13,7 @@ class RunInfo(
     var taskId: String,
 
     var packetGroupName: String,
-    var status: String,
+    var status: String? = null,
```
absternator, line 16: when is this null? im sure when we can give it the initial status of "PENDING" when we submit... and then when we fetch status this can/may be updated

api/app/src/main/kotlin/packit/service/RunnerService.kt
```diff
@@ -45,4 +50,18 @@ class BaseRunnerService(
...
+
+    override fun submitRun(info: SubmitRunInfo): String
+    {
+        val res = orderlyRunnerClient.submitRun(info)
+        val runInfo = RunInfo(
+            res.taskId,
+            packetGroupName = info.name,
+            commitHash = info.hash,
+            branch = info.branch,
+            parameters = info.parameters
```
absternator, line 62: as per other comment add status = "PENDING" initiallu?

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
```diff
@@ -0,0 +1,8 @@
+package packit.model.dto
+
+data class SubmitRunInfo(
+    val name: String,
```
absternator, line 4:
```suggestion
    val packetGroupName: String,
```

api/app/src/test/kotlin/packit/integration/controllers/RunnerControllerTest.kt
```diff
@@ -2,20 +2,35 @@ package packit.integration.controllers
...
 import kotlin.test.assertEquals
 
 class RunnerControllerTest : IntegrationTest()
 {
+    @Autowired
+    private lateinit var runInfoRepository: RunInfoRepository
+
+    private fun getSubmitRunInfo(branch: String, commitHash: String): String
+    {
+        return "{\n" +
```
absternator, line 27:
```suggestion
        return jacksonObjectMapper().writeValueAsString(SubmitRunInfo("data", branch, commithash))
```

api/app/src/main/kotlin/packit/model/dto/RunInfoDto.kt
M-Kusumgar: agreed, ill make status an enum and i think both you and anmol (in prev PR) pointed out that the basicRunInfo and runInfo are a bit weird, ive just removed the basic one and the logs field can just be null in the `RunInfoDto`

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
M-Kusumgar: both are required in the current implementation of the runner

M-Kusumgar: so the orderly runner just takes in hash not commit hash, we could do json property annotation and use commitHash in the kotlin code, idm but then the frontend will have to send `hash` as well anyway, im happy to change it and add the annotation though, what do you think?

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
M-Kusumgar: it can be boolean, integer, number or string currently, orderly parameters were being passed around as string but i changed that because that doesnt make sense for when people want to input a number

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
M-Kusumgar: similarly to the comment above with emma suggesting `commitHash` instead of `hash`, the runner takes `name` so idm doing `packetGroupName` and doing a json property annotation of `name`?

api/app/src/test/kotlin/packit/integration/controllers/RunnerControllerTest.kt
M-Kusumgar: nice, that is much better :laughing:

M-Kusumgar: done

M-Kusumgar:
i have prematurely made the naming changes to `SubmitRunInfo` data class, lmk what you guys think

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
absternator: makes sense happy to leave as is

absternator (approved):
Nice work looks good to me :smile:

api/app/src/main/kotlin/packit/model/dto/RunInfoDto.kt
M-Kusumgar: oh btw enums are from https://github.com/mrc-ide/rrq/blob/ba2784057e8ad83e304ac9ae44081110c6b3d19c/R/common.R

api/app/src/main/kotlin/packit/model/dto/SubmitRunInfo.kt
EmmaLRussell: Ah, fair enough, not an issue then!

EmmaLRussell (approved):
Looks good!

PR #111, opened 2024-08-23
Add missing integrity and resolved fields to package-lock.json.

plietar:
When building Packit with Nix (and other sandboxed build systems), these fields are used to prefetch the dependencies. For some reason they are missing from our file for all but one dependency. There seems to be some npm bug that causes these fields to be removed under certain conditions (npm/cli#4263, npm/cli#6301).

Deleting the lockfile and running `npm install` again produces a new file that has the missing fields. However in doing so all the dependencies are updated, which I'd rather not right now. Instead I used the script at https://github.com/jeslie0/npm-lockfile to add the fields without changing the version numbers.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -65,7 +65,9 @@
     },
     "node_modules/@adobe/css-tools": {
       "version": "4.1.0",
-      "license": "MIT"
+      "license": "MIT",
+      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.1.0.tgz",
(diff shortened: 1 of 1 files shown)


M-Kusumgar (commented):
quick question because i am a little bit confused, this does seem like a one time fix (or am i misunderstanding?) this seems like if we install another dep or something it can get into this state again where it has some missing fields? do we meed to run this script more regularly? Set up some sort of automatic CI when it detects these missing fields?

plietar:
No, npm normally adds these automatically. If I add something to `package.json` and `npm install` the field for that package gets added.
The fact that they got removed the last time this file was touched is a niche npm bug (linked above), and hopefully won’t happen too often.

M-Kusumgar (approved):

PR #112, opened 2024-08-28
Overwrite any existing token after redirect.

plietar:
When receiving a redirection, after a succesful OAuth-based login, we should store the JWT in the local storage regardless of whether a token already exists.

Previously tokens would not be overwritten, which would cause issues with expired tokens never being replaced, preventing the user from logging in succesfully.

This used to be fine because getBearerToken would delete expired tokens, but after mrc-ide/packit#104 this was not the case anymore.


absternator (approved):
LGTM!!

PR #113, opened 2024-08-29
Mrc-5660 Run page branch info

absternator:
This ticket does the first part of the run page as seen in screenshot. It adds the branch select + info and the git fetch button. Subsequent PRs will finish the rest of the run page.

Testing: 
- go to run page and should see default branch in select with relevant info
- go inside debug docker container, navigate to /orderly-root-remote and create a new git branch... Then head to web app and press fetch button. Then click on branch select and new branch should now be there

image


M-Kusumgar (commented):
Looks good to me! really nice to finally see it working, just got a couple of comments that are optional

app/src/app/components/contents/runner/PacketRun.tsx
```diff
@@ -1,3 +1,43 @@
...
+
 export const PacketRun = () => {
-  return <div>Run</div>;
+  const { branchData, error, isLoading, mutate } = useGetGitBranches();
+
+  return (
+    <div>
+      <div>
+        <h2 className="text-2xl font-bold tracking-tight">Run</h2>
+        <p className="text-muted-foreground">Run a packet group to create packets</p>
```
EmmaLRussell, line 13:
```suggestion
        <p className="text-muted-foreground">Run a packet group to create a new packet</p>
```

Maybe slightly more accurate - you'll only create one new packet when you run a packet group (once)!

app/src/app/components/contents/runner/PacketRun.tsx
```diff
@@ -1,3 +1,43 @@
...
+            <Skeleton className="h-12 w-96" />
+          </li>
+          <li className="mb-8 flex">
+            <Skeleton className="h-12 w-96" />
+          </li>
+          <li className="mb-8 flex">
+            <Skeleton className="h-12 w-24" />
+          </li>
+        </ul>
+      )}
```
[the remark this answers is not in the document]
EmmaLRussell: This is maybe a slightly different configuration of Skeleton to match a different type of UI? But that could potentially be abstracted out to a single LoadingSkeleton component, with maybe an enum prop to indicate UI type (table, form etc)?

app/src/app/components/contents/runner/run/PacketRunBranchField.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <>
+      <FormLabel className="font-semibold text-lg">Branch</FormLabel>
+      <div className="flex space-x-3">
+        <div className="flex-grow">
+          <FormField
+            control={form.control}
+            name="branch"
+            render={({ field }) => (
+              <FormItem>
+                <Select defaultValue={field.value} onValueChange={field.onChange}>
```
EmmaLRussell, line 47: Sorry, just to clarify as I'm still not super-familiar with this style of React forms - if I understand it, the field comes out of the  form which is passed as a prop into the component from the parent form, and it's selecting the field it wants to use from the "name" prop in `FormField` - and we're setting an `onChange` event handler here, which will trigger `onChange` on that field in the form, if it has one defined (which it doesn't currently - or is there a default handler implied by `useForm` which just updates the value?).

So it's a way to let components like this handle rendering of fields, but not needing to know anything about how to handle changes to values, as those event handlers are defined in whatever creates the 'form' object (the parent form component in this case)..?

app/src/app/components/contents/runner/run/PacketRunBranchField.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const lastCommitTime = getTimeDifferenceToDisplay(selectedBranch.time);
+
+  const gitFetch = async () => {
+    try {
+      await fetcher({
+        url: `${appConfig.apiUrl()}/runner/git/fetch`,
+        method: "POST"
+      });
+    } catch (error) {
+      console.error(error);
```
EmmaLRussell, line 32: Shouldn't we display an error?

EmmaLRussell (commented):
Looks great! Couple of tiny comments and one slightly confused question..!

app/src/app/components/contents/runner/PacketRun.tsx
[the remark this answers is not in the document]
absternator: yeah, the issue is its different depending on where it is and how the skeleton looks...

app/src/app/components/contents/runner/run/PacketRunForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+  packetGroupName: z.string(),
+  parameters: z.record(z.any())
+});
+export const PacketRunForm = ({ defaultBranch, branches, mutate }: PacketRunFormProps) => {
+  const form = useForm<z.infer<typeof packetRunFormSchema>>({
+    resolver: zodResolver(packetRunFormSchema),
+    defaultValues: {
+      branch: defaultBranch,
+      packetGroupName: "",
+      parameters: {}
```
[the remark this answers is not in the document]
absternator: have removed this in next PR :smile:

app/src/app/components/contents/runner/run/PacketRunBranchField.tsx
absternator: yip pretty much :smile

app/src/app/components/contents/runner/run/PacketRunBranchField.tsx
absternator: good idead done

EmmaLRussell (approved):

M-Kusumgar (approved):

PR #116, opened 2024-09-01
Mrc-5662 Run submit

absternator:
This PR continues on from #115 and contains the run button and submission of run. After the run button is clicked and is successful, user will get redirected to logs page for that run (taskId).

Testing:
1. on runner page fill out fields and click run. Shold get redirected to page with task logs for that task id


M-Kusumgar (approved):
LGTM!

api/app/src/main/kotlin/packit/controllers/RunnerController.kt
```diff
@@ -4,64 +4,53 @@ import org.springframework.http.ResponseEntity
...
-import packit.model.dto.SubmitRunInfo
+import packit.model.dto.*
 import packit.service.RunnerService
 
 @RestController
 @RequestMapping("/runner")
 @PreAuthorize("hasAuthority('packet.run')")
-class RunnerController(private val runnerService: RunnerService)
-{
+class RunnerController(private val runnerService: RunnerService) {
```
EmmaLRussell, line 13: What's with the formatting changes in this file - they're not consistent with the other modules in the project I don't think.. ?

EmmaLRussell (commented):
Guess I missed the boat on this one too! 😢 

It seems to make a lot of API calls when you hit run even when the form isn't valid. Run button probably shouldn't be enabled in that case?
image (13)

absternator:
> Guess I missed the boat on this one too! 😢
> 
> It seems to make a lot of API calls when you hit run even when the form isn't valid. Run button probably shouldn't be enabled in that case?

could not replicate this

EmmaLRussell:
> > Guess I missed the boat on this one too! 😢
> > It seems to make a lot of API calls when you hit run even when the form isn't valid. Run button probably shouldn't be enabled in that case?
> 
> could not replicate this

Weird! I'll try to show it to you next time we're both in the office.

PR #118, opened 2024-09-05
Fix incorrect uses of `argThat`.

plietar:
In a couple of places, `argThat` was being called with a sequence of boolean expressions. That pattern of use is incorrect, and while all the expressions are evaluated only the last line of the block is returned and used by `argThat`. Any of the preceding expressions could evaluate to false without failing the test.

The most direct fix to this would be to take the conjunction of all the expressions by interspersing some `&&` between each. While that would work, it would also lead to terribly obscure error message on a failure, as it would be impossible to figure out which expression evaluated to false.

The chosen fix is to instead use a series of assertions, and then return true at the end of the block. Given that any failed assertion raises an exception, this doesn't have the same issue as the old code.

It would have been nice to avoid the confusing `true` at the end. In fact, Mockito has added a [new `assertArg` method][assertArg] in recent versions. I tried to update mockito but ran into too many issues to deal with.

[assertArg]: https://github.com/mockito/mockito/pull/2949

### DIFF
--- a/api/app/src/test/kotlin/packit/unit/service/RoleServiceTest.kt
+++ b/api/app/src/test/kotlin/packit/unit/service/RoleServiceTest.kt
@@ -99,8 +99,9 @@ class RoleServiceTest
 
         verify(roleRepository).save(
             argThat {
-                this.name == createRole.name
-                this.rolePermissions.size == 2
+                assertEquals(this.name, createRole.name)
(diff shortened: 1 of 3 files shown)


absternator (approved):
thanks for fixing nd teaching me how to use argthat :smile:

PR #119, opened 2024-09-06
Catch exceptions when checking for org and team membership.

plietar:
When we check for org and team membership, we may get a 403 error if the GitHub token does not have sufficient permissions. This can happen quite easily if a user creates a PAT without the right scopes. Without this, Packit was returning empty error responses, with no body attached to it.


api/app/config/detekt.yml
```diff
@@ -49,6 +49,8 @@ style:
     active: false
   UnusedPrivateMember:
     active: false
+  ThrowsCount:
+    active: false
```
plietar, line 53: Detekt was complaining about too many throws statement, and honestly I find this style rule questionable. Let me know if you would prefer me to abide by it and refactor the code to comply instead.

EmmaLRussell: Agreed, I think we probably just haven't triggered this one before..

EmmaLRussell (approved):
Thanks! LGTM.

PR #120, opened 2024-09-06
Properly render boolean valued parameters.

plietar:
For some reason, React does not render booleans when using the obvious `{val}` syntax. I can't find any authoritative source on this in the React docs, but it seems to be [a fairly well attested fact][stackoverflow]. This was hiding the value of boolean parameters when rendering them in the front-end.

The workaround is simply to call `String` with the value of the parameters, which will reliably turn all parameter types into something React can render.

I've added a packet to the included repository that uses boolean parameters. I've also removed the orderly index - this is a cache and I do not think it makes sense to include it.

[stackoverflow]: https://stackoverflow.com/questions/38337262/cannot-render-boolean-value-in-jsx


absternator (commented):
Yeah makes sense why you can't see bools but I noticed this was happening in my PR that is up right now and fixed it too 🤣🤣🤣 maybe just leave this PR for now and close it 🙂

plietar:
Sure.

PR #122, opened 2024-09-10
Mrc-5708 Get stasuses of tasks

absternator:
This PR adds the endpoint that gets the status of all tasks. A new BasicRunInfoDto is created that has more limited info that will be displayed on the all logs page. 

Testing:
1. go to runner page and submit multiple jobs
2. hit endpoint http://localhost:8080/runner/list/status . this should return expected statuses of tasks 

Note: DO not merge until orderly.runner PR #11 is merged first.. the branch in common will need to be updated then to main before merge


M-Kusumgar (approved):
LGTM!

PR #123, opened 2024-09-12
Add an Actions workflow to check that migrations aren't being changed.

plietar:
Changing existing migrations causes all sorts of problems when deploying
to an existing server. Flyway will complain that the migration's
checksum has changed, and will require manual intervention to fix it,
either by dropping the whole database, or by manually modifying the
schema to match the new version of the migration. This defeats the
purpose of migrations, which should prevent manual management of the
schema.

This has already happened multiple times, so prevent any further
incidents by adding a workflow on PRs to prevent it. The workflow is
intentionally separate from the others to make it easier to ignore in
case of false positives.

### DIFF
--- a/.github/workflows/check-migrations.yml
+++ b/.github/workflows/check-migrations.yml
@@ -1,0 +1,18 @@
+# This checks that pull requests to main don't modify existing migrations.
+# Doing so causes problems for deployment to existing instances.
+name: check-migrations
+on:
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  check-migrations:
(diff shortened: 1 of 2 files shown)


plietar:
See https://github.com/mrc-ide/packit/actions/runs/10834351430/job/30063312461 for an example of a failed run.

absternator (approved):
So yeah the whole point of flyaway and the migrations are they are immutable. They should not be changed at all!!!

Nonetheless like the change here to make sure people don't change it. 

Can you maybe a sentence in the readme too please stating this 🙂

PR #124, opened 2024-09-13
Mrc-5760 Add Username to Runinfo table

absternator:
Add Username to Runinfo table. this info needs to be displayed in the FE. 

Testing: 
- Submit a test run 
- hit the http://localhost:8080/runner/list/status and http://localhost:8080/runner/status/{task_id}. should see username show up in response

### DIFF
--- a/api/app/src/main/resources/db/migration/V20240912__update_run_info.sql
+++ b/api/app/src/main/resources/db/migration/V20240912__update_run_info.sql
@@ -1,2 +1,2 @@
 ALTER TABLE "run_info"
-    ADD COLUMN "username" TEXT NOT NULL;
+    ADD COLUMN "ran_by" TEXT NOT NULL;
--- a/api/app/src/main/resources/db/migration/V20240912__update_run_info.sql
+++ b/api/app/src/main/resources/db/migration/V20240912__update_run_info.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "run_info"
+    ADD COLUMN "username" TEXT NOT NULL;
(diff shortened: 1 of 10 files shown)


api/app/src/main/resources/db/migration/V20240912__update_run_info.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "run_info"
+    ADD COLUMN "username" TEXT NOT NULL;
```
plietar, line 2: Why a username and not a user id?

absternator: good point i had that initially and tie the User entity. But because we only need the username and no other info it sounded easier and more clean to store as username is unique also... Also the way we store this is when we submit a run, we get this from the token; which username is available.... if we did userID, we would have to hit DB get user first then store the userID meaning an extra db call

M-Kusumgar (commented):
Nice, works as intended, however i think we should use the `displayName` is available and fall back on `username` if that doesnt exist no? if so maybe we rename field to `ranBy` or something like that

absternator:
> Nice, works as intended, however i think we should use the `displayName` is available and fall back on `username` if that doesnt exist no? if so maybe we rename field to `ranBy` or something like that

Good idea!! done

api/app/src/main/resources/db/migration/V20240912__update_run_info.sql
plietar: It seems like there's even more of a reason to use the ID now that you sometimes show the displayName, sometimes the username. Especially given that the display name isn't unique, and may even change. In the future we will probably want to filter jobs by user, in which case a "ranBy" column won't help you.

If the number of DB roundtrips really are a concern (I doubt they are), a join between the two tables will get you everything in a single call. I don't know enough about JPA to know how easy this is to do, but it seems like a basic enough task that it should be supported.

The user ID can be added to the token if that helps.

absternator: yup you are right done :smile:

M-Kusumgar (approved):
LGTM!

PR #125, opened 2024-09-13
Mrc-5761 Paginate + filter support for statuses endpoint

absternator:
This PR adds pagination + filter support for packetgroup name. as per mockups for FE we need this filter, also logs list will grow thus need to paginated. Also sort is applied as want most recent thus is sorted by timeQueued

Testing:
1. submit several runs with different packet groups
2. hit /runner/list/status endpoint with different query params and ensure right behaviour.. `/runner/list/status?pageNumber=$pageNumber&pageSize=$pageSize&filterPacketGroupName=$filterPacketGroupName`


M-Kusumgar (approved):
just tested this out! looks good to me!

PR #126, opened 2024-09-17
Expose metrics over HTTP.

plietar:
This enables Spring Boot Actuator, which exposes lots of monitoring information about the service. In particular, this exposes metrics in a Prometheus-compatible format. There are a lot of built-in metrics that come with Spring already. In addition, this add a metric with the Git revision of the repo that was used at build time, similar to what is done in outpack_server.

To avoid exposing everything over the default HTTP port, the management interface is bound to a separate port, 8081 by default.


EmmaLRussell (commented):
Looks fine to me as far as I can tell! Would it be possible to include an e2e or integration test?

api/app/src/test/kotlin/packit/integration/MetricsTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.boot.test.autoconfigure.actuate.observability.AutoConfigureObservability
+import org.springframework.boot.test.web.client.TestRestTemplate
+import org.springframework.boot.test.web.server.LocalManagementPort
+import org.springframework.boot.web.client.RestTemplateBuilder
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+@AutoConfigureObservability
```
EmmaLRussell, line 13: Just curious - why did you need to add this?

EmmaLRussell (approved):
LGTM!

api/app/src/test/kotlin/packit/integration/MetricsTest.kt
plietar: Spring disables metrics in tests by default

https://stackoverflow.com/a/65560631
https://github.com/spring-projects/spring-boot/pull/21658

PR #127, opened 2024-09-18
Mrc-5767 Task logs page

absternator:
The PR is the FE for the logs/summary of a run task. The design is from [figma](https://www.figma.com/design/9mO1pIrQ3k4IlOqbTehybr/Packit?node-id=0-1&node-type=canvas&t=1AQkxddRQtUEbXOv-0).The only thing updates is showing logs with black background and white text. Also have added seperator b/w parameters and branch info, as with lots of parameters the branch info got lost. There is also a refactor to `getTimeDifferenceToDisplay` and it has been moved to `time.ts` where we can all add all time related generic functions

I have made a  separate ticket for polling the status.

Testing:

To See completed and running layouts:
- run a packetgroup from the runner in the Web UI... you will see running layouts. then refresh and you will see completed layouts.

To see error layout:
- hit API submit with bad body, then copy the taskId you get back and in the web UI go to /runner/logs/{taskId}
Ex bad request: hit `http://localhost:8080/runner/run` and body
```
{   
 "name": "t2",
    "branch": "master",
    "hash": "06e9c200fb9c9bdc494ac2e4913216d8c86a43ce"
}
``` 
image
image

### DIFF
--- a/app/src/app/components/contents/explorer/PacketGroupSummaryListItem.tsx
+++ b/app/src/app/components/contents/explorer/PacketGroupSummaryListItem.tsx
@@ -1,7 +1,7 @@
 import { ExternalLink, History, Hourglass } from "lucide-react";
 import { Link } from "react-router-dom";
 import { PacketGroupSummary } from "../../../../types";
-import { getTimeDifferenceToDisplay } from "./utils/getTimeDifferenceToDisplay";
+import { getTimeDifferenceToDisplay } from "../../../../lib/time";
 
 interface PacketGroupSummaryListItemProps {
   packet: PacketGroupSummary;
--- a/app/src/app/components/contents/explorer/PacketGroupSummaryListItem.tsx
+++ b/app/src/app/components/contents/explorer/PacketGroupSummaryListItem.tsx
@@ -8,7 +8,7 @@ interface PacketGroupSummaryListItemProps {
 }
 
 export const PacketGroupSummaryListItem = ({ packet }: PacketGroupSummaryListItemProps) => {
-  const { unit, value } = getTimeDifferenceToDisplay(packet.latestTime);
+  const { unit, value } = getTimeDifferenceToDisplay(packet.latestTime)[0];
 
   return (
     <li key={packet.latestId} className="p-4 flex flex-col border-b space-y-1">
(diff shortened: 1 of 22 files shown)


M-Kusumgar (commented):
looks great! almost there, couple of small comments:
* i think it would be nice to have the parameters below git info in summary because parameters might potentially be unbounded (not sure how we're dealing with that in the UI actually)
* while the task is running i think it would be nice to have some sort of animation like in the mockup, just to let users know that something is loading and the page isnt stuck
* the worker git config should probably be baked into the worker start up so i can work on that or if you fancy doing some R you can also do that, up to you! idm, basically just have to set the worker's git config `user.name` and `user.email` at start up: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5783/orderly.runner-git-config-in-start-up (not related to this PR)
* i found a problem (not related to us) in the logging, if you run the `explicit` report in the runner then the log is little bit messed up, i might need to investigate this one


edit, couple of UI suggestions too:
* in the summary, the two columns of the bottom summary are vertically misaligned and different font sizes, i think the horizontal line between them also makes me notice it more, would be good to make ditch that horizontal line, make them the same font size so everything aligns nicely?:
image

sorry another edit, got another suggestion:
* when i click the `view` button it takes me to the summary page for the packet details, but theres no logs button in the nav for me to go back, maybe we can have a discussion about how to combine these two nav bars because they are basically the same thing

absternator:
@M-Kusumgar have updates as per suggestions and tickets made for extra work. 
image

M-Kusumgar (approved):
think you have an unused import, cos the fe test is failing, the loading symbol is all good but i think its a little bigger than the check symbol so there seems to be no margin between it and the time right next to it, maybe another ticket for the symbols or feel free to fix it in this one! but happy to approve this

absternator:
> think you have an unused import, cos the fe test is failing, the loading symbol is all good but i think its a little bigger than the check symbol so there seems to be no margin between it and the time right next to it, maybe another ticket for the symbols or feel free to fix it in this one! but happy to approve this

oo yes good catch have made smaller and will fix tests and import before merge thanks!!

PR #128, opened 2024-09-18
Mrc-5767 Poll status for task logs page

absternator:
Poll status when status is "pending" or "running". 

Testing:
- submit a task, and in running mode will auto update to complete. Check network requests to ensure these have stopped


M-Kusumgar (approved):
looks good to me!

PR #129, opened 2024-09-20
Add a `/user` endpoint to the API.

plietar:
Previously the frontend got the list of users by listing the roles and filtering on the isUsername flag. This was a little awkward, and meant that extra information about the user isn't accessible to the frontend.

As part of mrc-5763, I want to add a "service account", which will need to be displayed differently, or maybe won't even be displayed at all. Because the endpoint the front end is calling doesn't get the full user details, it cannot tell whether a user is a service account or not.

For this to work, a new `GET /user` endpoint is added, which lists the users and all their properties. The frontend is modified to use this new endpoint. Both the `/role` and the `/user` endpoints are now setup to never return username roles. From the perspective of the frontend, these special roles don't exist anymore. The `/user` endpoint includes the permissions that are specific to the user.

There's a lot of churn in the frontend from renaming some files and classes into something I feel makes more sense.

I've removed a few bits of code that were dead, especially in the User and Role repositories.


api/app/src/main/kotlin/packit/model/Role.kt
```diff
@@ -35,11 +35,22 @@ class Role(
...
+
+    fun isHidden(): Boolean = isUsername
 }
 
 fun Role.toDto() =
     RoleDto(
-        name, rolePermissions.map { it.toDto() }, users.map { it.toBasicDto() }, id!!, isUsername
+        name,
+        rolePermissions.map { it.toDto() }.sortedByDescending {
+            it.tag == null && it.packet == null && it.packetGroup == null
```
plietar, line 46: I copied this from `getSortedRoleDtos`. I'm not sure I really understand what it is trying to achieve tbh. Does it put global permissions at the top, and scoped ones further down?

api/app/src/main/kotlin/packit/model/Role.kt
```diff
@@ -35,11 +35,22 @@ class Role(
...
+        },
+        users.toBasicDto(),
+        id!!,
+        isUsername
     )
 
 fun Role.toBasicDto() = BasicRoleDto(name, id!!)
+
+fun List<Role>.toDto() = filter { !it.isHidden() }.sortedBy { it.name }.map { it.toDto() }
+fun List<Role>.toBasicDto() = filter { !it.isHidden() }.sortedBy { it.name }.map { it.toBasicDto() }
```
plietar, line 56: Let me know if you think the `toDto` methods are too clever now. The issue with the old approach of having a `getSortedRoleDtos`-style method was that it was tricky to apply filtering and sorting consistently, especially in the nested structures (eg. making sure the roles inside of a user are sorted, and vice versa). Putting that logic on `List<...>.toDto` helps.

api/app/src/main/kotlin/packit/model/User.kt
```diff
@@ -17,20 +18,42 @@ class User(
...
+
+    @OneToOne()
+    @JoinColumn(
+        name = "username",
+        referencedColumnName = "name",
+        insertable = false,
+        updatable = false
+    )
+    @Filter(name = "isUsername", condition = "isUsername = TRUE")
+    var userRole: Role? = null,
```
plietar, line 41: I'm honestly not all that sure of what I'm doing here, but it seems to be working. I wanted the `userRole` to be available from within the `toDto` methods and a join seemed like the cleanest way.

api/app/src/test/kotlin/packit/integration/controllers/RoleControllerTest.kt
```diff
@@ -1,9 +1,8 @@
 package packit.integration.controllers
```
plietar, line 1: I got a bit carried away tidying this file. In particular I removed all the manual calls to jacksonObjectMapper and instead use the dtos to deserialize the objects more conveniently.

app/src/app/components/contents/manageAccess/types/User.ts
```diff
@@ -0,0 +1,13 @@
+import { RolePermission } from "./Role";
+
+interface BasicRole {
+  name: string;
+  id: string;
+}
+
+export interface User {
```
plietar, line 8: This is just the existing `UserWithRoles`, renamed and tidied up a bit.

PR #130, opened 2024-09-20
Add a service account.

plietar:
As part of adding support for authentication from OIDC tokens for GitHub actions, we need a user that we can assign to the tokens we issue. This adds a new user source, called "service", and populates the database with one of these accounts.

The service user is hidden from the API, and therefore is not visible in the control panel. Moreover, it cannot be given additional roles and permissions.

The service user has the ADMIN role, but in practice when issuing the access tokens we will restrict the actual scopes given to each token. As a precaution, I added some checks to make sure the different user sources cannot use users created by another.


api/app/src/main/resources/db/migration/V20240919__create_service_account.sql
```diff
@@ -0,0 +1,10 @@
+WITH inserted_user AS (
```
plietar, line 1: I intentionally did not add a `service` role for the user. I don't think it is useful, and it opens the door to modifying that role through the API, which I'd rather avoid.

If we ever have a use case for it we can add it later as another migration.

api/app/src/main/kotlin/packit/service/RoleService.kt
```diff
@@ -62,12 +51,22 @@ class BaseRoleService(
         return saveRole(createRole.name, permissions)
     }
 
+    override fun createUsernameRole(username: String): Role
```
plietar, line 54: This renames and changes the semantics of `getUsernameRole` a bit.

I found it confusing that a method calls `get` has side-effects. I also don't think there's ever a valid use-case for returning an existing role, and it was opening up the door to someone creating a GitHub account whose username collides with an existing role, thereby gaining all of that roles permissions.

plietar:
It's not really ready yet, but if you are curious you can look at https://github.com/mrc-ide/packit/pull/131 for how this gets used.

api/app/src/main/kotlin/packit/service/UserService.kt
```diff
@@ -149,7 +153,17 @@ class BaseUserService(
...
+            throw PackitException("cannotUpdateServiceUser", HttpStatus.BAD_REQUEST)
+        }
+
         userRepository.delete(user)
         roleService.deleteUsernameRole(username)
     }
+
+    override fun getServiceUser(): User
+    {
+        return userRepository.findByUsernameAndUserSource("SERVICE", "service")
```
absternator, line 166: maybe we can get name from application.properties? and be injected in on a per app level

PR #131, opened 2024-09-20
 Add secret-less authentication for CI jobs.

plietar:
This change allows automated environments, such as Github Actions or Buildkite to authenticate to the Packit API without needing to be manually provisioned with a secret.

The client's environment creates and signs a JWT containing claims describing the job. For example, on Github Actions, GitHub provides a token which includes the organisation name, repository name, type of event, branch name, etc. Packit needs to be configured to trust the 3rd party JWT issuer, with policies specifying what claims are needed.

The client sends the JWT it received from its environment to the Packit `/auth/login/jwt` endpoint, and in exchange receives a Packit JWT which may be used to access the rest of the API.


plietar:
The naming for this in other projects is kind of a mess:

- GitHub calls this feature "OpenID Connect". I disagree with their naming, as there's nothing related about OIDC in the protocol. Both this scheme and OIDC use signed JWTs to prove identity, but OIDC is a much broader standard, and this scheme actually differs in some subtle points.
- Buildkite also calls this OIDC, I suspect they'll have copied GitHub's naming
- PyPI calls this "trusted publishers": https://docs.pypi.org/trusted-publishers/
- Vault calls it `jwt` (or `oidc` when the actual OIDC bits are used): https://developer.hashicorp.com/vault/docs/auth/jwt
- Kubernetes calls this "Bound Service Account Tokens": https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md

I decided to avoid the OIDC name and just stick to JWT, or "External JWT" when it is ambiguous wrt Packit's own tokens. I'm happy to use a more catchy name if we want

plietar:
This is an example of it being used in practice: https://github.com/plietar/orderly-ci-test/blob/f1987986d5fceeaf7b4cceb9aacf2708ae776f0a/run.R

api/app/src/main/kotlin/packit/config/JwtLoginConfig.kt
```diff
@@ -0,0 +1,22 @@
...
+  val grantedPermissions: List<String> = listOf(),
+
+  @DurationUnit(ChronoUnit.SECONDS)
+  val tokenDuration: Duration? = null,
+)
+
+@ConfigurationProperties(prefix = "auth.external-jwt")
+data class JwtLoginConfig(
+  val audience: String?,
+  val policy: List<JwtPolicy> = listOf()
```
absternator, line 21: policies instead of policy?

api/app/src/main/kotlin/packit/service/JwtLoginService.kt
```diff
@@ -0,0 +1,103 @@
...
+  val jwtIssuer: JwtIssuer,
+  val userService: UserService,
+  val jwtLoginConfig: JwtLoginConfig,
+  val restOperations: RestOperations = RestTemplate(),
+) {
+  lateinit var policies: List<TokenPolicy>
+
+  val audience: String? = jwtLoginConfig.audience
+
+  init {
```
absternator, line 39: all the init should probs be in `JwtLoginConfig` and then it has the policies it can pass.. and thus we can keep the service clean

api/app/src/main/kotlin/packit/security/provider/TokenProvider.kt
```diff
@@ -23,23 +39,59 @@ class TokenProvider(val config: AppConfig) : JwtIssuer
         const val TOKEN_AUDIENCE = "packit"
     }
 
-    override fun issue(userPrincipal: UserPrincipal): String
-    {
+    inner class Builder(val userPrincipal: UserPrincipal) : JwtBuilder {
```
absternator, line 42: the inner class stuff makes a bit messy... could move inner class outside and just pass along the config when calling instantiating class

api/app/src/main/kotlin/packit/security/provider/TokenProvider.kt
```diff
@@ -23,23 +39,59 @@ class TokenProvider(val config: AppConfig) : JwtIssuer
...
+        var duration: Duration = Duration.of(config.authExpiryDays, ChronoUnit.DAYS)
```
absternator, line 43: should probs be private

api/app/src/main/kotlin/packit/service/JwtLoginService.kt
```diff
@@ -0,0 +1,103 @@
...
+        decoder.setJwtValidator(DelegatingOAuth2TokenValidator(validators))
+        TokenPolicy(decoder, policy)
+      }
+    } else {
+      policies = listOf()
+    }
+  }
+
+  private fun issueToken(verifiedToken: Jwt, policy: JwtPolicy): String {
+      val user = userService.getServiceUser()
```
absternator, line 68: also this all seems coupled to a service account... thus would a better name for all files and called be `ServiceLogin...`

absternator (commented):
Nice work!!!! looks great and i learn lots just reading it!! just left some comments about code structure

api/app/src/main/kotlin/packit/controllers/LoginController.kt
```diff
@@ -50,6 +52,31 @@ class LoginController(
         return ResponseEntity.ok(token)
     }
 
+    @PostMapping("/login/jwt")
```
absternator, line 55: as per other comment about naming.. maybe should be `/login/service`

api/app/src/main/kotlin/packit/config/JwtLoginConfig.kt
```diff
@@ -0,0 +1,22 @@
...
+  val jwkSetURI: String,
+  val issuer: String,
+  val requiredClaims: Map<String, String> = mapOf(),
+  val grantedPermissions: List<String> = listOf(),
+
+  @DurationUnit(ChronoUnit.SECONDS)
+  val tokenDuration: Duration? = null,
+)
+
+@ConfigurationProperties(prefix = "auth.external-jwt")
```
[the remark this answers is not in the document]
plietar: > do we need to add these into the applcicaton.properties

Because these are much more structured, I don't think it makes sense to use flat string environment variables to pass them in, but instead need to pass them as structured data.

There's a number of ways we can do it, see: https://docs.spring.io/spring-boot/reference/features/external-config.html

api/app/src/main/kotlin/packit/controllers/LoginController.kt
```diff
@@ -50,6 +52,31 @@ class LoginController(
...
+    {
+        if (jwtLoginService.audience == null) {
+            throw PackitException("jwtLoginDisabled", HttpStatus.FORBIDDEN)
+        }
+
+        val token = jwtLoginService.authenticateAndIssueToken(user)
+        return ResponseEntity.ok(token)
+    }
+
+    @GetMapping("/login/jwt/audience")
```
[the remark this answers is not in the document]
plietar: That we can't change it (or at least not easily).

Most of the Packit API endpoints are only used by the Packit webapp, so they can easily be changed in a single PR. A few endpoints, mostly `/auth/login/api` and `/outpack` are used by orderly, and we couldn't change these easily. We'd have to change orderly, but also take into account the fact that people might have it installed already and worry about how that affect compatibility.

We need whatever returns the audience to be stable. It could be `/config`, or it could be a dedicated endpoint like I did here. Having a dedicated endpoint means we have to worry less about changing `/config`.

api/app/src/main/kotlin/packit/service/JwtLoginService.kt
```diff
@@ -0,0 +1,103 @@
...
+import packit.model.dto.LoginWithToken
+import packit.security.provider.JwtIssuer
+import java.util.function.Predicate
+
+data class TokenPolicy(
+  val decoder: NimbusJwtDecoder,
+  val config: JwtPolicy,
+)
+
+@Component
```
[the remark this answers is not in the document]
plietar: Thanks

api/app/src/main/kotlin/packit/config/JwtLoginConfig.kt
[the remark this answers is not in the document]
plietar: > as per docs? or does it hyphen the camelcase?

So Spring is kind of weird and I got errors here if I didn't use kebab case here. It seems to apply some normalization to the input properties, which means that it doesn't actually matter what case people use, and both `external-jwt` and `externalJwt` would work.

See https://github.com/spring-projects/spring-boot/wiki/Canonical-properties
And https://github.com/spring-projects/spring-boot/blob/eb7b6a776dd8b01849a0ea35d097b81f7eeec23e/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/diagnostics/analyzer/InvalidConfigurationPropertyNameFailureAnalyzer.java#L51-L52

Although you are right the docs I've written are inconsistent and use both. I'll switch to just one.

api/app/src/main/kotlin/packit/service/JwtLoginService.kt
plietar: I'm not sure I agree. The `JwtLoginConfig` is a plain data class that just represents the properties. It seems odd that it would be initializing some business logic by creating validators and JWT decoders.

api/app/src/main/kotlin/packit/service/JwtLoginService.kt
plietar: Yeah makes sense

api/app/src/main/kotlin/packit/security/provider/TokenProvider.kt
plietar: Done

api/app/src/main/kotlin/packit/security/provider/TokenProvider.kt
plietar: Done

api/app/src/main/kotlin/packit/config/JwtLoginConfig.kt
plietar: Done

api/app/src/main/kotlin/packit/config/JwtLoginConfig.kt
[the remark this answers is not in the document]
absternator: > I don't think it makes sense to use flat string environment variables to pass them in, but instead need to pass them as structured data.

so yeah I thought this structured data is picked up from the application.properties? because how does it know what to set the values to ? i thought you need what you have in your docs as well?

api/app/src/main/kotlin/packit/service/JwtLoginService.kt
absternator: yeah that is fair but it seems like it does derive directly from config and doesn't need anything else right? and yeah the init in the service seems a bit odd.. maybe could create a bean with this logic and then directly inject into jwtlogin service?

something like this maybe? 
```
@Configuration
class JwtConfig {
    @Bean
    fun tokenPolicies(jwtLoginConfig: JwtLoginConfig, restOperations: RestOperations): List<TokenPolicy> {
        // init logic
    }
}
```

then can pass into service straight up
```
class JwtLoginService(
    val policies: List<TokenPolicy>,
 val jwtIssuer: JwtIssuer,
    val userService: UserService,
)
```


but if ya disagree feel free to keep your way :smile:

api/app/src/main/kotlin/packit/controllers/LoginController.kt
```diff
@@ -50,6 +52,31 @@ class LoginController(
         return ResponseEntity.ok(token)
     }
 
+    @PostMapping("/login/service")
+    @ResponseBody
+    fun loginJWT(
```
absternator, line 57:
```suggestion
    fun loginService(
```

absternator (approved):
Nice work looks good to me!!! just a couple small comments but feel free to merge after response

api/app/src/main/kotlin/packit/config/JwtLoginConfig.kt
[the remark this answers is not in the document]
plietar: It does get populated from the properties in the `Environment`, but these can come from any number of sources, not just the built-in`application.properties`.

See https://docs.spring.io/spring-boot/reference/features/external-config.html

The simplest way of doing it is probably through the command line arguments.

api/app/src/main/kotlin/packit/service/JwtLoginService.kt
plietar: I think I'll keep it this way. The contents of the `TokenPolicy` is so tightly coupled with the implementation of the service it doesn't really make sense to move it elsewhere. The only way the decoupling makes sense IMO is if we make the TokenPolicy an interface and move all the logic into it, at which point the service is nothing much more than a for-loop. Seems like an unnecessary abstraction, especially given there's only one kind of TokenPolicy.

I've changed it from a `lateinit var` to a `val` though, since I realised that works just as well.

PR #132, opened 2024-09-20
Mrc-5787 Tasks logs page table

absternator:
This is the work for the tasks logs table as per [figma](https://www.figma.com/design/9mO1pIrQ3k4IlOqbTehybr/Packit?node-id=0-1&node-type=canvas&t=7VoZKUjA3rdE8WhQ-0).
There has been slight changes from figma: 
- icon is now matching task log page 
- column for user 
- branch and commit both showing as per mantra suggestion 
- Times are the created at times (as per what buildkite uses)
- button link for packetId with hover for info when available (would like feedback on this was not sure)

The table is also paginated and allows filtering by packetname currently. Future tickets will alow further filtering.

Also next ticket will be to poll this page 

Testing: 

Testing:

To See completed and running layouts:
- run multiple packet group from the runner in the Web UI... then click logs on the side bar
- when complete should be able to navigate to packet page

To see error layout:
- hit API submit with bad body, then copy the taskId you get back and in the web UI go to /runner/logs/{taskId}
Ex bad request: hit `http://localhost:8080/runner/run` and body
```
{   
 "name": "t2",
    "branch": "master",
    "hash": "06e9c200fb9c9bdc494ac2e4913216d8c86a43ce"
}
```
image
image


EmmaLRussell (commented):
Looks great! Couple of small UI comments:

When you're on the log page for a particular packet run, the sidebar menu is shown for the  Runner area, but neither page is selected (Run or Logs). This sort of makes sense since you're not on the Logs list page, but it looks a bit weird. Feels like it might make more sense to show the Logs item as active since we're on *a* logs page...

I think it would be good to make the packet links consistent between the Logs list page and the single log page. On one, it's an icon net to the packet id, and on the other it's a 'View' button which is very separated from the packet id. I think it would be better to have the icon for each - or just make the packet id itself a link (with 'Go to packet' hover hint).

I've just noticed we seem to be using American style dates ("9/23/2024") both in the packet group list and the logs pages. We usually go for UK type dates as that's where most of our users are.

Could we change "filter packet group names" to "Filter by packet group name" - I think that's slightly more precise. 

The list uses "Created" term, but packet log page says "Started". Are they the same thing, or is one queued time and the other is run start time? Feels like that could be more explicit, it's a little confusing - the details page should maybe have both?

Also "User" (list page) vs "Ran by" (details page)

Are we going add duration anywhere? I know you can see timestamps on the logs, including last log message so could work it out, but it could be useful for users to easily see how long a packet took to run.

M-Kusumgar (commented):
niceee, we're almost done! got a couple of nitpicks/comments on the UI:
* in dark mode, the circle icons (success, loading and error) have a black outline, i think it would be nice without that outline, i am assuming in light mode they might have a white outline but cant really see it because our background is white
* in light mode theres background color change for the row of the table you are hovering over, this is inconsistent with dark mode (i do like the hover bg color change)
* in the logs table, the first column icon is not vertically aligned in the centre of the table row, like the alignment is off between the icon and the packet group name + task id
* the go to packet icon usually suggests "open in new tab" in most ui, i think it is worth changing that icon, i was surprised when it didnt open a new tab
* parameter tags arent centred, like if theres only 3 parameters, i would expect them to be in the centre of the row rather than at the top
* i think the octocat was good for the github info in the runner page but for the commit hash in the table we should use the commit icon, the branch icon is also different on github:
* for branch this is the standard icon
image

* for commits:
image

absternator:
> When you're on the log page for a particular packet run, the sidebar menu is shown for the Runner area, but neither page is selected (Run or Logs). This sort of makes sense since you're not on the Logs list page, but it looks a bit weird. Feels like it might make more sense to show the Logs item as active since we're on _a_ logs page...
> 
> I've just noticed we seem to be using American style dates ("9/23/2024") both in the packet group list and the logs pages. We usually go for UK type dates as that's where most of our users are.
> 
> Could we change "filter packet group names" to "Filter by packet group name" - I think that's slightly more precise.
> 
> The list uses "Created" term, but packet log page says "Started". Are they the same thing, or is one queued time and the other is run start time? Feels like that could be more explicit, it's a little confusing - the details page should maybe have both?
> 
> Also "User" (list page) vs "Ran by" (details page)
> 
> Are we going add duration anywhere? I know you can see timestamps on the logs, including last log message so could work it out, but it could be useful for users to easily see how long a packet took to run.

1. currently the highlight matches the the path name and sidebar name... thus its not highlighted.. i tried matching by start by but then messes up in other places... so if its okay will leave for now?
2. it should be locale time.. mine is UK, I'm guessing your is USA time
image
3. done
4. Created at time is shown on table.. and on full log view depending on status.. eg. if queued would say queued on ...
3. makes sense right? as ran by is the action made by the user? and the table just shows user
5. the duration is can be seen on the full log page for task... or do you think its good idea to show on table as well?

absternator:
> * in dark mode, the circle icons (success, loading and error) have a black outline, i think it would be nice without that outline, i am assuming in light mode they might have a white outline but cant really see it because our background is white
> * in light mode theres background color change for the row of the table you are hovering over, this is inconsistent with dark mode (i do like the hover bg color change)
> * in the logs table, the first column icon is not vertically aligned in the centre of the table row, like the alignment is off between the icon and the packet group name + task id
> * the go to packet icon usually suggests "open in new tab" in most ui, i think it is worth changing that icon, i was surprised when it didnt open a new tab
> * parameter tags arent centred, like if theres only 3 parameters, i would expect them to be in the centre of the row rather than at the top
> * btw if we put comically long parameter values for the parameters report, it doesnt do word wrap so if you do like a really long name, the git info column for example goes off the screen, but also the logs dont word wrap either so the text goes off the edge of the container
>   image
> 
> image
> 
> * i think the octocat was good for the github info in the runner page but for the commit hash in the table we should use the commit icon, the branch icon is also different on github:


1. have updated icons to remove outer ring
2. it does change on hover its the behaviour of table
3.  good catch have aligned
4. have moved and updaetd as discuessed at end of table
5.as discussed its the behaiour of scroll area.. as its fixed height and needs the whole area
6. done word break
7. have updated icons. for git branch that was closes available from lucide icons(our library)
image

absternator:
image
have added created, started, finish time to details page

M-Kusumgar (commented):
I think that plus in the branch icon is a little misleading, i think the old one is better if we dont have the standard git one but feel free to merge after that!

edit: sorry another word wrap issue, i think the parameters on log for single run page need to be fit to some max height with scroll:
image

obviously these are comically long again but this is a legitimate issue if they have like 20 parameters with slightly longer names but yh feel free to either address this here or another ticket

absternator:
> I think that plus in the branch icon is a little misleading, i think the old one is better if we dont have the standard git one but feel free to merge after that!
> 
> edit: sorry another word wrap issue, i think the parameters on log for single run page need to be fit to some max height with scroll: image
> 
> obviously these are comically long again but this is a legitimate issue if they have like 20 parameters with slightly longer names but yh feel free to either address this here or another ticket

good catches to both done

PR #134, opened 2024-09-25
Rewrite WebSecurityConfig to use the lambda DSL.

plietar:
Spring Security has two forms of DSLs for configuring the HttpSecurity object, one based on method chaining and [one based on lambdas][lambda-dsl]. The latter is generally more readable, as it makes it clear the scope of each configuration block. In fact the method chaining syntax is deprecated in Spring Security 6.2, and [will be removed in 7.0][migration].

We were using an inconsistent mix of the two. This cleans it up and uses the lambda DSL throughout. The only exception to this is the `AuthenticationManager` configuration, which [for some reason doesn't have a lambda DSL][sec13003]

[lambda-dsl]: https://spring.io/blog/2019/11/21/spring-security-lambda-dsl
[migration]: https://docs.spring.io/spring-security/reference/migration-7/configuration.html
[sec13003]:  https://github.com/spring-projects/spring-security/issues/13003

### DIFF
--- a/api/app/src/main/kotlin/packit/security/WebSecurityConfig.kt
+++ b/api/app/src/main/kotlin/packit/security/WebSecurityConfig.kt
@@ -41,18 +41,14 @@ class WebSecurityConfig(
     ): SecurityFilterChain
     {
         httpSecurity
-            .cors().configurationSource(getCorsConfigurationSource())
-            .and()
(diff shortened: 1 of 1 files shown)


absternator (approved):
Nice cleanups!!! lgtm

PR #139, opened 2024-11-20
MRC-6053: Log user out when token expired

david-mears-2:
After the user had been logged in for longer than the expiry of their token, the app header was still showing the account dropdown, even while correctly redirecting the user to the login page.

To fix this, we log out the user (using removeUser()) when they fail authentication due to expiry time, so that localStorage is cleaned up.

You can test out the changes locally if you edit the 'exp' value in localStorage while logged in, and then try to reload a protected route.

### DIFF
--- a/app/src/app/components/login/Login.tsx
+++ b/app/src/app/components/login/Login.tsx
@@ -1,8 +1,14 @@
 import { UserAuthForm } from "./UserAuthForm";
+import { useSearchParams } from "react-router-dom";
 
 export default function Login() {
+  const [searchParams] = useSearchParams();
+  const sessionExpiryMsg = searchParams.get("info");
   return (
     <>
+      {sessionExpiryMsg && (
+        <div className="text-xs text-info justify-self-center">{sessionExpiryMsg}</div>
+      )}
       <div className="flex flex-col space-y-2 text-center">
         <h1 className="text-2xl font-semibold tracking-tight">Login to account</h1>
       </div>
--- a/app/src/lib/isAuthenticated.ts
+++ b/app/src/lib/isAuthenticated.ts
@@ -1,5 +1,7 @@
 import { AuthConfig } from "../app/components/providers/types/AuthConfigTypes";
 import { UserState } from "../app/components/providers/types/UserTypes";
 
+export const authExpiryIsValid = (user: UserState | null): boolean => !!user?.exp && user?.exp * 1000 >= Date.now();
+
 export const isAuthenticated = (authConfig: AuthConfig | null, user: UserState | null): boolean =>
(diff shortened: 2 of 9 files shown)


david-mears-2:
Not sure why checks aren't triggered...

app/src/app/components/login/Login.tsx
```diff
@@ -1,8 +1,14 @@
 import { UserAuthForm } from "./UserAuthForm";
+import { useSearchParams } from "react-router-dom";
 
 export default function Login() {
+  const [searchParams] = useSearchParams();
+  const sessionExpiryMsg = searchParams.get("info");
```
absternator, line 6:
```suggestion
  const sessionExpiryMessage = searchParams.get("expiryMessage");
```

app/src/lib/isAuthenticated.ts
```diff
@@ -1,5 +1,7 @@
 import { AuthConfig } from "../app/components/providers/types/AuthConfigTypes";
 import { UserState } from "../app/components/providers/types/UserTypes";
 
+export const authIsPreExpiry = (user: UserState | null): boolean => !!user?.exp && user?.exp * 1000 >= Date.now();
```
absternator, line 4: at this point token is expired? so more suitable name would be `isAuthTokenExpired`

absternator (commented):
LGTM!!! works great thanks for the fix.. just left some naming related comments

app/src/app/components/login/Login.tsx
david-mears-2: I'm using the more generic search param 'info' by analogy to how elsewhere we're using the search param 'success' or 'error'. I'm thinking of alerts/notices as coming in three or four types - success (green), error (red), info (blue) and warning (yellow); we currently only use the first two.

absternator: cool if that's the case should the variable be renamed? like `infoMessage`  or something

app/src/lib/isAuthenticated.ts
absternator: ohh yesss... yup maybe checking expiry is better then?  or maybe `authIsNotExpired` is better then if you want to check not expired

absternator (approved):
cool just those 2 comments on renaming but feel free merge after addressed

app/src/app/components/login/Login.tsx
david-mears-2: My model is
```ts
  const updatePasswordSuccess = searchParams.get("success");
```
in BasicUserAuthForm.tsx

So I'll rename the variable so it mentions 'info', but I think it makes sense to be specific about what kind of message we're expecting in this template (a message about session expiry)

PR #140, opened 2024-11-20
Spelling corrections

david-mears-2:
The noun is 'login' and the verb is 'log in'. https://grammarist.com/spelling/log-in-login/


absternator (approved):
LGTM thanks for fixing

PR #142, opened 2024-11-25
MRC-6040: Import display name from orderly custom metadata, store it in DB, and display it

david-mears-2:
In this branch there's a new packet group in `/demos`, since none of the existing ones were making use of the `description.display_name` orderly property.

That property is referenced during the scheduled job that imports packets into the proxy db, so that the db column 'display_name' is now being used for the first time rather than just duplicating the name. When there's no display name, the column defaults to the name.

Then, the home page (the packet group index page) is now able to display the display name for packet groups, which is found by the API looking up the latest display name at the same time as it looks up the latest ID and latest time. The api also changes to allow you to filter by display_name, not just by name, from this page. The display name is also displayed on the packet group summary page (PacketGroup.tsx) and on the packet page and subpages (PacketHeader.tsx). In all those places, the component will render a little differently depending on whether the display name differs from the name.

I also introduce displaying the `description.long` orderly property (a lengthy description) in two places (the packet group page PacketGroup.tsx, and the packet summary page PacketDetails.tsx). This does not use the db, it uses a request for packet metadata.

# Screenshots

## Home page showing packet groups with and without display names

Screenshot from 2024-12-04 17-04-48

## Home page showing filtering on name

Screenshot from 2024-12-04 17-04-58

## Home page showing filtering on display name

Screenshot from 2024-12-04 17-05-08

## Packet group page for packets that have display names and descriptions

Screenshot from 2024-12-04 17-05-21

## Packet group page for packets that have neither a display name nor a description

Screenshot from 2024-12-04 17-06-33

## Packet header (and summary page with description) for packets with display names and descriptions

Screenshot from 2024-12-04 17-07-36

## Packet header (and summary page with description) for packets with neither a display names nor a description (no change)

Screenshot from 2024-12-04 17-07-50

## NB

I have *not* made changes to the runner logs page, which is shown below, as it would be fairly in depth (requires either another request to get the packet metadata from outpack server, or a db query to get the packet display name) and I'm not sure it's needed. But I could do it in another PR if required.

image


api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -44,14 +45,21 @@ class BasePacketService(
     private val outpackServerClient: OutpackServer
 ) : PacketService
 {
+    private fun getDisplayNameForPacket(packet: OutpackMetadata): String {
+        val metadata = outpackServerClient.getMetadataById(packet.id)
+        val orderlyMetadata = metadata?.custom?.get("orderly") as? Map<*, *>
+        val description = orderlyMetadata?.get("description") as? Map<*, *>
+        return description?.get("display") as? String ?: packet.name
+    }
+
```
EmmaLRussell, line 54: I'm surprised you need to do this - is the custom metadata not already in the metadata being returned for all new packets in line 59? If it's not it should be. Maybe it is but those values ujust aren't being passed out by `getMetadata`?  We shouldn't need to do a new call to the client per packet.

app/src/app/components/contents/PacketGroup/PacketTable.tsx
```diff
@@ -15,6 +15,7 @@ export const PacketTable = () => {
   const { packetName } = useParams();
   const [pageNumber, setPageNumber] = useState(0);
 
+  // Rename useGetPacketGroups to useGetPacketGroupPages, and packetGroups to packetGroupPages?
```
EmmaLRussell, line 18: Why?

david-mears-2: When you iterate over those things, the elements aren't packet groups, they are pages of packet groups. So later in the file we have eg 'packetGroups.first' which looks like it means 'the first packet group' but actually means 'the first page of packet groups'.

api/app/src/main/kotlin/packit/service/PacketService.kt
david-mears-2: It does seem inefficient doesn't it. Line 59 corresponds to calling `/packit/metadata` on the outpack server, which returns only id, name, parameters, and time, contrary to the documentation of outpack server: https://github.com/mrc-ide/outpack_server/pull/68

david-mears-2: So we'll change the outpack server

app/src/app/components/contents/PacketGroup/PacketTable.tsx
david-mears-2: Oh, I mistook this file for being `PacketGroupSummaryList.tsx`. In fact, each item returned by useGetPacketGroups (each so-called packetGroup) is a _page of packets_, not a packet group nor a page of packet groups. So these things should be renamed to useGetPacketPages and packetPages respectively.

david-mears-2: or more simply, `useGetPackets` and `pages`

api/app/src/main/kotlin/packit/service/PacketService.kt
david-mears-2: Changed outpack server in https://github.com/mrc-ide/outpack_server/pull/69

david-mears-2:
I wonder if we have any CI checks designed to notice when test coverage is too low -- I know my last commit is changing a lot of things that demand new unit tests!

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -41,15 +41,15 @@ class PacketController(private val packetService: PacketService)
         )
     }
 
-    @GetMapping("/packetGroupSummary")
-    fun getPacketGroupSummary(
+    @GetMapping("/packetGroupSummaries")
```
david-mears-2, line 44: This is one of a few renamings i did for increased precision. The endpoint returns a summary for each packet group (or, if a name to filter on is provided, a summary for each _matching_ packet group).

api/app/src/test/kotlin/packit/integration/repository/PacketRepositoryTest.kt
```diff
@@ -119,38 +119,41 @@ class PacketRepositoryTest : RepositoryTest()
     @Test
     @WithAuthenticatedUser(authorities = ["packet.read"])
 
-    fun `can get right order and data expected from findPacketGroupSummaryByName`()
+    fun `can get right order and data expected from findPacketGroupSummaryBySearchString`()
```
david-mears-2, line 122: Note to self should say get not find

api/app/src/test/kotlin/packit/integration/repository/PacketRepositoryTest.kt
```diff
@@ -159,6 +162,32 @@ class PacketRepositoryTest : RepositoryTest()
         assertEquals(result[0].name, "test4")
         assertEquals(result[0].latestId, "20170819-164847-7574113a")
         assertEquals(result[0].latestTime, now + 4)
+        assertEquals(result[0].latestDisplayName, "test name4")
+        assertEquals(result[0].packetCount, 2)
+    }
+
+    @Test
+    @WithAuthenticatedUser(authorities = ["packet.read"])
+    fun `can filter correctly when calling findPacketGroupSummaryBySearchString with part of a packet display name`()
```
david-mears-2, line 171: Note to self fix name

app/src/app/components/contents/PacketGroup/PacketTable.tsx
```diff
@@ -4,18 +4,17 @@ import { Skeleton } from "../../Base/Skeleton";
 import { DataTable } from "../common/DataTable";
 import { ErrorComponent } from "../common/ErrorComponent";
 import { Pagination } from "../common/Pagination";
-import { useGetPacketGroups } from "./hooks/useGetPacketGroups";
+import { useGetPacketPages } from "./hooks/useGetPacketPages";
```
david-mears-2, line 7: Another renaming: what’s returned is a paginated list of packets, not a packet group.

app/src/app/components/contents/PacketGroup/hooks/useGetPacketPages.ts
```diff
@@ -3,14 +3,16 @@ import appConfig from "../../../../../config/appConfig";
 import { fetcher } from "../../../../../lib/fetch";
 import { PageablePackets } from "../../../../../types";
 
-export const useGetPacketGroups = (packetName: string | undefined, pageNumber: number, pageSize: number) => {
+export const useGetPacketPages = (packetGroupName: string | undefined, pageNumber: number, pageSize: number) => {
```
david-mears-2, line 6: Note to self, file name is not in sync with function name. In the back end this queries on packets so function rename is questionable.

app/src/app/components/routes/Router.tsx
```diff
@@ -40,7 +40,7 @@ export function Router() {
             </Route>
             {/* <Route path="run-workflow" element={<WorkflowRunner />} /> */}
             {/* <Route path="documentation" element={<ProjectDocumentation />} /> */}
-            <Route path="/:packetName" element={<PacketGroup />} />
+            <Route path="/:packetGroupName" element={<PacketGroup />} />
```
david-mears-2, line 43: Questionable

app/tailwind.config.js
```diff
@@ -56,6 +56,9 @@ export const theme = {
       sans: ["Graphik", "sans-serif"],
       serif: ["Merriweather", "serif"]
     },
+    fontSize: {
```
david-mears-2, line 59: Remove, not used

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -44,14 +45,33 @@ class BasePacketService(
...
 {
+    // Return the display name for a packet if its custom metadata schema conforms to the orderly schema and contains
+    // a display name.
+    // Also check for 'display name' keys that may exist in non-orderly outpack custom schemas.
+    // Falls back to name if no display name.
+    private fun getDisplayNameForPacket(packet: OutpackMetadata): String {
+        val orderlyMetadata = packet.custom?.get("orderly") as? Map<*, *>
+        val description = orderlyMetadata?.get("description") as? Map<*, *>
+        val orderlyDisplayName = description?.get("display") as? String
+        return if (!orderlyDisplayName.isNullOrBlank()) {
```
absternator, line 56: can use kotlin when here for concise 
```suggestion
       return when {
        !orderlyDisplayName.isNullOrBlank() -> orderlyDisplayName
        packet.custom?.get("display") is String -> packet.custom["display"] as String
        packet.custom?.get("display_name") is String -> packet.custom["display_name"] as String
        else -> packet.name
    }
```

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -44,14 +45,33 @@ class BasePacketService(
     private val outpackServerClient: OutpackServer
 ) : PacketService
 {
+    // Return the display name for a packet if its custom metadata schema conforms to the orderly schema and contains
+    // a display name.
+    // Also check for 'display name' keys that may exist in non-orderly outpack custom schemas.
+    // Falls back to name if no display name.
+    private fun getDisplayNameForPacket(packet: OutpackMetadata): String {
```
absternator, line 52: also didnt see unit tests for this method..... change to internal to be able to test and expose

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
```diff
@@ -1,19 +1,48 @@
-import { useParams } from "react-router-dom";
 import { PacketTable } from "./PacketTable";
+import {useParams} from "react-router-dom";
```
absternator, line 2: formatting issues? there is no space b/w word and brackets

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
```diff
@@ -1,19 +1,48 @@
...
+import {useGetPacketGroupSummaries} from "../explorer/hooks/useGetPacketGroupSummaries";
+import {HttpStatus} from "../../../../lib/types/HttpStatus";
+import {Unauthorized} from "../common/Unauthorized";
+import {ErrorComponent} from "../common/ErrorComponent";
+import {useGetPacketById} from "../common/hooks/useGetPacketById";
 
 export const PacketGroup = () => {
   const { packetName } = useParams();
+  const { packetGroupSummaries, error: packetGroupError } =
```
absternator, line 11: there are now 3 network requests here including in Packettable... surely it can all be combined into 1... does pulling the one from packettable work? that should get you all the data you need

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
```diff
@@ -1,19 +1,48 @@
...
+    useGetPacketGroupSummaries(0, 1, packetName as string);
+  const latestPacketId = packetGroupSummaries?.content[0]?.latestId;
```
absternator, line 13: also just realised since display name is packet specific it doesn't make sense to show it at the top? since not all packets will have the same display name?

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
```diff
@@ -1,19 +1,48 @@
...
+  const { packet: latestPacket, error: packetError } = useGetPacketById(latestPacketId)
+  const packetGroupDisplayName = packetGroupSummaries?.content[0]?.latestDisplayName || "";
+  const latestDescription = latestPacket?.custom?.orderly.description.long;
+
+  if ([packetError?.status, packetGroupError?.status].includes(HttpStatus.Unauthorized)) return <Unauthorized />;
+  if (packetGroupError || !packetGroupDisplayName) {
```
absternator, line 19: these get  cleaned up when using just 1 API call....

app/src/app/components/contents/PacketGroup/hooks/useGetPacketPages.ts
```diff
@@ -3,14 +3,16 @@ import appConfig from "../../../../../config/appConfig";
 import { fetcher } from "../../../../../lib/fetch";
 import { PageablePackets } from "../../../../../types";
 
-export const useGetPacketGroups = (packetName: string | undefined, pageNumber: number, pageSize: number) => {
+export const useGetPacketPages = (packetName: string | undefined, pageNumber: number, pageSize: number) => {
+  const url = `${appConfig.apiUrl()}/packets/${packetName}?pageNumber=${pageNumber}&pageSize=${pageSize}`
+
   const { data, isLoading, error } = useSWR<PageablePackets>(
-    `${appConfig.apiUrl()}/packets/${packetName}?pageNumber=${pageNumber}&pageSize=${pageSize}`,
+    packetName ? url : null,
```
absternator, line 10: nice good catch!!!

app/src/app/components/contents/common/hooks/useGetPacketById.ts
```diff
@@ -5,12 +5,17 @@ import { PacketMetadata } from "../../../../../types";
 
 export const useGetPacketById = (packetId: string | undefined) => {
   const { data, isLoading, error } = useSWR<PacketMetadata>(
-    `${appConfig.apiUrl()}/packets/metadata/${packetId}`,
+    packetId ?
```
absternator, line 8: good catch nice again

app/src/app/components/contents/common/hooks/useGetPacketById.ts
```diff
@@ -5,12 +5,17 @@ import { PacketMetadata } from "../../../../../types";
...
+      `${appConfig.apiUrl()}/packets/metadata/${packetId}` : null,
     (url: string) => fetcher({ url })
   );
 
+  const displayName = data?.custom?.orderly?.description?.display;
+
+  const packet = displayName ? { ...data, displayName } : data;
```
absternator, line 15: there is displayName in the interface? whats that for?
.. also instead of spreading you can probs just do `{data, displayName: data?.custom?.orderly?.description?.display;` ... it will be undefined anyways If not there

app/src/app/components/contents/packets/PacketHeader.tsx
```diff
@@ -1,13 +1,27 @@
...
 
-export default function PacketHeader({ packetName, packetId }: PacketHeaderProps) {
-  return (
-    <div>
-      <h2 className="text-2xl font-bold tracking-tight">{packetName}</h2>
-      <p className="text-muted-foreground">{packetId}</p>
-    </div>
-  );
-}
+export default function PacketHeader({ displayName, packetName, packetId }: PacketHeaderProps) {
```
absternator, line 7: possoble to make displayName to allow undefined as well... saves you having to do displayName ?? "" everywhere.. and the check can just be if displayName

app/src/app/components/contents/packets/PacketHeader.tsx
```diff
@@ -1,13 +1,27 @@
...
+  if (displayName.length === 0 || displayName === packetName) {
+    return (
+      <div>
+        <h2 className="text-2xl font-bold tracking-tight">{packetName}</h2>
+        <p className="text-muted-foreground">{packetId}</p>
+      </div>
+    );
+  } else {
```
absternator, line 15: dont else :smile:

app/src/msw/handlers/packetGroupSummaryHandlers.ts
```diff
@@ -1,11 +1,14 @@
...
-
-const packetGroupSummaryIndexUri = `${appConfig.apiUrl()}/packets/packetGroupSummary`;
+import { mockPacketGroupSummaries, mockPacketGroupSummariesFiltered } from "../../tests/mocks";
 
 export const packetGroupSummaryHandlers = [
-  rest.get(packetGroupSummaryIndexUri, (req, res, ctx) => {
-    return res(ctx.json(mockPacketGroupSummary));
+  rest.get(`${appConfig.apiUrl()}/packets/packetGroupSummaries`, (req, res, ctx) => {
+    const url = new URL(req.url)
+    if (url.searchParams.get("filterName")) {
```
absternator, line 8: nice!!!

absternator (commented):
Great work works like a charm!! just left some comments.


just some extras:
- display description should probably match font (text-muted)
- in general avoid multiple API calls in a component 
-

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
```diff
@@ -24,24 +24,26 @@ interface PacketRepository : JpaRepository<Packet, String>
...
+                ) as RankedData
+                WHERE row_num = 1 AND (name ILIKE %?1% OR display_name ILIKE %?1%)
                 ORDER BY start_time DESC
          """,
-        countQuery = "SELECT count(distinct name) from packet WHERE name ILIKE  %?1%",
+        countQuery = "SELECT count(distinct name) from packet WHERE name ILIKE %?1%",
         nativeQuery = true
     )
-    fun findPacketGroupSummaryByName(filterName: String): List<PacketGroupSummary>
+    fun getPacketGroupSummariesBySearchString(filterName: String): List<PacketGroupSummary>
```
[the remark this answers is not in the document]
david-mears-2: The idea is that you can filter by name or display_name, so we shouldn't privelege 'name' by putting it in the function name. Do you prefer '...bySearchFilter'?

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
david-mears-2: Which formatter is meant to catch this?

api/app/src/main/kotlin/packit/service/PacketService.kt
david-mears-2: It's tested by the tests of importPackets. I think private functions are allowed, and that they shouldn't be separately unit tested.

david-mears-2: Because I should test the public interface ('guaranteeing' some level of stability for it), not the implementation details (which should be allowed to be changeable / not stable)

david-mears-2:
```kt
val expectedDisplayNames = mapOf(
            "20240101-090000-4321gaga" to "test", // This packet has no display name, so should fall back to its name.
            "20190203-120000-1234dada" to "test name (latest display name)",
            "20190403-120000-1234dfdf" to "test2 name"
        )
```

That's the bit that tests the 'fall back' condition.

I don't bother testing for the non-orderly (outpack) conditions that speculatively check for keys called 'display', but could be persuaded to, just makes the test suite more verbose.

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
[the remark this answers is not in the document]
EmmaLRussell: Or `getFilteredPackerGroupSummaries`..? `filterName` parameter could maybe just be `filter`?

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
EmmaLRussell: I'd have thought prettier would have picked that up.

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
EmmaLRussell: We're going to show the display name of the most recent packet.

app/src/app/components/contents/common/hooks/useGetPacketById.ts
david-mears-2: That change makes sense.

Yes, in the PacketMetadata interface, there is a displayName property. You added it in November 2023! And now we're using it for the first time, in three components, all of which use it to pass it as a prop to the PacketHeader.

david-mears-2: Actually no sorry you didn't add it, you just changed the indentation... it was added here https://github.com/mrc-ide/packit/commit/fc88d6393b336908f848cd63b1aa096371d42ad4

app/src/app/components/contents/packets/PacketHeader.tsx
david-mears-2: Fair point. I'll still keep the verbose check in this file though because we need to see if the display name matches the name.

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
```diff
@@ -1,19 +1,48 @@
...
+import {useGetPacketGroupSummaries} from "../explorer/hooks/useGetPacketGroupSummaries";
+import {HttpStatus} from "../../../../lib/types/HttpStatus";
+import {Unauthorized} from "../common/Unauthorized";
+import {ErrorComponent} from "../common/ErrorComponent";
+import {useGetPacketById} from "../common/hooks/useGetPacketById";
 
 export const PacketGroup = () => {
   const { packetName } = useParams();
+  const { packetGroupSummaries, error: packetGroupError } =
+    useGetPacketGroupSummaries(0, 1, packetName as string);
```
EmmaLRussell, line 12: Is this just calling the search endpoint? It feels like it would be more efficient to have an endpoint which only gets a single packet group with the known name?

EmmaLRussell (commented):
I like the styling on this! Seems pretty intuitive to me. 

> I have not made changes to the runner logs page, which is shown below, as it would be fairly in depth (requires either another request to get the packet metadata from outpack server, or a db query to get the packet display name) and I'm not sure it's needed. But I could do it in another PR if required.

It would definitely be one for another PR. Not convinced it's needed.

I feel like the text "Here's a list of all packets for the packet group 'Packet with description':" is not really needed, and is now just taking up space...

As discussed, maybe it would make sense to save the display name of the packet group on packet ingestion, from the latest packet per group. That might reduce the number of queries we need to do.

david-mears-2:
> display description should probably match font (text-muted)

I will make resolve the inconsistency in the other direction:

image

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
david-mears-2: Cool, I enabled prettier in my code editor and it picked this up.

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
```diff
@@ -24,24 +24,26 @@ interface PacketRepository : JpaRepository<Packet, String>
...
                         COUNT(id) OVER (PARTITION BY name) AS id_count 
                     FROM packet 
-                ) as RankedData 
-                WHERE row_num = 1 AND name ILIKE %?1% 
+                ) as RankedData
+                WHERE row_num = 1 AND (name ILIKE %?1% OR display_name ILIKE %?1%)
                 ORDER BY start_time DESC
          """,
-        countQuery = "SELECT count(distinct name) from packet WHERE name ILIKE  %?1%",
+        countQuery = "SELECT count(distinct name) from packet WHERE name ILIKE %?1%",
```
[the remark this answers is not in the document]
david-mears-2: Good point. I'm not sure what it's used for either - @absternator ?

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -44,14 +45,33 @@ class BasePacketService(
...
+        val orderlyMetadata = packet.custom?.get("orderly") as? Map<*, *>
+        val description = orderlyMetadata?.get("description") as? Map<*, *>
+        val orderlyDisplayName = description?.get("display") as? String
+        return if (!orderlyDisplayName.isNullOrBlank()) {
+            orderlyDisplayName
+        } else if (packet.custom?.get("display") is String) {
+            packet.custom["display"] as String
```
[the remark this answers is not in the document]
david-mears-2: Ah yes - I forgot about the key that tells you which schema you're in. So I should in fact unit test this behaviour.

api/app/src/main/kotlin/packit/repository/PacketRepository.kt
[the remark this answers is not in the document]
david-mears-2: I see
> Defines a special count query that shall be used for pagination queries to lookup the total number of elements for a page. If none is configured we will derive the count query from the original query or countProjection() query if any.

So I'm inclined to just remove it if it falls back to the original query?

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
david-mears-2: There's no guarantee that there is a 'single' packet group with that name (also, the column isn't unique), so we have to do a search.

david-mears-2: Yes.
The new endpoint would need to also return the id of the latest packet, meaning its quite a lot of extra code just for a neater call here.

david-mears-2: Too late for the separate ticket I'm afraid. Added a new endpoint that gives back the single packet group's display name, along with the needed id of the latest packet.

api/app/src/main/kotlin/packit/controllers/PacketGroupController.kt
```diff
@@ -4,15 +4,17 @@ import org.springframework.data.domain.Page
...
 import org.springframework.web.bind.annotation.RequestParam
 import packit.model.PageablePayload
 import packit.model.dto.PacketGroupDto
+import packit.model.dto.PacketIdAndDisplayName
 import packit.model.toDto
 import packit.service.PacketGroupService
 
 @Controller
-@RequestMapping("/packetGroup")
+@RequestMapping("/packetGroups")
```
david-mears-2, line 17: by convention I think resources are plural in CRUD; in any case we already use the plural form for the packets controller.

david-mears-2:
Main changes since last reviews:
* I added a 'latest_display_name' column to the packet **_group_** table, which is populated when packets are ingested.
* I added an endpoint (and various related gubbins like repository methods) that, when given the name of a packet group, provides its display name and the id of the latest packet with the same name. That was for use by PacketGroup.tsx and the benefit is in having SQL that can look up by name rather than rely on the LIKE function.

Neither of these allow us to have fewer network calls in the PacketGroup component.

app/src/app/components/contents/PacketGroup/PacketGroup.tsx
david-mears-2: I ended up implementing a new endpoint that would give us everything needed by the PacketGroup component: given a packet group name, return the display name and the latest packet's description. So now this only makes one network call! (Except PacketTable still makes a separate call for different info)

api/app/src/main/kotlin/packit/model/dto/PacketGroupDisplay.kt
```diff
@@ -0,0 +1,11 @@
...
+
+interface PacketGroupDisplay {
+    val displayName: String
+    val packetDescription: String?
+}
+
+data class PacketGroupDisplayImpl(
+    override val displayName: String,
+    override val packetDescription: String?
+) : PacketGroupDisplay
```
EmmaLRussell, line 11: We wouldn't usually bother with an interface for a data class .

api/app/src/main/kotlin/packit/model/dto/PacketGroupDto.kt
```diff
@@ -2,5 +2,6 @@ package packit.model.dto
 
 data class PacketGroupDto(
     val name: String,
+    val latestDisplayName: String,
```
EmmaLRussell, line 5: I'm a little bit uneasy about the inconsistency between "displayName" for `PacketGroupDisplay` and "latestDisplayName" for `PacketGroupDto`. I guess there's potentially a subtle difference here which justifies the discrepancy in that `PacketGroupDisplay` is an ephemeral sort of thing so its display name is just the display name now, whereas `PacketGroup` persists and can have different "latestDisplayNames" over time..? But i think I'd still be inclined to just use the same prop for both to keep things simple.

api/app/src/main/kotlin/packit/model/dto/PacketGroupDisplay.kt
```diff
@@ -0,0 +1,11 @@
+package packit.model.dto
+
+interface PacketGroupDisplay {
+    val displayName: String
+    val packetDescription: String?
```
EmmaLRussell, line 5: Is this the latest packet description, in the same way the display name is the latest packet display name?  I think just "description" would probably be a more consistent name. 

Also - should latest description be in the packet group table as well as latest display name?

api/app/src/main/kotlin/packit/service/PacketGroupService.kt
```diff
@@ -2,25 +2,42 @@ package packit.service
...
+    override fun getPacketGroupDisplay(name: String): PacketGroupDisplay{
+        val latestPacketId = packetGroupRepository.findLatestPacketIdForGroup(name)?.id
+        val packetGroup = packetGroupRepository.findByName(name)
+        if (latestPacketId == null || packetGroup == null) {
+            throw PackitException("doesNotExist", HttpStatus.NOT_FOUND)
+        }
+        val packetOrderlyMetadata = packetService.getMetadataBy(latestPacketId).custom?.get("orderly") as? Map<*, *>
+        val packetLongDescription = (packetOrderlyMetadata?.get("description") as? Map<*, *>)?.get("long") as? String
+        return PacketGroupDisplayImpl(packetGroup.latestDisplayName, packetLongDescription)
+    }
```
EmmaLRussell, line 42: I'm a bit confused by what's going on here- it's going to outpack server to fetch metadata at the time when the endpoint is called? I hadnt imagined that would happen, but rather that during `importPackets`. it would check for any new packet display names and update the packet group table accordingly - and then we'd return that display information when requested by front end. 

Oh... it looks like you are doing that. So this get from outpack server is just for descriptions? Why not just put those in the packet group table too?

api/app/src/main/resources/db/migration/V20241206__add_latest_display_name_column_to_packet_group.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE "packet_group"
+    ADD COLUMN "latest_display_name" text DEFAULT '';
+
+ALTER TABLE "packet_group"
+    ALTER COLUMN "latest_display_name" SET NOT NULL;
+
+ALTER TABLE "packet_group"
+    ALTER COLUMN "latest_display_name" DROP DEFAULT;
```
EmmaLRussell, line 8: We always want the `latest_display_name`columns to be populated with non-empty values  don't we? So perhaps here we should also update any existing packet groups with display name. However if we were going to do that properly (not just copying over latest packet name) we'd need to get into the metadata, which would need to be done in code (getting metadata from outpack server). One for another ticket?

app/src/app/components/contents/PacketGroup/PacketTable.tsx
EmmaLRussell: But what we're showing here is a single page of packet groups, isn't it? It seems odd to send "packetPages.content" into the data table as if it's going to display multiple pages...?
Would it be better as packetPage singular?

app/src/app/components/contents/PacketGroup/hooks/useGetPacketPages.ts
```diff
@@ -3,14 +3,16 @@ import appConfig from "../../../../../config/appConfig";
 import { fetcher } from "../../../../../lib/fetch";
 import { PageablePackets } from "../../../../../types";
 
-export const useGetPacketGroups = (packetName: string | undefined, pageNumber: number, pageSize: number) => {
+export const useGetPacketPages = (packetName: string | undefined, pageNumber: number, pageSize: number) => {
```
EmmaLRussell, line 6: ..similarly here, would `useGetPacketPage` be slightly more accurate?

EmmaLRussell (commented):
Good stuff. 

Mostly nitpicks about naming. I think we might also need to think more about migration for existing packet groups, since there are some in the wild. And I'm wondering why we don't treat packet group description in the same way we do display name. 

It would be good to show the display name in the breadcrumb - in another ticket!
image

I'd be in favour of doing any substantive changes in other tickets, as I feel like this branch has become quite a catch-all already, so it would be good to merge and then refine... though we should consider any deployment schedules if we do that (in context of migrations).

david-mears-2:
> It would be good to show the display name in the breadcrumb - in another ticket!

Created: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6115/Use-the-packet-group-display-name-in-the-breadcrumbs

api/app/src/main/kotlin/packit/model/dto/PacketGroupDisplay.kt
david-mears-2: Renamed

api/app/src/main/kotlin/packit/model/dto/PacketGroupDto.kt
david-mears-2: Renamed all to 'latest...'

api/app/src/main/kotlin/packit/service/PacketGroupService.kt
david-mears-2: I guess I thought display names would be much more commonly referred to than descriptions, justifying the db column, but in practice that turned out not to be true.

david-mears-2:
I see two possible ways forward with regard to the two packet group (henceforth 'group') properties, display name and description, that are ultimately drawn from their latest packet.

1) Treat groups' descriptions like this branch currently treats display_names: add a db column for latest description, and add a script (in a future PR) for back-filling the new columns with reference to the metadata from outpack_server, run that script, and remove it.

2) Revert the db changes and have the /packetGroups/{name}/display endpoint query only the outpack_server for both display_name and description (as it currently does in this branch for description only); this prevents any complication / work about migrating existing groups using a script that queries outpack_server.

Here are all of the places when we'll want to refer in the front end to those properties of groups:
a) Breadcrumbs (display name)
b) Group page (display name and description)

(A bit counter-intuitively, the home page's table of groups makes no reference to packet_group db table, only packet db table.)

Other features we might bring in that relate to group display names:

i) Parameterised display names, e.g. (e.g., "diagnostic report for ${disease}"), as mentioned [here](https://resideglobal.slack.com/archives/C03FU2C9PMG/p1733485896428619?thread_ts=1733416423.989409&cid=C03FU2C9PMG). This would vary per packet in the group, and, in this situation, it would make less sense to privilege the latest packet over others; we might want to use the (constant) _name_ for the packet group instead of the display name. This reduces the need for a (latest_)display_name column; we'd probably want to get rid of that column in order to avoid confusing ourselves about the canonical display name for any group.

ii) On the packet group page, an info / tooltip warning that the packets' display names differ, as described in [the same thread](https://resideglobal.slack.com/archives/C03FU2C9PMG/p1733493113856579?thread_ts=1733416423.989409&cid=C03FU2C9PMG) and ticketised [here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6117/Packet-group-page-warn-user-when-packets-display-names-differ). This implies that the packet group display page would need to check the display_name on each of its member packets anyway, obviating the savings achieved by putting a column on the group table.

There is a cost to using our proxy db less: more outward calls to outpack_server. Breadcrumbs would be rendered a little more slowly.

But overall I feel that the database columns don't gain us much, since the latest_display_name column would be used only in two components, and the description column only in one. It doesn't feel worth the added complexity of cacheing the info in the group table, potentially to unpick the code later when we try (i) or (ii).

Thoughts?

david-mears-2:
Closing in favour of https://github.com/mrc-ide/packit/pull/150

Issue #137, opened 2024-11-15
Add more explanation of what this app is for


david-mears-2:
As a newcomer to VIMC I'm not sure if there is a more in-depth explanation of what packit is somewhere online, besides the one sentence "Web app for serving outpack and/or orderly2 metadata" at the top of this repo's README. If there's more explanation somewhere else then we could link to it from this README.

Perhaps a screenshot of the app in action would help to convey what kind of metadata you are talking about? (Or does the app even have a front end or is it just a pure server?)

PR #143, opened 2024-12-03
Update README.md

david-mears-2:
This is an attempt to help past me, who was very confused about everything VIMC-related, and couldn't work out what packit was in relation to orderly. I think I was particularly led astray by the word 'serving' as this is not the main server (outpack_server is).

Fixes https://github.com/mrc-ide/packit/issues/137

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # packit
 
-Web app for serving outpack and/or orderly2 metadata.
+Web app providing an interface for outpack (and/or orderly2) metadata, such as packets' timestamps, parameters, associated files, groups, etc. Intended to be used as a portal by researchers, to manage and track their packets, and by those who requested the analyses, to view packet outputs. Here's a screenshot of the app in action to give an impresssion:
+
+![Screenshot from 2024-12-03 18-14-12](https://github.com/user-attachments/assets/77f79ded-a900-48d1-9c73-fbae6131fd5a)
+
 
 ## Pre-requisites
 
--- a/README.md
+++ b/README.md
@@ -1,9 +1,18 @@
 # packit
 
-Web app providing an interface for outpack (and/or orderly2) metadata, such as packets' timestamps, parameters, associated files, groups, etc. Intended to be used as a portal by researchers, to manage and track their packets, and by those who requested the analyses, to view packet outputs. Here's a screenshot of the app in action to give an impresssion:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,9 @@
 # packit
 
-Web app for serving outpack and/or orderly2 metadata.
+Web app providing an interface for outpack (and/or orderly2) metadata, such as packets' timestamps, parameters, associated files, groups, etc. Intended to be used as a portal by researchers, to manage and track their packets, and by those who requested the analyses, to view packet outputs. Here's a screenshot of the app in action to give an impresssion:
+
+![Screenshot from 2024-12-03 18-14-12](https://github.com/user-attachments/assets/77f79ded-a900-48d1-9c73-fbae6131fd5a)
+
```
EmmaLRussell, line 6: - "provide an interface" is maybe a bit too vague. It can be used as a web app but can also forward requests to outpack server which is also worth mentioning I think.
- "outpack and/or orderly2" is a little misleading. It's always outpack and it might be orderly. orderly2 will shortly be reverted back to just "orderly" so I think we should just gloss over that.. 
- And it provides access to data as well as metadata, and can also be used to run packets. 
- I'm loath to include screenshots in web repos as they go stale so quickly. 

So how about expanding to something like:

```suggestion
Web application for surfacing packet data and metadata. Packets are sets of files with related metadata conforming to the [outpack](https://github.com/mrc-ide/outpack/) format. Packets are produced as the result of running analysis scripts, typically using a package which supports outpack format, such as [orderly](https://github.com/mrc-ide/orderly2) or [pyorderly](https://github.com/mrc-ide/pyorderly), which also add their own custom metadata to the packets produced. 

Packet metadata includes:
- timestamp
- parameter
- display name
- file metadata e.g. whether a file is input ("resource") or output ("artefact")

Packet data are the files included in the packet, which can be downloaded through the application. 

The application can be used through the browser as a portal by researchers, to manage and track their packets, and by those who requested the analyses, to view packet outputs. Packets can also be created through the app by running an analysis. 

Packit can also be used programmatically through its `/outpack` route, which forwards requests to [outpack server](https://github.com/mrc-ide/outpack_server). 
```

EmmaLRussell (commented):
This is great. If we're going to expand this, I feel like we could provide even more detail..

david-mears-2:
I'm merging your entire suggestion. Thanks!!

PR #144, opened 2024-12-04
Change grammatical error

david-mears-2:
The full sentence would be 'the task was run by X', not 'the task was ran by X', at least for british english

https://www.quora.com/Which-is-more-grammatically-correct-was-run-or-was-ran
https://english.stackexchange.com/questions/390341/using-have-ran-or-have-run
https://forum.wordreference.com/threads/ran-by-run-by.45092/


absternator (approved):
LGTM

PR #145, opened 2024-12-05
Rename explorer directory to home

david-mears-2:
We no longer have any Explorer component so this naming is obsolete.


absternator (approved):
love the cleanup

PR #148, opened 2024-12-18
Spike into using database 'views'

david-mears-2:
It would be nice to have some interim feedback on whether this exploration looks like its going in a good direction.

Database 'views' are like virtual tables made from pre-defined queries. The intent here is to tidy up some of the back-end logic by putting commonly used information about groups of packets into a single reusable piece of code. Since the intention of the view is to gather what's used for displaying information about packet groups, I've provisionally called it 'packet_group_display_view'.

This reduces (to zero) the amount of SQL needed in repositories, and it also simplifies away the need to ingest information about packet groups (beyond their name) in the packet ingestion phase, i.e., we don't need to store a group's latest display name, description etc. in the database. (We probably can’t get rid of the packet groups table entirely to replace it with ephemeral views, since we use its persistence to do role permissions. See [packit db diagram](https://dbdiagram.io/d/Packit-Auth-6626248503593b6b619c05e8).)

There is also in this branch a new column on the packet table, 'description', which is populated on ingestion so that it can be referred to by the view.

I hewed to [this Baeldung guide](https://www.baeldung.com/spring-data-jpa-repository-view) about 'Spring Data JPA Repository for Database View' so that our Kotlin backend can have a repository for the view in the way we're used to it doing for tables. Hence, there is a file ViewRepository that is a base class for repositories that deal with views rather than tables.

Views come in two types: ‘standard’ and ‘materialized’, where, as I understand it, the latter basically uses a (manually refreshed?) cache while the former runs the query every time it's referenced. We could move to materialized to save on queries - would require refreshing the cache periodically. See [here](https://www.baeldung.com/spring-data-jpa-repository-view) for more.


api/app/src/main/kotlin/packit/repository/PacketGroupDisplayRepository.kt
```diff
@@ -0,0 +1,11 @@
+package packit.repository
+
+import org.springframework.security.access.prepost.PostFilter
+import packit.model.PacketGroupDisplay
+
+interface PacketGroupDisplayRepository : ViewRepository<PacketGroupDisplay, String> {
```
david-mears-2, line 6: string should be int - left over from when i was using name as the identifier rather than packet group id.

EmmaLRussell, line 6: Spring works out from the name of the T type you pass it what view it needs to query from?

EmmaLRussell (commented):
This seems like a good approach to me - I definitely like keeping the sql separate from the kotlin code!

I think I still prefer PacketGroupSummary as a name tbh - it does form a summary, and whether the front end chooses to display it or stick it in an excel file isn't the concern of the api. But could live with either. 

We could use a materialised view, and always do a refresh when we load new packets from outpack server if we found any new packets, as we know that's the only time the view data will have changed. 

I was a bit surprised to see no Dto for PacketGroupDisplay. Is the idea that we'll only have a Dto if it differs from the Entity (Model) class, otherwise we'll just use that as our transfer class?

david-mears-2:
> I was a bit surprised to see no Dto for PacketGroupDisplay. Is the idea that we'll only have a Dto if it differs from the Entity (Model) class, otherwise we'll just use that as our transfer class?

It's more that I haven't wrapped my head around what a DTO is for.

api/app/src/main/kotlin/packit/repository/PacketGroupDisplayRepository.kt
david-mears-2: I think instead it reads the slightly-misnamed 'Table' annotation that lives on PacketGroupDisplay.

david-mears-2:
> I think I still prefer PacketGroupSummary as a name tbh - it does form a summary, and whether the front end chooses to display it or stick it in an excel file isn't the concern of the api. But could live with either.

Makes sense. Using 'display' as I change the code helps me keep track of what code is new vs old, but I can switch it all over to 'summary' once I've done all the changes.

david-mears-2:
Helpful links for trying out materialized views in spring boot:

1. General overview https://www.baeldung.com/sql/databases-views-simple-complex-materialized#materialized-view
2. How to add a refresh function to a repository https://ishansoninitj.medium.com/views-materialized-views-and-spring-data-jpa-05faeb8c0c67

plietar:
Honestly, unless we have factual evidence of SQL requests slowing us down (and I doubt it is), I don't think there is much benefit to using views, whether they are materialized or not.

My main concern with using views is that it makes the code more stateful. If today you want to change what information is computed / displayed by the API, that's a simple Kotlin change. This change is built and deployed pretty straight forwardly.

With views, if you want to make a change in the future, you now have to think about migrations, and instead of just writing code you have to think it terms of difference with the previous one and `ALTER VIEW`, etc... Deployment becomes more complicated, and if we ever for any reason need to rollback a deployed change we are in a world of pain. I can only imagine that materialized view will be even worse in that regard. Then if I want to understand what a view does, there isn't a single file I can look at and understand what is going on, I have to piece information across many migrations.

Sure all these issues exist with the main table schema, but storage schema (probably) change less often than logic, and the storage is not something we really have other options for, whereas for logic keeping it in Kotlin seems like a reasonable option.

api/app/src/main/resources/db/migration/V20241206__add_latest_display_name_column_to_packet_group.sql
plietar, line 1: You cannot change/delete any existing migration file. It will make the new version impossible to deploy. It happened a couple months ago and I had to manually fudge the database to get things to reconcile. If you need to undo an `ADD COLUMN` then you need a new migration to do this.

plietar: Ah I just realised this file isn't even in master yet. Yeah there's no problem deleting it from an yet to be merged branch then.

EmmaLRussell:
> Honestly, unless we have factual evidence of SQL requests slowing us down (and I doubt it is), I don't think there is much benefit to using views, whether they are materialized or not.
> 
> My main concern with using views is that it makes the code more stateful. If today you want to change what information is computed / displayed by the API, that's a simple Kotlin change. This change is built and deployed pretty straight forwardly.
> 

I suspect performance may become an issue once we get a decent number of packets in some of these instances, but the main problem we were trying to solve was code readability iirc. 

Good points on migration transparency  - the danger is that we just transfer unreadability to the sql files - and rollback (urgh). Maybe we could have a policy of not altering views but instead dropping and recreating them? But you'd still have the issue of logic fragmented across parts of the system. 

If we wanted to back out of this approach and instead try to make the SQL queries embedded in the Kotlin more readable I'm not sure there any options beyond string interpolation for repeated sub-queries or WITH clauses - but we could do that!

david-mears-2:
Thanks for the thoughtful comments @plietar .

> With views, if you want to make a change in the future, you now have to think about migrations, and instead of just writing code you have to think it terms of difference with the previous one and ALTER VIEW, etc... Deployment becomes more complicated, and if we ever for any reason need to rollback a deployed change we are in a world of pain. I can only imagine that materialized view will be even worse in that regard. Then if I want to understand what a view does, there isn't a single file I can look at and understand what is going on, I have to piece information across many migrations.

These are reasonable points. It suggests that instead of defining a view, we move the query it contains into the Kotlin back end, wrapped in some function. Then we re-use that function, enjoying the same benefits (such as re-usability and not having to ingest packet group information) as we would get from a view, while avoiding the risk of db migration headaches.

> Honestly, unless we have factual evidence of SQL requests slowing us down (and I doubt it is), I don't think there is much benefit to using views, whether they are materialized or not. [...] If today you want to change what information is computed / displayed by the API, that's a simple Kotlin change.

The intended benefit is about simplifying our work, rather than about making the app more performant. Having this pre-defined query (whether view or not) was an alternative to making ingestion complicated. To be specific: as we were introducing display names and descriptions at the packet-group level, ingestion was going to derive these from the packets, and store them in columns on the packet_groups. But this 'persistent columns' idea didn't work in the case of packet groups that already existed (those would need to be backfilled) (not so much a problem today but it would set the precedent for future such issues); it involved some duplication of data across tables; and it didn't feel totally future-proof because Rich had mentioned some features like parameterised display names that would mean that the semantics of the display_name column would change. That last one, I now realise, is an equally good argument against fossilising the current logic (namely, get properties from the latest packet) in a view.

I want to now investigate how we can do a reusable query without engraving it into the db.

EmmaLRussell:
> fossilising the current logic (namely, get properties from the latest packet) in a view.

With migrations, you should be able to avoid fossilising the logic itself, as that can change, so long as you supported the view's schema while the code was using it. But, as I say, you'd still have the logic fragmented across sql/kotlin.

EmmaLRussell:
> > With migrations, you should be able to avoid fossilising the logic itself, as that can change, so long as you supported the view's schema while the code was using it.
> 
> I'm not sure what this means. The logic that would be fossilised in the view would be the bit about relying on the latest packet for the info.

So if you still wanted to provide packet display rows, but wanted to change the logic used to actually generate them (e.g. not going to the last packet to get display name, but deriving display name in some other way), I think it would work to add a migration which dropped the view and then recreated it with the new logic. (It should be fine to do this with views, even materialised views, as they're derived from the actual data tables.)

As I mentioned in the previous comment (the one replying to Paul) the drop and recreate approach might be preferable when changing views in other ways (e.g. add a column) anyway, to avoid the logic fragmentation issue he mentions. 

(However I haven't ever tried this with migrations and views, so I'm not sure it would definitely work!)

david-mears-2:
> I want to now investigate how we can do a reusable query without engraving it into the db.

Here is an alternative proposal: https://github.com/mrc-ide/packit/pull/149

Let me know your thoughts @EmmaLRussell !

EmmaLRussell:
A second alternative, which avoids both the split logic of using a view, and the ickiness of big SQL strings in kotlin annotations, would be to use something like [jooq](http://www.jooq.org/). That examines your db query and generates classes which you can then use to write validated SQL-like queries (which are compiled to actual SQL) as kotlin code. 

One nice thing about that is that you can write shared subqueries once and re-use them e.g. return from a helper method and chain that result within the query that uses them. Here's an example of a nested query in jooq: https://www.jooq.org/doc/latest/manual/sql-building/table-expressions/derived-tables/

We've used jooq with Spark, but it works with Spring too: https://www.baeldung.com/jooq-with-spring

Here's an example of some of our jooq code in the wild: https://github.com/vimc/montagu-api/blob/c0a655d87955acc059c336c197deddbcdaf4a1f9/src/app/src/main/kotlin/org/vaccineimpact/api/app/repositories/jooq/JooqBurdenEstimateRepository.kt

The downside of jooq is that it can be a bit heavy - you need to remember to regenerate the classes, which get spat out as a pile of rather gruesome java. But I think the issue we're having here might demonstrate that we're starting to want to do sufficiently complex things with the database that might merit the advantages it can provide. 

Would suggest moving this discussion out of the motivating ticket, do the simple kotlin implementation for that, and discuss further in the new year.

PR #149, opened 2024-12-19
MRC-6040: Packet-group display names and descriptions

david-mears-2:
Currently a spike branch!

This builds off the long-standing branch https://github.com/mrc-ide/packit/pull/142 but is made in response to feedback to the db view spike branch https://github.com/mrc-ide/packit/pull/148.

Relative to https://github.com/mrc-ide/packit/pull/142, this adds:
* Nullable 'description' column to packet table, which is populated at ingestion time.
* Both queries over packet groups are now routed via a single SQL definition (in PacketGroupRepository) in order to minimize SQL usage in Kotlin (without having to use a db view).


api/app/src/main/resources/db/migration/V20241219__add_description_to_packet_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE packet
+    ADD COLUMN description TEXT;
```
david-mears-2, line 2: This raises the question of how to back-fill existing packets' descriptions.

david-mears-2: Something like this, in the PacketService?

```kt
fun backfillPacketDescriptions() {
    val packets = packetRepository.findAll()
    val updatedPackets = packets.map { packet ->
        val metadata = outpackServerClient.getMetadataById(packet.id)
        packet.description = getDescriptionForPacket(metadata)
        packet
    }
    packetRepository.saveAll(updatedPackets)
}
```
Then ssh into each production instance and invoke this function somehow, and then remove the function from the codebase

david-mears-2:
This solution doesn't quite work, because it uses ILIKE to match both 'searching for entities by a search filter' as well as for 'looking up an entity by its name', whereas the latter should use exact-matching.

api/app/src/main/kotlin/packit/service/PacketGroupService.kt
```diff
@@ -28,15 +29,9 @@ class BasePacketGroupService(
...
-        val packetGroup = packetGroupRepository.findByName(name)
-        if (latestPacketId == null || packetGroup == null) {
-            throw PackitException("doesNotExist", HttpStatus.NOT_FOUND)
-        }
-        val packetOrderlyMetadata = packetService.getMetadataBy(latestPacketId).custom?.get("orderly") as? Map<*, *>
-        val packetLongDescription = (packetOrderlyMetadata?.get("description") as? Map<*, *>)?.get("long") as? String
-        return PacketGroupDisplay(packetGroup.latestDisplayName, packetLongDescription)
+    override fun getPacketGroupDisplay(name: String): PacketGroupDisplayDto{
+        val packetGroupSummary = packetGroupRepository.getFilteredPacketGroupSummaries(name).firstOrNull()
+            ?: throw PackitException("No packets groups found for name $name", HttpStatus.NOT_FOUND)
```
EmmaLRussell, line 34: As discussed, I don't think this quite works, as `getFilteredPacketGroupSummaries` uses `ILIKE` so, if there was no exact match for `name`, it could return an inexact match instead, whereas we would want it to return a 404 in that case.

david-mears-2:
Not quite right!

PR #151, opened 2025-01-10
Downloads page (without zip downloads)

david-mears-2:
Note: As per components.md, the code of Card.tsx is not written by us but is copied from shadcn.

This updates the view of the downloads page, with some small behavioural changes too:
- artefacts (groups of output files) are displayed with their descriptions and member files
- orderly scripts are omitted from the downloads
- 'shared resource' files are noted with text
- directories are omitted from file names (to avoid cluttery repetition)
- depending on the file extension, different icons are shown, for quick discovery of the file you're looking for

You can check on all these features using the new `download-types` demo orderly packet group. 

I think there was meant to be a permissions distinction for the ability to see resources vs. artefacts (some people may only see artefacts). But I don't know how our permissions model works / which permission we should be checking for for that (the db shows these permissions: packet.read, packet.run, outpack.read, outpack.write, user.manage). Maybe that's something coming down the line and should be a separate ticket.

The feature of zip downloads is left for a separate ticket.


app/src/app/components/contents/downloads/Artefacts.tsx
```diff
@@ -0,0 +1,32 @@
+import { Card, CardContent, CardHeader } from "../../Base/Card";
+import FileRow from "./FileRow";
+import { Artefact, PacketMetadata } from "../../../../types";
+
+interface ArtefactProps {
+  artefacts: Artefact[];
+  packet: PacketMetadata;
+}
+
+export default function Artefacts({ artefacts, packet }: ArtefactProps) {
```
absternator, line 10: probs best to use named exports for consistency and discoverability. 

il leave this comment but applies to all files default export used

app/src/app/components/contents/downloads/Artefacts.tsx
```diff
@@ -0,0 +1,32 @@
+import { Card, CardContent, CardHeader } from "../../Base/Card";
+import FileRow from "./FileRow";
+import { Artefact, PacketMetadata } from "../../../../types";
+
+interface ArtefactProps {
```
absternator, line 5:
```suggestion
interface ArtefactsProps {
```

app/src/app/components/contents/downloads/Downloads.tsx
```diff
@@ -0,0 +1,67 @@
...
+import Artefacts from "./Artefacts";
+import OtherFiles from "./OtherFiles";
+import { Card, CardContent } from "../../Base/Card";
+import FileRow from "./FileRow";
+
+export default function Downloads() {
+  const { packetId, packetName } = useParams();
+  const { packet } = usePacketOutletContext();
+  const artefacts = packet?.custom?.orderly.artefacts;
+  const inputs = packet?.custom?.orderly.role
```
absternator, line 15: sorry what exactly is inputs? maybe a more descriptive name if it refers to something else but I'm unsure

app/src/app/components/contents/downloads/Downloads.tsx
```diff
@@ -0,0 +1,67 @@
...
+            <CardContent className="p-0">
+              <ul>
+                {packet.files.map((file, index) => (<li key={index}>
+                  <FileRow path={file.path} packet={packet} />
+                </li>))}
+              </ul>
+            </CardContent>
+          </Card>
+        </>) : (<>
+          {!!artefacts?.length && !!inputs?.length ? (
```
absternator, line 35: these nested ternary operators are hard to read and understand. Maybe best to extract into another component and use `If statements`

app/src/app/components/contents/downloads/FileRow.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const file = packet.files.filter((file) => {
+    return file.path === path.replace("//", "/");
+  })[0];
+  const fileName = path.split("/").pop();
+
+  return (file &&
+    <div className="p-2 flex justify-between">
+      <div className="flex items-center truncate">
+        <span className="min-w-fit">
+          {extension && (() => {
```
absternator, line 23: another component for this seems appropriate rather than the anonymous function

absternator (commented):
Looks awesome and works like a charm. Just some minor things below

- not using default exports and named ones.. better for discoverability and naming.. also more constant with rest of project
- may need to run `npm run format:write` to format files. Also id reccomned running prettier and formating files as you go.. 
-

app/src/app/components/contents/downloads/Downloads.tsx
david-mears-2: It's my 'friendly name' for 'role' which is an even more opaque name - see https://github.com/mrc-ide/orderly2/blob/48cb8fba7fc9cf46a6d2187931064bb83756e4c4/inst/schema/orderly/orderly.json#L43

We could call this `inputFiles` or `resources` or `otherFiles`

app/src/app/components/contents/downloads/Artefacts.tsx
david-mears-2: There are quite a few stragglers using `export default`, so I have separated this change to a new PR: https://github.com/mrc-ide/packit/pull/156

david-mears-2:
> may need to run npm run format:write to format files. Also id reccomned running prettier and formating files as you go..

I think prettier is 'running' in my IDE (IntelliJ), but doesn't work - the plugin claims to be toggled on and all the 'format on save' stuff is toggled on, but I have to trigger it manually for files to actually re-format. I'm going to have to do this in a separate PR as there are 800 lines of changes from `npm run format:write`.

Could I add lint-staged as described at https://prettier.io/docs/en/precommit.html ? The effect will be that it runs prettier (and anything else we want e.g. eslint) before any commit, only on the staged files.

app/src/app/components/contents/downloads/FileRow.tsx
david-mears-2: Or how about [this](https://github.com/mrc-ide/packit/pull/151/commits/f674f0213f31596a60b8a5c47a9dc80b3928b2b0)?

app/src/app/components/contents/downloads/Downloads.tsx
absternator: if ya feel ya drilling too deep, feel free to call the context again from the nested child component and not pass as props... thats what the context provides us; not needing to drill too deep.... Its accessing the same state so new "request" is being made

app/src/app/components/contents/downloads/FileRow.tsx
absternator: just saw and is alot nicer.. but `getFileIcon` is actually a jsx function but its being used like a regular function? so maybe best to just rename to `FileIcon` and extract out and use like `<Fileicon extension={extension}/>

absternator:
> > may need to run npm run format:write to format files. Also id reccomned running prettier and formating files as you go..
> 
> I think prettier is 'running' in my IDE (IntelliJ), but doesn't work - the plugin claims to be toggled on and all the 'format on save' stuff is toggled on, but I have to trigger it manually for files to actually re-format. It's not an uncommon problem based on Googling... but I can't fix it. I'm going to have to do this in a separate PR as there are 800 lines of changes from `npm run format:write`.
> 
> Could I add lint-staged as described at https://prettier.io/docs/en/precommit.html ? The effect will be that it runs prettier (and anything else we want e.g. eslint) before any commit, only on the staged files. I used in daedalus to enforce code styles.

ohh must be annoying... I'm not a big fan of precommit hooks because its easy to pass by them (I've done this a lot before ahha).... maybe in the CI you can just run format:check and the pipeline will fail ... `- name: format check
        run: npm run format:check`

absternator (commented):
Nice work LGTM.. just minor comments.... 

not approving yet as pipelines failing.. ping me and il approve when the pass

app/src/app/components/contents/downloads/Downloads.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { packet } = usePacketOutletContext();
+  const packetIsFromOrderly = !!packet?.custom?.orderly;
+
+  return (
+    <>
+      {packet && (
+        <>
+          <PacketHeader packetName={packetName ?? ""} packetId={packetId ?? ""} displayName={packet?.displayName} />
+          {packetIsFromOrderly && <OrderlyDownloads packet={packet} />}
+          {!packetIsFromOrderly && (
```
EmmaLRussell, line 19: We don't have any non-orderly packets in our demo set, do we? Maybe we should make a ticket to add one so we can easily manually test code like this.

app/src/app/components/contents/downloads/Artefacts.tsx
```diff
@@ -0,0 +1,33 @@
+import { Card, CardContent, CardHeader } from "../../Base/Card";
+import FileRow from "./FileRow";
+import { Artefact, PacketMetadata } from "../../../../types";
+
+interface ArtefactsProps {
+  artefacts: Artefact[];
+  packet: PacketMetadata;
+}
+
+export default function Artefacts({ artefacts, packet }: ArtefactsProps) {
```
EmmaLRussell, line 10: If Artefacts and OtherFiles are orderly specific, maybe they should be named "OrderlyArtefacts" etc, or maybe we should have an orderly subfolder for orderly specific stuff. I've a feeling this might get a bit tangled between orderly and non-orderly implementations and would be good to keep them cleanly separated.

app/src/app/components/contents/downloads/FileRow.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const presentationExtensions = ["pdf", "html", "ppt", "pptm", "pptx", "potx", "potm", "pps", "xps"];
+const tableExtensions = ["csv", "xls", "xlsx", "xlsm", "xltx", "ods"];
+const imageExtensions = ["jpeg", "jpg", "png", "jiff", "bmp", "gif"];
+const defaultFileIcon = <File className="text-gray-400" />;
+
+const getFileIcon = (extension: string | undefined) => {
+  if (!extension) return defaultFileIcon;
+
+  if (presentationExtensions.includes(extension)) return <Presentation />;
```
EmmaLRussell, line 20: Inevitably someone's going to have upper case file extensions, so probably should make this case insensitive!

app/src/types.ts
```diff
@@ -100,11 +102,28 @@ interface Description {
...
+  Resource = "resource",
+  Dependency = "dependency",
+  Shared = "shared",
+}
+
+export interface SharedResource {
+  here: string;
+  there: string;
+}
+
```
EmmaLRussell, line 126: "Role" is a really confusing type here, given that we have permission-related roles as well. It's more verbose, but could we call these something more precise like "OrderlyInputRole"..?

EmmaLRussell (commented):
Looks great! I really like the file type icons.. 

What was the thinking behind excluding orderly scripts? I'd have thought they'd be useful for some users - they'll have other means of viewing the scripts (through github) typically, but could be convenient to be able to access the script if you're viewing the packet anyway. 

..particularly if we're going to include resource view permission for some users as a separate permission. I vaguely remember that discussion! Yes, we'd need to add that to our current resource model, and we should implement it in the back end. Let's discuss that at next web dev meeting. 

> Anmol:
> maybe in the CI you can just run format:check and the pipeline will fail ...
:+1:

app/src/types.ts
david-mears-2: Good point

app/src/app/components/contents/downloads/Downloads.tsx
david-mears-2: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6157/Add-non-orderly-packet-to-demos

app/src/app/components/contents/downloads/Downloads.tsx
david-mears-2: OK great, thanks

david-mears-2:
Ticket for changing permissions model: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6158/Add-resource-view-permission

david-mears-2:
>  maybe in the CI you can just run format:check and the pipeline will fail

So what I'll do is, after this PR is merged, I'll run the formatter over the whole codebase, and add this check to the pipeline.

absternator (approved):

EmmaLRussell (approved):
looks great, thanks!

PR #159, opened 2025-01-15
Update to new orderly.runner interface.

plietar:
The new runner does not share an on-disk repository with outpack server anymore. Instead it uses it supports an arbitrary number of repositories, and each request needs to provide the repository URL.

For now, a most one Git repository can configured in the application properties. In the future, we may expand this to multiple repositories, or even allow them to be configured from the frontend and stored in the database.

When submitting a run request, we also need to provide details of the outpack location. We use the same URL to outpack_server that Packit uses itself. For now at least, Packit, outpack_server and the runner workers all run on the same machine. In the future, we may make it so that the runner workers communicate via Packit, and add an authentication mechanism.


scripts/git-daemon.dockerfile
```diff
@@ -0,0 +1,14 @@
+# This sets up a container image that serves an Orderly source repository as a
+# Git repository, for use as a remote repository the runner can fetch.
+# The repository is exposed over the Git protocol.
```
M-Kusumgar, line 3: just to be clear about the port
```suggestion
# The repository is exposed over the Git protocol (port 9418).
```

app/src/app/components/contents/runner/run/PacketRunPacketGroupFields.tsx
```diff
@@ -9,10 +9,10 @@ import { PacketRunParamsField } from "./PacketRunParamsField";
 
 interface PacketRunPacketGroupFieldsProps {
   form: UseFormReturn<z.infer<typeof packetRunFormSchema>>;
-  branchCommit: string;
+  branchCommit: string | undefined;
 }
 export const PacketRunPacketGroupFields = ({ form, branchCommit }: PacketRunPacketGroupFieldsProps) => {
-  const { packetGroups, isLoading, error } = useGetRunnerPacketGroups(branchCommit);
+  const { packetGroups, isLoading, error } = useGetRunnerPacketGroups(branchCommit ?? null);
```
M-Kusumgar, line 15: instead of converting from `undefined` to `null` why not just feed it in as `undefined`?
```suggestion
  const { packetGroups, isLoading, error } = useGetRunnerPacketGroups(branchCommit);
```
and the `use` hooks can accept undefined and still work the same

M-Kusumgar (commented):
Only got a couple of small comments but after those should be good!

app/src/app/components/contents/runner/run/PacketRunPacketGroupFields.tsx
plietar: Yeah good point thanks.

M-Kusumgar:
ps: do you need to add location to the submit thing to get the tests passing?

plietar:
> ps: do you need to add location to the submit thing to get the tests passing?

I'm not sure, it is working locally ...

There's an error happening in the runner, and unfortunately those errors don't actually get logged: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6152
I'm making a quick branch of the runner just so I can see them: https://github.com/mrc-ide/orderly.runner/pull/18

PR #160, opened 2025-01-15
Make breadcrumb behave responsively

david-mears-2:
## Down with this sort of thing

Screenshot from 2025-01-15 20-51-28
Screenshot from 2025-01-15 20-51-49

## Up with this sort of thing

Screenshot from 2025-01-15 20-50-46
image


absternator (approved):
like the fix!!! lgtm

PR #161, opened 2025-01-16
Zip downloads

david-mears-2:
Adds buttons that download groups of files as a zip. The groups are different subsets of the files associated with some packet.

### Back end stuff:

In order to avoid having to write to a zip file and then stream that (to avoid loading the whole file in memory), in the new endpoint we write the zip file as it's being created directly into the response output stream, copying from the outpack_server response input stream into the zip output stream. When streaming, it's [recommended](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/mvc/method/annotation/StreamingResponseBody.html) to explicitly configure the 'task executor', which [by default](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/task/SimpleAsyncTaskExecutor.html) does not use a thread pool but just spins up a new thread for every task. If we create threads uncontrollably, we may run out of operating system-level threads quickly. Also, for operations bottlenecked on I/O, having a large number of threads just takes up RAM and adds more context-switching overhead, rather than increasing through-put.

#### How to manually test the back end

You'll want to verify that the server does not run out of memory when transferring large files. To do this:

0. Create a new orderly packet called e.g. massive-file following the instructions in demos/README.md. Add the following R code in the orderly script to create an 870MB file, running the script as per demos/README.md:

```R
write.csv(matrix(rnorm(5e7), ncol = 10), "massive.csv", row.names = FALSE)
orderly2::orderly_artefact("Massive file", "massive.csv")
```

1. Run scripts/run-dependencies and wait for it to complete. At this point, the large file will have been stored in the outpack server, ready for us to try to download it.
1. Create a build of the packit backend using ./api/scripts/build, and (crucially) run this build using a reduced heap size, e.g. 128MB, by running this command from the ./api directory: `java -Xmx128m -jar app/build/libs/app.jar`
1. Run the front-end (if you want to test the full stack) from the /app directory as normal (npm start). Navigate to massive-files packet download page and click 'Download all files'.
1. You can also test if the main thread gets blocked by the download by trying to make requests while the download is in progress, such as by making other download requests. (With reduced memory, you'll find that multi-threading allows zip downloads to be carried out simultaneously, but singular-file downloads will fail if the file size is large.)
1. Clean up your downloads folder! These files are big.

The streaming code was originally adapted from this blog post: https://woroniecki.pl/efficient-rest-endpoint-in-spring-boot-for-streaming-multiple-files-to-an-instant-zip-download/

### Front end stuff:

The following buttons are now available:

* Download all files associated with a packet
* Download all artefacts associated with a packet
* Download all files associated with an artefact
* Download all non-artefact files associated with a packet


api/app/src/main/kotlin/packit/AppConfig.kt
```diff
@@ -1,21 +1,24 @@
...
 import org.springframework.core.env.Environment
 import org.springframework.core.env.get
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder
 import org.springframework.security.crypto.password.PasswordEncoder
 import org.springframework.stereotype.Component
+import packit.config.TaskExecutorProperties
 
 @Component
-class AppConfig(private val enviroment: Environment)
+@EnableConfigurationProperties(TaskExecutorProperties::class)
```
plietar, line 13: I don't think you need `EnableConfigurationProperties`. I've used `ConfigurationProperties` in other places fine without it (see `ServiceLoginConfig`). There's a `@ConfigurationPropertiesScan` annotation on `packit.App` that automatically loads any `ConfigurationProperties` found in the `packit` package.

plietar, line 13: I don't think you need this line. I've used `ConfigurationProperties` in other places fine without it (see `ServiceLoginConfig`). There's a `@ConfigurationPropertiesScan` annotation on `packit.App` that automatically loads any `ConfigurationProperties` found in the `packit` package.

api/app/src/main/kotlin/packit/config/TaskExecutorProperties.kt
```diff
@@ -0,0 +1,56 @@
...
+*
+* References/docs:
+* 1) https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/mvc/method/annotation/StreamingResponseBody.html
+* 2) https://docs.spring.io/spring-framework/docs/4.3.12.RELEASE/spring-framework-reference/html/mvc.html#mvc-ann-async-output-stream
+* 3) https://docs.spring.io/spring-framework/docs/4.3.12.RELEASE/spring-framework-reference/html/mvc.html#mvc-ann-async-configuration-spring-mvc
+* 4) https://www.baeldung.com/thread-pool-java-and-guava
+* 5) https://docs.spring.io/spring-framework/reference/integration/scheduling.html#scheduling-task-scheduler
+* 6) https://codingtim.github.io/spring-threadpooltaskexecutor/
+*/
+@Configuration
```
plietar, line 32: Don't think this is needed

plietar, line 32: Don't think this is needed

plietar, line 32: Don't think this is needed, `@ConfigurationProperties` is enough.

From what I understand of Spring Boot, `@Configuration` is used on classes that have `@Bean` methods, which this class does not.

plietar, line 32: Don't think this is needed, `@ConfigurationProperties` is enough.

From what I understand of Spring Boot, `@Configuration` is used on classes that have `@Bean` methods, which this class does not.

api/app/src/main/kotlin/packit/config/TaskExecutorProperties.kt
```diff
@@ -0,0 +1,56 @@
...
+* machine, and aim to minimize the risk of overloading the machine, rather than to optimize response time.
+* If you know the properties of your hosting machine and context (how many threads can it handle concurrently? how
+* significant are the bottlenecks of disk-reading I/O and network-writing I/O? how many users do you need to handle at
+* once?), you can adjust these values to better fit your needs by setting the environment variables read in by
+* application.properties.
+*
+* If the machine can theoretically handle a large number of threads, maximizing this won't necessarily improve
+* throughput, since file-streaming is an I/O-bound operation, which means that threads will spend most of their time
+* waiting for their turn to use the I/O (taking up about 1MB of RAM each, in the case of a 64-bit JVM), rather than
+* doing any useful work with the CPU.
```
plietar, line 22: I would remove this paragraph. File streaming alone may be I/O bound, but the tasks we schedule on the thread pool might do more than just streaming. In this instance for example, the worker threads are doing Zip compression at the same time as streaming, which is probably going to be CPU bound.

plietar, line 22: I would remove this paragraph. File streaming alone may be I/O bound, but the tasks we schedule on the thread pool might do more than just streaming. In this instance for example, the worker threads are doing Zip compression at the same time as streaming, which is probably going to be CPU bound.

plietar, line 22: I would remove this paragraph. File streaming alone may be I/O bound, but the tasks we schedule on the thread pool might do more than just streaming. In this instance for example, the worker threads are doing Zip compression at the same time as streaming, which is probably going to be CPU bound.

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -77,4 +81,29 @@ class PacketController(
...
+    @PreAuthorize("@authz.canReadPacket(#root, #id, #name)")
+    // TODO: Verify authorization for accessing specific files: whether user has access to all files, or artefacts only.
+    fun downloadZip(
+        @PathVariable id: String,
+        @PathVariable name: String,
+        @RequestParam(required = true) hashes: List<String>,
+        @RequestParam(required = true) filenames: List<String>,
+    ): ResponseEntity<StreamingResponseBody>
+    {
+        val filesWithNames = hashes.mapIndexed { index, hash ->
```
plietar, line 95: I would use a [`zip`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/zip.html) (unfortunate naming collision) here rather than indexing, allowing you to iterate over two lists at the same time.

Something like:
```kotlin
val filesWithNames = hashes.zip(filenames) { hash, filename ->
    val fileResource = packetService.getFileByHash(hash, false, hash).first
    fileResource to filename
}
```

Alternatively make this a POST and put a JSON "filename -> hash" map in the request body.

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -77,4 +81,29 @@ class PacketController(
...
+            val fileResource = packetService.getFileByHash(hash, false, hash).first
```
plietar, line 96: The call to `getFileByHash` is going to return the file contents as a `ByteArray`.

The `filesWithNames` variable is going to end up holding the entire packet in memory, which defeats all the work done to stream the output of the Zip.

What you need to do is extend the `PacketService` class to have a streaming version of `getFileByHash`. You need to move the calls to that function inside the `StreamingResponseBody` to make sure they happen on the thread pool, and stream the file bodies into the Zip without loading the whole contents in memory.

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -77,4 +81,29 @@ class PacketController(
             .headers(response.second)
             .body(response.first)
     }
+
+    @GetMapping("/{name}/{id}/zip")
```
plietar, line 85: Do we really need the name here? Can't it be derived from the `id`?

Also I think it would be best to have the format as `/zip/{id}` to avoid name collisions (and change `getPacketsByName`).

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -77,4 +81,29 @@ class PacketController(
...
+    @PreAuthorize("@authz.canReadPacket(#root, #id, #name)")
+    // TODO: Verify authorization for accessing specific files: whether user has access to all files, or artefacts only.
```
plietar, line 87: Most importantly, you need to check that the hashes that have been specified really belong to the packet in question. The current implementation does access checks on `id` and `name`, but then doesn't really use these values.

Nothing prevents a user from specifying a packet ID that they do have access to, but with a list of hashes from a different packet for which they do not have access. The API server will happily return those files.

This is actually a bug in `findFile` and will need to be fixed there as well (I also don't understand why `findFile` uses `canReadPacketMetadata` and not `canReadPacket`).

plietar (changes requested):

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -77,4 +81,29 @@ class PacketController(
...
+    fun downloadZip(
+        @PathVariable id: String,
+        @PathVariable name: String,
+        @RequestParam(required = true) hashes: List<String>,
+        @RequestParam(required = true) filenames: List<String>,
```
EmmaLRussell, line 92: These are hashes and filenames which should be in sync e.g. hashes[0] is hash for filenames[0]? Would be better to send that in a more structured way to avoid length mismatch e.g. a map of filenames to hashes. 

Do you actually need the filename, or could that be discoverable from the hash? (Maybe not - I suppose in theory you could have two files with different names and same content/hash in a packet..?)

Need to return 400 if invalid hash for packet (or filename not matching packet metadata..?)
(..or {id} not in {name})

api/app/src/main/kotlin/packit/controllers/PacketController.kt
EmmaLRussell, line 96: Does this read in the whole file? Probably want to deal with streaming across each file in turn in the zip service, some packet files are massive!.

EmmaLRussell (commented):
Sorry, github's been weird, might get duplicate coments :scream: :

api/app/src/main/kotlin/packit/controllers/PacketController.kt
plietar: I've fixed the `findFile` bug in https://github.com/mrc-ide/packit/pull/171

> I also don't understand why findFile uses canReadPacketMetadata and not canReadPacket

It turned out `canReadPacket` and `canReadPacketMetadata` basically do the same thing, but the existing `canReadPacket` was very error prone (in your code, the user controls `name` and can use it to access any ID if they have permission for at least one packet). I've tried to clarify the interface. For this endpoint, after my PR is merged, you should be able to use:
```kotlin
@PreAuthorize("@authz.canReadPacket(#root, #id")` 
```

api/app/src/main/resources/application.properties
```diff
@@ -18,6 +18,11 @@ management.endpoints.web.exposure.include=health,prometheus
 spring.mvc.throw-exception-if-no-handler-found=true
 spring.web.resources.add-mappings=false
 server.error.include-stacktrace=never
+#task execution
+spring.task.execution.pool.core-size=${PACKIT_TASK_EXECUTOR_CORE_POOL_SIZE:4}
+spring.task.execution.pool.max-size=${PACKIT_TASK_EXECUTOR_MAX_POOL_SIZE:10}
+spring.task.execution.pool.queue-capacity=${PACKIT_TASK_EXECUTOR_QUEUE_CAPACITY:100}
+spring.task.execution.pool.keep-alive=${PACKIT_TASK_EXECUTOR_KEEP_ALIVE_SECONDS:60}
```
plietar, line 25:
```suggestion
spring.task.execution.pool.core-size=4
spring.task.execution.pool.max-size=10
spring.task.execution.pool.queue-capacity=100
spring.task.execution.pool.keep-alive=60
```

I understand you're just following the existing pattern in this file, but there's really no need for us to define any of these environment variables. Spring Boot supports "externalized configuration", which means it will read properties from all kinds of sources and merge them before loading the application. This includes environment variables already. These values can be overridden by setting `SPRING_TASK_EXECUTION_POOL_CORESIZE` etc..

https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.typesafe-configuration-properties.relaxed-binding.environment-variables

api/app/src/main/kotlin/packit/controllers/PacketController.kt
plietar: I’ve thought some more about it and maybe `/{id}/zip` is alright actually and consistent with the `file` and `metadata` one, as long as we shuffle the overlapping routes around eventually.

plietar: I've started re-arranging these routes to make a bit more sense in https://github.com/mrc-ide/packit/pull/173

david-mears-2: Looks like we're moving to `/packets/<id>/file` for file downloads, and so I'll go for `/packets/<id>/zip`.

api/app/src/main/kotlin/packit/controllers/PacketController.kt
david-mears-2: > Do you actually need the filename, or could that be discoverable from the hash?

I believe we could look up the filename from the packet's metadata using the hash, but this would imply one extra request to retrieve the necessary metadata. I'd rather complicate the request param (as a map of hashes and filenames) than add another request in.

api/app/src/main/kotlin/packit/service/OutpackServerClient.kt
```diff
@@ -75,6 +79,24 @@ class OutpackServerClient(appConfig: AppConfig) : OutpackServer
         return GenericClient.get(constructUrl(url))
     }
 
+    override fun addFileToZip(
+        filename: String,
+        hash: String,
+        zipOutputStream: ZipOutputStream,
+        request: HttpServletRequest,
+        response: HttpServletResponse
```
plietar, line 86: You shouldn't need request and response at all here. Just the output stream is enough. I think this points at the GenericClient interface being off.

Also from a separation of concerns perspective, I don't think `OutpackServerClient` should have any Zip related code. Th interface should be the same as the existing `getFileByHash` method, but with a streaming output instead of a `ByteArray`. All the Zip manipulation code should be contained in `PacketService`.

I would expect the signature to look like:
```kt
override fun getFileByHash(hash: String, output: OutputStream)
```

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -31,7 +34,7 @@ interface PacketService
     fun getPacketsByName(
         name: String
     ): List<Packet>
-
+    fun streamZip(hashes: List<String>, id: String, request: HttpServletRequest, response: HttpServletResponse)
```
plietar, line 37:
```suggestion
    fun streamZip(hashes: List<String>, id: String, output: OutputStream)
```

This is all that this function should need as its arguments

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -82,28 +83,21 @@ class PacketController(
...
-        @RequestParam(required = true) hashes: List<String>,
-        @RequestParam(required = true) filenames: List<String>,
-    ): ResponseEntity<StreamingResponseBody>
-    {
-        val filesWithNames = hashes.mapIndexed { index, hash ->
-            val fileResource = packetService.getFileByHash(hash, false, hash).first
-            fileResource to filenames[index]
-        }
+        @RequestParam hashes: List<String>,
+        request: HttpServletRequest,
```
plietar, line 92:
```suggestion
```

This is a GET request with no body associated with it. There's not really any reason why you should want access to the request stream.

plietar (changes requested):

api/app/src/main/kotlin/packit/service/OutpackServerClient.kt
```diff
@@ -75,6 +79,24 @@ class OutpackServerClient(appConfig: AppConfig) : OutpackServer
...
+    ) {
+        val url = constructUrl("/file/$hash")
+        GenericClient.proxyRequest(url, request, response, false) { serverResponse ->
+            serverResponse.body.use { inputStream ->
+                zipOutputStream.putNextEntry(ZipEntry(filename))
+                streamInputToOutput(inputStream, zipOutputStream)
```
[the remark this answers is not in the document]
david-mears-2: You seem to be commenting on an outdated version of the code, maybe using an old browser tab. I deleted the function `streamInputToOutput` yesterday when I realised IOUtils.copy did the same thing.

api/app/src/main/kotlin/packit/service/GenericClient.kt
```diff
@@ -57,7 +58,8 @@ object GenericClient
         url: String,
         request: HttpServletRequest,
         response: HttpServletResponse,
-        copyRequestBody: Boolean
+        copyRequestBody: Boolean,
+        responseExtractor: ((ClientHttpResponse) -> Boolean)? = null
     )
```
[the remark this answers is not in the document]
david-mears-2: (This comment is on yesterday's version of this function, but still validly applies to the latest version)

api/app/build.gradle.kts
```diff
@@ -52,6 +52,7 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-data-jpa")
     implementation("org.springframework.boot:spring-boot-starter-validation")
     implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("org.springframework.boot:spring-boot-starter-webflux")
```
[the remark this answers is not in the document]
david-mears-2: (Yep, did so before asking for re-review!)

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -82,28 +83,21 @@ class PacketController(
...
-        @PathVariable name: String,
-        @RequestParam(required = true) hashes: List<String>,
-        @RequestParam(required = true) filenames: List<String>,
-    ): ResponseEntity<StreamingResponseBody>
-    {
-        val filesWithNames = hashes.mapIndexed { index, hash ->
-            val fileResource = packetService.getFileByHash(hash, false, hash).first
-            fileResource to filenames[index]
-        }
+        @RequestParam hashes: List<String>,
```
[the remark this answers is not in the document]
plietar: > (The [hash function appears](https://github.com/mrc-ide/orderly2/blob/48cb8fba7fc9cf46a6d2187931064bb83756e4c4/R/outpack_hash.R#L45) to hash only the file contents rather than its name, so we're talking about any time somebody has files with identical content.)

Yes, this is very much by design. It helps optimize storage by only keeping a single copy on the server.
This is a pretty standard approach called "content addressable storage" used by Git and many other systems.

plietar:
I had a quick look through the streaming code and I think it looks good now!

EmmaLRussell (commented):
LGTM, really tidy. Tiny question unrelated to the streaming stuff,

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
```diff
@@ -197,4 +204,45 @@ class PacketControllerTest : IntegrationTest()
...
+        assertEquals(result.statusCode, HttpStatus.OK)
+        assertEquals(result.headers["Transfer-Encoding"]?.firstOrNull(), "chunked") // Header denoting streaming
+        assertEquals(result.headers.contentType.toString(), "application/zip")
+
+        // Read the stream into a zip file
+        val zipInputStream = ZipInputStream(result.body!!.inputStream())
+        val entries = mutableListOf<String>()
+        var entry: ZipEntry? = zipInputStream.nextEntry!!
+        while (entry != null) {
+            println("entry.name: " + entry.name)
```
EmmaLRussell, line 239: Was this just for debugging? Can remove now?

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
```diff
@@ -197,4 +204,45 @@ class PacketControllerTest : IntegrationTest()
...
+            entries.add(entry.name)
```
EmmaLRussell, line 240: Might be nice to check sizes here as well.

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
```diff
@@ -197,4 +204,45 @@ class PacketControllerTest : IntegrationTest()
             assertThat(it.get("name").textValue()).isEqualTo("artefact-types")
         }
     }
+
+    @Test
+    @WithAuthenticatedUser(authorities = ["packet.read"])
```
EmmaLRussell, line 209: Could scope this to the packet group as the other integration tests do.

api/app/src/test/kotlin/packit/integration/services/OutpackServerClientTest.kt
```diff
@@ -24,4 +25,16 @@ class OutpackServerClientTest : IntegrationTest()
         val result = sut.getMetadata()
         assert(result.map { it.name }.containsAll(listOf("parameters", "explicit", "depends", "computed-resource")))
     }
+
+    @Test
+    fun `can write the input stream from a get response to an output stream`()
+    {
+        val hash = "sha256:1a1b649d911106d45dcb58e535aae97904f465e66b11a38d7a70828b53e3a2eb"
```
EmmaLRussell, line 32: Might be nice to include a comment here saying what file from what packet you're expecting this to be!

api/app/src/test/kotlin/packit/unit/controllers/PacketControllerTest.kt
```diff
@@ -142,4 +145,17 @@ class PacketControllerTest
         }
         assertEquals("doesNotExist", error.key)
     }
+
+    @Test
+    fun `streamZip should set correct response headers and content type`() {
```
EmmaLRussell, line 150: Could also check that it calls the service as expected.

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -117,6 +120,29 @@ class BasePacketService(
         return this.joinToString("") { "%02x".format(it) }
     }
 
+    override fun streamZip(paths: List<String>, id: String, output: OutputStream) {
+        val hashesByPath = getMetadataBy(id).files
+            .filter { it.path in paths }
+            .associateBy({ it.path }, { it.hash })
+
+        if (hashesByPath.size != paths.size) {
```
EmmaLRussell, line 128: Could throw a slightly friendlier error here including the filenames which weren't found - would just need to filter to the paths which weren't in metadata files mapped to their names.

app/src/app/components/contents/downloads/DownloadButton.tsx
```diff
@@ -11,21 +11,31 @@ interface DownloadButtonProps {
 
 export const DownloadButton = ({ file, packetId }: DownloadButtonProps) => {
   const [error, setError] = useState("");
+  const [downloading, setDownloading] = useState(false);
```
EmmaLRussell, line 14: Since there's now also FileGroupDownloadButton, maybe this one should now be called "FileDownloadButton" or similar

app/src/app/components/contents/downloads/FileGroupDownloadButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+interface FileGroupDownloadButtonProps {
+  files: FileMetadata[];
+  zipName: string;
+  buttonText?: string;
+  variant?: VariantProps<typeof buttonVariants>["variant"];
+  className?: string;
+}
+
+export const FileGroupDownloadButton = ({
```
EmmaLRussell, line 19: Maybe call this ZipDownloadButton or similar?

app/src/app/components/contents/downloads/FileGroupDownloadButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const [error, setError] = useState("");
+  const [downloading, setDownloading] = useState(false);
+
+  const downloadFiles = async () => {
+    if (!packet || !packetId || files.length < 1) {
+      return;
+    }
+    setDownloading(true);
+    await download(getZipUrl(packetId, files), zipName)
+      .then(() => setError(""))
```
EmmaLRussell, line 37: Maybe error should be reset when you start a download rather than when it successfully completes. If you've had an error then start a new attempt which takes a while, it would be confusing to have the old error from the previous attempt hanging around.

EmmaLRussell (commented):
Great stuff. The front end looks really good. 

The instructions to create a large-file demo  packet seem generally really useful, and would be good to capture in README or KB articles, and might even be good to automate optionally in the bootstrapping scripts (in another ticket!). 

I did find that if I had a big download going (all files) and requested a download of massive.csv before the zip download had completed, then the csv downloaded as an empty (0 bytes) file - using Firefox. 

A couple of bits of behaviour that surprised me:
- multi-file artefacts get their own folder but single-file artefacts don't. I guess that's to cut down on the number of folders, but if you do have some meaning embedded in the artefact name, that will be lost...
- ..and currently it's lost anyway, even for the multi-file artefacts because the folders get called e.g. "artefact1" rather than getting the artefact name, which I think would be useful. 

We really must put some e2e tests into packit....

api/app/src/main/resources/application.properties
```diff
@@ -18,6 +18,11 @@ management.endpoints.web.exposure.include=health,prometheus
 spring.mvc.throw-exception-if-no-handler-found=true
 spring.web.resources.add-mappings=false
 server.error.include-stacktrace=never
+#task execution
+spring.task.execution.pool.core-size=4
+spring.task.execution.pool.max-size=10
+spring.task.execution.pool.queue-capacity=100
+spring.task.execution.pool.keep-alive=60s
```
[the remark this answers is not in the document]
david-mears-2: This is really helpful - I didn't realise Tomcat was managing a different thread pool.

david-mears-2:
> I did find that if I had a big download going (all files) and requested a download of massive.csv before the zip download had completed, then the csv downloaded as an empty (0 bytes) file - using Firefox.

This is a pre-existing bug / lack-of-feature, that is addressed by https://github.com/mrc-ide/packit/pull/176 and mentioned in this PR description!

david-mears-2:
> The instructions to create a large-file demo packet seem generally really useful, and would be good to capture in README or KB articles, and might even be good to automate optionally in the bootstrapping scripts (in another ticket!).

I've now added instructions to the demos/README.md.

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
david-mears-2: Unfortunately, the getters 'size' and 'compressedSize' are both returning -1, which denotes an unknown size.

app/src/app/components/contents/downloads/DownloadButton.tsx
david-mears-2: Will do in https://github.com/mrc-ide/packit/pull/176

EmmaLRussell:
> Actually, what you are seeing @EmmaLRussell is an artefact (no pun intended) of what paths the files in the download-types packet have, which (unusually for our demo packets) have directories such as `artefact1/` within their paths (in the case of the multi-file artefact only) - see below snippet. When wrapping the files up into a zip, we faithfully use the files' paths as-is, and do not do any additional folderizing. But we could! Then download-types' zip would have folders as follows (note that `input_files` is also a user-generated folder here, hence its being wrapped by `resources`):

Ah, yup, completely misinterpreted that despite having read the code! OK, yeah, I think probably users will expect orderly structure to be maintained.

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
plietar: Amazingly, this works:
```kt
val nextEntry = zipInputStream.nextEntry
println(entry.size)
println(entry.compressedSize)
entry = nextEntry
```

found here https://stackoverflow.com/questions/25602406/zipentry-unknown-size-although-set

plietar (approved):
Backend bit LGTM!

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -121,12 +121,18 @@ class BasePacketService(
...
+        val notFoundPaths = paths.filter { path -> metadataFiles.none { it.path == path } }
 
-        if (hashesByPath.size != paths.size) {
-            throw PackitException("notAllFilesFound", HttpStatus.NOT_FOUND)
+        if (notFoundPaths.isNotEmpty()) {
+            throw PackitException("notAllFilesFound: ${notFoundPaths.joinToString(", ")}", HttpStatus.NOT_FOUND)
+        }
+
+        if (paths.isEmpty()) {
+            throw PackitException("noFilesProvided", HttpStatus.BAD_REQUEST)
```
EmmaLRussell, line 135: Sorry, I should have spotted before that "notAllFilesFound", and now "noFilesProvided" need to be added to the resource bundle (`src/main/resources/errorBundle.properties`) so they can be mapped to friendlier error messages (which in theory may eventually include non-English) translations.  However, we don't currently support error messages in the bundle which can be interpolated as you're doing here with the file names, but that would be fairly easy to add in - would suggest doing that in another ticket and just using a friendly error messages here for `notAllFilesFound` for now (if the string passed to `PackitException` isn't a resource key, it just uses it as the error message).

EmmaLRussell (approved):
Looks great, thank you!

I've realised there's a little bit of sorting out to do around interpolated error message templates, but I don't think that needs to happen in this branch, will add a ticket to try to explain what I mean. 

You might want to try out Paul's suggestion for getting compressedSize in the integration test.?

api/app/src/main/kotlin/packit/service/PacketService.kt
EmmaLRussell: Added a ticket for this: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6221/Interpolated-error-messages

PR #164, opened 2025-01-16
 Add a formatting check to the front-end-test CI workflow

david-mears-2:
Introduced a format check into the front-end test GH. Saw it fail successfully on CI. Ran the formatter on the codebase to bring files up to date with prettier's preferences. Bon appetit.


absternator (approved):
Thanks for this!!!

PR #168, opened 2025-01-22
Mrc-6166 Add filter + sort to packet table

absternator:
The following adds sorting and filtering to packet table. The following is done in PR:

- update /packets/{name} so it fetches all data and allows us to do client side pagination, sorting and filtering
- add sorting & filtering to packet column
- add filtering to parameters column (allow filtering by param name or param value)

Testing:
on packet table page ensure sort and filter works

image

### DIFF
--- a/app/src/app/components/contents/PacketGroup/hooks/useGetPacketPage.ts
+++ b/app/src/app/components/contents/PacketGroup/hooks/useGetPacketPage.ts
@@ -1,17 +1,15 @@
 import useSWR from "swr";
 import appConfig from "../../../../../config/appConfig";
 import { fetcher } from "../../../../../lib/fetch";
-import { PageablePackets } from "../../../../../types";
+import { Packet } from "../../../../../types";
 
-export const useGetPacketPage = (packetName: string | undefined, pageNumber: number, pageSize: number) => {
-  const url = `${appConfig.apiUrl()}/packets/${packetName}?pageNumber=${pageNumber}&pageSize=${pageSize}`;
+export const useGetPacketPage = (packetName: string | undefined) => {
+  const url = `${appConfig.apiUrl()}/packets/${packetName}`;
 
-  const { data, isLoading, error } = useSWR<PageablePackets>(packetName ? url : null, (url: string) =>
-    fetcher({ url })
-  );
+  const { data, isLoading, error } = useSWR<Packet[]>(packetName ? url : null, (url: string) => fetcher({ url }));
 
   return {
-    packetPage: data,
+    packets: data,
     isLoading,
     error
   };
--- a/app/src/app/components/contents/PacketGroup/packetColumns.tsx
+++ b/app/src/app/components/contents/PacketGroup/packetColumns.tsx
@@ -32,6 +32,7 @@ export const packetColumns = [
             setFilter={column.setFilterValue}
             placeholder="Search..."
             inputClassNames="h-8 sm:w-36 lg:w-48"
+            showResetButton={false}
           />
         </div>
       );
@@ -71,6 +72,7 @@ export const packetColumns = [
             setFilter={column.setFilterValue}
             placeholder="Search..."
             inputClassNames="h-8 sm:w-36 lg:w-48"
+            showResetButton={false}
           />
         </div>
       );
--- a/app/src/app/components/contents/PacketGroup/packetColumns.tsx
+++ b/app/src/app/components/contents/PacketGroup/packetColumns.tsx
@@ -1,13 +1,42 @@
-import { createColumnHelper } from "@tanstack/react-table";
+import { createColumnHelper, SortDirection } from "@tanstack/react-table";
+import { ArrowDown, ArrowUp, ArrowUpDown } from "lucide-react";
 import { Link } from "react-router-dom";
 import { Packet } from "../../../../types";
+import { Button } from "../../Base/Button";
+import { FilterInput } from "../common/FilterInput";
 import { ParameterContainer } from "../common/ParameterContainer";
 
 const columnHelper = createColumnHelper<Packet>();
+const getSortingIcon = (sortedDirection: false | SortDirection) => {
+  if (!sortedDirection) return <ArrowUpDown className="ml-2 h-4 w-4" />;
+  return sortedDirection === "asc" ? <ArrowUp className="ml-2 h-4 w-4" /> : <ArrowDown className="ml-2 h-4 w-4" />;
+};
 
 export const packetColumns = [
   columnHelper.accessor("id", {
-    header: "Packet",
+    header: ({ column }) => {
+      const sortedDirection = column.getIsSorted();
+
+      return (
+        <div className="flex flex-col items-start space-y-1 mb-2">
+          <Button
+            variant="ghost"
+            size="sm"
+            className="sm:w-36 lg:w-48 justify-start"
+            onClick={() => column.toggleSorting(sortedDirection === "asc")}
+          >
+            Packet
+            {getSortingIcon(sortedDirection)}
+          </Button>
+          <FilterInput
+            setFilter={column.setFilterValue}
+            placeholder="Search..."
+            inputClassNames="h-8 sm:w-36 lg:w-48"
+          />
+        </div>
+      );
+    },
+    filterFn: "includesString",
     cell: ({ getValue, row }) => {
       const id = getValue();
       const startTime = new Date(row.original.startTime * 1000); // convert from seconds to milliseconds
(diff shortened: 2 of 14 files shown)


app/src/msw/handlers/packetHandler.ts
```diff
@@ -8,8 +8,8 @@ export const packetHandlers = [
   rest.get(`${packetIndexUri}/metadata/${mockPacket.id}`, (req, res, ctx) => {
     return res(ctx.json(mockPacket));
   }),
-  rest.get(`${packetIndexUri}/${mockPacketGroupResponse.content[0].name}`, (req, res, ctx) => {
-    return res(ctx.json(mockPacketGroupResponse));
+  rest.get(`${packetIndexUri}/${mockPacket.name}`, (req, res, ctx) => {
+    return res(ctx.json(mockPacketGroupResponse.content));
```
david-mears-2, line 12: If you're returning something that depends on `mockPacketGroupResponse`, why not stick with the input depending on it too?

app/src/app/components/contents/PacketGroup/packetColumns.tsx
```diff
@@ -1,13 +1,43 @@
...
-    header: "Packet",
+    header: ({ column }) => {
+      const sortedDirection = column.getIsSorted();
+
+      return (
+        <div className="flex flex-col items-start space-y-1 mb-2">
+          <Button
+            variant="ghost"
+            size="sm"
+            className="sm:w-36 lg:w-48 justify-start"
```
david-mears-2, line 25: To my eye, this button and the 'Parameters' div take up too much vertical space - particularly the 'Parameters' text which is further from the input below it than the button is.

I suggest adding `h-fit py-1.5` to each, and make sure both div containers have the same `space-y-[n]` class (they differ).

Because of the background color appearing when the button is hovered, the containers will also need a little padding on top.

app/src/tests/components/contents/PacketGroup/PacketTable.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      })
+    );
+    renderComponent();
+
+    await waitFor(() => {
+      expect(screen.getByText(/unauthorized/i)).toBeVisible();
+    });
+  });
+
+  it("should be able to sort by packet column", async () => {
```
david-mears-2, line 55: This test doesn't test sorting, only that the button icon changes.

app/src/tests/components/contents/PacketGroup/PacketTable.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+  it("should be able to filter by packet column", async () => {
+    const filterSearch = mockPacketGroupResponse.content[1].id.substring(0, 8);
+    renderComponent();
+
+    await screen.findByRole("table");
+
+    userEvent.type(screen.getAllByPlaceholderText("Search...")[0], filterSearch);
+
+    await waitFor(() => {
+      expect(screen.getAllByRole("cell")).toHaveLength(2); // packet + parameter cell
```
david-mears-2, line 84: It's non-trivial for a reader to verify that this length was different before the userEvent. Can you assert this in the test?

app/src/tests/components/contents/PacketGroup/PacketTable.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const filterSearch = mockPacketGroupResponse.content[4].parameters["b"].toLocaleString();
+    renderComponent();
+
+    await screen.findByRole("table");
+
+    userEvent.type(screen.getAllByPlaceholderText("Search...")[1], filterSearch);
+
+    await waitFor(() => {
+      expect(screen.getAllByRole("cell")).toHaveLength(2); // packet + parameter cell
+    });
```
david-mears-2, line 98: Again, it would be nice if the test could verify that this number changed.

app/src/tests/components/contents/common/FilterInput.test.tsx
```diff
@@ -0,0 +1,55 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import { FilterInput } from "../../../../app/components/contents/common/FilterInput";
+import userEvent from "@testing-library/user-event";
+
+describe("FilterByName component", () => {
```
david-mears-2, line 5: Old component name left here

app/src/app/components/contents/common/FilterInput.tsx
```diff
@@ -3,13 +3,22 @@ import { Dispatch, SetStateAction, useCallback, useEffect, useRef } from "react"
 import { Input } from "../../Base/Input";
 import { Button } from "../../Base/Button";
 import { X } from "lucide-react";
+import { cn } from "../../../../lib/cn";
 
 interface FilterByNameProps {
   setFilter: Dispatch<SetStateAction<string>>;
   postFilterAction?: () => void;
   placeholder: string;
+  inputClassNames?: string;
```
david-mears-2, line 12: I won't be a stickler about it but maybe it would be more correct for this to be a list not a string. And it's how the interface of the `cn` function is designed.

app/src/tests/components/contents/common/FilterInput.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+        inputClassNames="sm:w-[1000px] h-14"
+      />
+    );
+
+    const filterInput = await screen.findByPlaceholderText(/placeholder/i);
+    userEvent.type(filterInput, "random");
+
+    await waitFor(() => {
+      expect(screen.queryByRole("button", { name: /reset/i })).not.toBeInTheDocument();
+    });
```
david-mears-2, line 53: This would give more confidence if in another of these tests you asserted that the reset button existed - then we know for sure that this expectation is passing for good reason rather than because e.g. the text of the button was changed.

david-mears-2 (changes requested):

app/src/app/components/contents/PacketGroup/packetColumns.tsx
```diff
@@ -1,13 +1,43 @@
...
 import { ParameterContainer } from "../common/ParameterContainer";
 
-const columnHelper = createColumnHelper<Packet>();
+const SortIcon = ({ sortedDirection }: { sortedDirection: false | SortDirection }) => {
+  if (!sortedDirection) return <ArrowUpDown className="ml-2 h-4 w-4" />;
+  return sortedDirection === "asc" ? <ArrowUp className="ml-2 h-4 w-4" /> : <ArrowDown className="ml-2 h-4 w-4" />;
+};
 
+const columnHelper = createColumnHelper<Packet>();
 export const packetColumns = [
```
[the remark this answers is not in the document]
absternator: because its not really a component....

app/src/msw/handlers/packetHandler.ts
absternator: that true have reverted

app/src/tests/components/contents/PacketGroup/PacketTable.test.tsx
absternator: yeah ive renamed text... because sorting is done by the underlying library (thus not testing) that behaviour

app/src/app/components/contents/common/FilterInput.tsx
absternator: yeah there wont be a point in passing in each class name and spreading because I don't want to merge each one (if needed).... the base components in shadcn all put in multiple classnames into a string as well!!

app/src/tests/components/contents/common/FilterInput.test.tsx
absternator: ive given up spent way too long cant get to to work isolated... this Is already being tested in 4 other components (manageroles, manageusers, runner, home)

app/src/app/components/contents/PacketGroup/hooks/useGetPacketPage.ts
```diff
@@ -1,17 +1,15 @@
 import useSWR from "swr";
 import appConfig from "../../../../../config/appConfig";
 import { fetcher } from "../../../../../lib/fetch";
-import { PageablePackets } from "../../../../../types";
+import { Packet } from "../../../../../types";
```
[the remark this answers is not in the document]
absternator: its still being used for mock data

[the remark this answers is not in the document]
EmmaLRussell: It probably shouldn't be though, should it, if that's not what's coming back from the back end any more - could just be a list of Packet, or whatever..?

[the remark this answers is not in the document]
absternator: its because the index route "/packets" is still paged... its  "/packets/{name}" that changed and  returns the list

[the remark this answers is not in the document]
EmmaLRussell: And we still want to support that route? We're not using that in the website are we - the front page uses the packet group summaries..?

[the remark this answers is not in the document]
absternator: yeah its still there i guess. but I don't see the benefit of removing it (we may need in future)... you right though its not being used by FE currently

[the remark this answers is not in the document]
david-mears-2: Hm, i think there's always benefit to removing unused code, because then you don't have to maintain it. If we want it back again, we can get it from the git history.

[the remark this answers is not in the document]
david-mears-2: I'll tag you since this is a merged PR and you might not otherwise see the comment @absternator @EmmaLRussell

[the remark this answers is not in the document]
EmmaLRussell: We should probably revisit once the current round of work is done and see if there's anything that could either be removed, or moved into the api route...

PR #169, opened 2025-01-24
Mrc-6165 Parameter columns

absternator:
This PR contains work to be able to show parameters as their own columns. The following is done
- allow users to toggle parameters columns on and off 
- allow users to sort / filter on these columns (currently just basic sorting and string filtering)
- if parameter column is visible, then it is removed from overall parameters column pills

image

Testing: 
run up and head to http://localhost:3000/parameters ... play around with toggling on and off columns, filtering and sorting


app/src/tests/components/contents/PacketGroup/PacketTable.test.tsx
```diff
@@ -51,67 +50,4 @@ describe("Packet table", () => {
...
-    renderComponent();
-
-    await screen.findByRole("table");
-
-    userEvent.type(screen.getAllByPlaceholderText("Search...")[1], parameterValue);
-
-    await waitFor(() => {
-      expect(screen.getAllByRole("cell")).toHaveLength(8); // 4 packets with key a parameter
-    });
-  });
```
absternator, line 116: these are moved to new PacketDataTable

app/src/app/components/contents/common/DataTable.tsx
```diff
@@ -16,24 +17,40 @@ interface DataTableProps<TData> {
   pagination?: {
     pageSize: number;
   };
+  visibility?: {
+    columnVisibility: Record<string, boolean>;
+    setColumnVisibility: Dispatch<SetStateAction<Record<string, boolean>>>;
```
EmmaLRussell, line 22: Why does the table need setColumnVisibility? We're not setting from with the table (yet..?)

app/src/app/components/contents/PacketGroup/PacketDataTable.tsx
```diff
@@ -0,0 +1,35 @@
...
+          allParametersKeys={allParametersKeys}
+          columnVisibility={columnVisibility}
+          setColumnVisibility={setColumnVisibility}
+        />
+      </div>
+      <DataTable
+        columns={setupPacketColumns(allParametersKeys)}
+        data={packets}
+        pagination={{ pageSize: PAGE_SIZE }}
+        visibility={{ columnVisibility, setColumnVisibility }}
```
EmmaLRussell, line 30: Ah, so all possible columns are coming from `packetColumns`, but we only show the required parameter columns, with the `visibility` prop here..?

app/src/tests/components/contents/PacketGroup/PacketDataTable.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+    renderComponent();
+
+    await screen.findByRole("table");
+
+    expect(screen.getAllByRole("cell")).toHaveLength(10);
+
+    userEvent.type(screen.getAllByPlaceholderText("Search...")[0], filterSearch);
+
+    await waitFor(() => {
+      expect(screen.getAllByRole("cell")).toHaveLength(2); // packet + parameter cell
```
EmmaLRussell, line 45: I guess you could check for expected content too.

app/src/tests/components/contents/PacketGroup/ToggleParamColumnsDropdown.test.tsx
```diff
@@ -0,0 +1,59 @@
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+// eslint-disable-next-line max-len
+import { ToggleParamColumnsDropdownColumns } from "../../../../app/components/contents/PacketGroup/ToggleParamColumnsDropdown";
```
EmmaLRussell, line 3: Would this line still be too long if you did it like 
```suggestion
import {
  ToggleParamColumnsDropdownColumns
} from "../../../../app/components/contents/PacketGroup/ToggleParamColumnsDropdown";
```
seems a shame to have to put the lint exception for this sort of stuff!
Could we have an `@` type alias for source root?

app/src/tests/components/contents/PacketGroup/utils/initializeColumnVisibility.test.ts
```diff
@@ -0,0 +1,28 @@
+// eslint-disable-next-line max-len
+import { initializeColumnVisibility } from "../../../../../app/components/contents/PacketGroup/utils/initializeColumnVisibility";
```
EmmaLRussell, line 2: ditto

EmmaLRussell (commented):
Cool!

The main "Parameters" column should probably be removed when all parameters have their own columns - currently it sticks around with no content. (and you can still search within the no content!)

It feels like remembering a user's parameter column preferences for a packet group might be worth persisting in local storage..? Definitely in another ticket though.

I feel like you could hide the Parameter Columns toggle entirely if a packet group doesn't have any params, rather than just disabling it.  

Did you mean to commit the orderly files under `demos`?

absternator:
> Cool!
> 
> The main "Parameters" column should probably be removed when all parameters have their own columns - currently it sticks around with no content. (and you can still search within the no content!)
> 
> It feels like remembering a user's parameter column preferences for a packet group might be worth persisting in local storage..? Definitely in another ticket though.
> 
> I feel like you could hide the Parameter Columns toggle entirely if a packet group doesn't have any params, rather than just disabling it.
> 
> Did you mean to commit the orderly files under `demos`?

1. Good idea done 
2. Yup that makes the most sense too... have made ticket here --> https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6192/Persist-user-packet-columns-visibility
3. Feel disabled is a bit better because it keeps the structure of the table and page.. also feels less confusing if users see some tables with toggle and some done... (but if you still disagree happy to change)
4. Yeah I did mean to commit, just more test data that provides variance in the table

app/src/tests/components/contents/PacketGroup/ToggleParamColumnsDropdown.test.tsx
absternator: its an issue with prettier it prefers single imports on 1 line https://github.com/prettier/prettier/issues/1954...

i tried doing alias ages before when did rewrite, but didn't work properly because it uses old version of react... think we have to upgrade project to vite!! made a ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6193/Upgrade-FE-to-vite

EmmaLRussell (approved):
> Feel disabled is a bit better because it keeps the structure of the table and page.. also feels less confusing if users see some tables with toggle and some done... (but if you still disagree happy to change)

Don't feel that strongly about it - I thought it would be more confusing to see  it even if disabled if there's no parameters in the group! But don't think there's much in it.

app/src/app/components/contents/PacketGroup/PacketDataTable.tsx
absternator: sorry my comment didnt save.. but all possible columns actually is the keys of parameters.. and we only show the ones that are toggled via visibility prop yes

PR #170, opened 2025-01-28
Fix the demos directory.

plietar:
Two of the metadata had been reformatted when they were committed to the repository, which broke the hash that is encoded in the `.outpack/location` files. Rewriting the JSON files compactly and removing the trailing newline produces files that do match the expected hash:

```
$ jq -r .hash < .outpack/location/local/20250117-142138-c026f2e0
sha256:c3936c804bebe55bb1ccd0a5caa7ee61014743b57386d84ede62c60d72b2cc24
$ jq -c < .outpack/metadata/20250117-142138-c026f2e0 | tr -d '\n' | sha256sum
c3936c804bebe55bb1ccd0a5caa7ee61014743b57386d84ede62c60d72b2cc24  -
```

Unfortunately orderly2 doesn't seem to have a function to validate the metadata hashes in a way we could run in CI, although this may be something we can implement.

I've removed the RDS index file, since that is only an opaque cache file that can be recreated locally when needed. There was also a submodule in `demos/practice` without any matching entry in `.gitmodules`. Not clear what the reason for it was.


absternator (approved):
OMG f**k that was my bad!! thanks for the fix

PR #171, opened 2025-01-30
Fix security check bypass in findFile endpoint.

plietar:
The `findFile` endpoint takes in a packet ID and a file hash and returns the contents of that file. It performs an access control check on that packet to make sure the user has access to it.

However, after permissions are checked, the packet ID was never used again: fetching a file from outpack server only requires the file hash.

This creates a vulnerability where one could provide a packet ID they have access to to download any file, provided they know the hash of the file. To fix this, we need to check the packet's metadata to make sure it really contains that file, and therefore the user is allowed to read those contents.

I clarified / simplified some of the authorization logic. There used to be two methods, named `canReadPacket` and `canReadPacketMetadata`. The existence of the latter suggests that there are cases in which one can see a packet (ie. read its metadata) but not get its contents. This is not the case and users can read a packet's metadata if and only if they can read the whole packet.

In reality the two methods provided the same functionality, and `canReadPacketMetadata` was even implemented in terms of `canReadPacket`. The real difference was that `canReadPacket` requires the caller to know the name of the packet group already, whereas `canReadPacketMetadata` performs a database lookup.

Since the behaviour is the same, I've changed them to have the same `canReadPacket` name, removing any mention of "metadata". Overload on the type of the arguments is used to distinguish between the two.

Finally, the version of the method that accepted a `name` argument was error-prone: it assumes that the packet with the given ID actually has the given name. However it would be easy to write an endpoint that accepts the two values as parameters and passes them both to the security filter. In this case the attacker could pick a name they have access to to gain access to any packet ID. The method should really only be used in cases where we've already performed a database lookup for the packet. Using the `Packet` class as the type of the unique argument makes this clearer.

### DIFF
--- a/api/app/src/main/kotlin/packit/controllers/PacketController.kt
+++ b/api/app/src/main/kotlin/packit/controllers/PacketController.kt
@@ -2,9 +2,11 @@ package packit.controllers
 
 import org.springframework.core.io.ByteArrayResource
 import org.springframework.data.domain.Page
+import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.*
+import packit.exceptions.PackitException
 import packit.model.PacketMetadata
 import packit.model.PageablePayload
 import packit.model.dto.PacketDto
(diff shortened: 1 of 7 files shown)


absternator (approved):
works like a charm!! LGTM!

PR #172, opened 2025-01-31
Tidy up role controller endpoints.

plietar:
This renames the role endpoint from `/role` to `/roles`, bringing it inline with the existing `/packets` and `/packetGroups`.

This also removes the `GET /roles/names` endpoint, which was not being used anywhere, and which overlapped with the existing `GET /roles/<name>` which is used to get details about a particular role. Having both routes could cause issue is a role was named `names`.

### DIFF
--- a/api/app/src/main/kotlin/packit/controllers/RoleController.kt
+++ b/api/app/src/main/kotlin/packit/controllers/RoleController.kt
@@ -24,7 +24,7 @@ class RoleController(private val roleService: RoleService, private val userRoleS
     {
         val role = roleService.createRole(createRole)
 
-        return ResponseEntity.created(URI.create("/roles/${role.id}")).body(role.toDto())
+        return ResponseEntity.created(URI.create("/roles/${role.name}")).body(role.toDto())
     }
 
     @DeleteMapping("/{roleName}")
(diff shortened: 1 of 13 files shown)


api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -16,15 +16,15 @@ import java.net.URI
...
+@RequestMapping("/roles")
 class RoleController(private val roleService: RoleService, private val userRoleService: UserRoleService)
 {
     @PostMapping()
     fun createRole(@RequestBody @Validated createRole: CreateRole): ResponseEntity<RoleDto>
     {
         val role = roleService.createRole(createRole)
 
-        return ResponseEntity.created(URI.create("/role/${role.id}")).body(role.toDto())
+        return ResponseEntity.created(URI.create("/roles/${role.id}")).body(role.toDto())
```
EmmaLRussell, line 27: Shouldn't the uri here use the role name rather than id as per GetMapping below? (If so, this was a bug before this changeset!)

```suggestion
        return ResponseEntity.created(URI.create("/roles/${role.name}")).body(role.toDto())
```

EmmaLRussell (commented):
Great stuff, just one query.

api/app/src/main/kotlin/packit/controllers/RoleController.kt
plietar: Good catch yes. Thanks

EmmaLRussell (approved):

PR #173, opened 2025-01-31
Tidy up packet and packet group controllers.

plietar:
The main motivation for this change was to remove some ambiguity in the old routes. In particular, we had both the `/packets/<name>` and `/packets/packetGroupSummaries` routes.

Some of the PacketController routes had more to do with packet groups and have been moved to the appropriate controller.

The PacketController routes had `/metadata/<id>` and `/file/<id>` routes, which have now been replaced by `/<id>` and `/<id>/file`, which I think is more consistent with the rest of the API.

This is an overview of the routes after these changes:
- `/packets`: returns a list of all packets, searchable by id / name.
- `/packets/<id>`: returns a packet's metadata
- `/packets/<id>/file`: returns a file from a packet
- `/packetGroups`: returns the list of packet groups, searchable by name. Only returns the id and name of the group.
- `/packetGroups/<name>/display`: returns high-level information of the group, ie. the display name and description.
- `/packetGroups/<name>/packets`: returns the list of packets in that group.
- `/packetGroupSummaries`: returns a list of packet groups, searchable by name and display name. Includes a "summary" for each group, including the number of packets and the latest packet ID.

I find the duplication between `/packetGroups` and `/packetGroupSummaries` a bit unfortunate. I think these should be merged into a single endpoint. I have left this for a future PR.

The `/packets` and `/packetGroups` endpoints are a bit special in that they are used by the frontend way in a generic way to set scoped permissions, therefore must have a relatively uniform interface (together with the `/tag` endpoint).

### DIFF
--- a/api/app/src/main/kotlin/packit/controllers/PacketGroupController.kt
+++ b/api/app/src/main/kotlin/packit/controllers/PacketGroupController.kt
@@ -5,21 +5,23 @@ import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Controller
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
-import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import packit.model.PageablePayload
+import packit.model.dto.PacketDto
 import packit.model.dto.PacketGroupDisplay
 import packit.model.dto.PacketGroupDto
+import packit.model.dto.PacketGroupSummary
 import packit.model.toDto
 import packit.service.PacketGroupService
+import packit.service.PacketService
 
 @Controller
-@RequestMapping("/packetGroups")
 class PacketGroupController(
+    private val packetService: PacketService,
     private val packetGroupService: PacketGroupService,
 )
 {
-    @GetMapping
+    @GetMapping("/packetGroups")
     fun getPacketGroups(
         @RequestParam(required = false, defaultValue = "0") pageNumber: Int,
         @RequestParam(required = false, defaultValue = "50") pageSize: Int,
@@ -30,12 +32,33 @@ class PacketGroupController(
         return ResponseEntity.ok(packetGroupService.getPacketGroups(payload, filterName).map { it.toDto() })
     }
 
-    @GetMapping("/{name}/display")
+    @GetMapping("/packetGroups/{name}/display")
     fun getDisplay(
         @PathVariable name: String
     ): ResponseEntity<PacketGroupDisplay> {
         val result = packetGroupService.getPacketGroupDisplay(name)
 
         return ResponseEntity.ok(result)
+    }
+
+    @GetMapping("/packetGroups/{name}/packets")
+    fun getPackets(
+        @PathVariable name: String,
+    ): ResponseEntity<List<PacketDto>>
+    {
+        return ResponseEntity.ok(
+            packetService.getPacketsByName(name).map { it.toDto() }
+        )
+    }
+
+    @GetMapping("/packetGroupSummaries")
+    fun getPacketGroupSummaries(
+        @RequestParam(required = false, defaultValue = "0") pageNumber: Int,
+        @RequestParam(required = false, defaultValue = "50") pageSize: Int,
+        @RequestParam(required = false, defaultValue = "") filter: String,
+    ): ResponseEntity<Page<PacketGroupSummary>>
+    {
+        val payload = PageablePayload(pageNumber, pageSize)
+        return ResponseEntity.ok(packetGroupService.getPacketGroupSummaries(payload, filter))
     }
 }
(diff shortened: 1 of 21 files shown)


api/app/src/main/kotlin/packit/controllers/PacketGroupController.kt
```diff
@@ -30,12 +32,33 @@ class PacketGroupController(
...
+    fun getPackets(
+        @PathVariable name: String,
+    ): ResponseEntity<List<PacketDto>>
+    {
+        return ResponseEntity.ok(
+            packetService.getPacketsByName(name).map { it.toDto() }
+        )
+    }
+
+    @GetMapping("/packetGroupSummaries")
```
EmmaLRussell, line 54: You'd rather have this than `/packetGroups/summaries`? At least then everything in this controller would have the same base route. I don't think that would be ambiguous as no other routes have two parts under `packetGroups`.

EmmaLRussell (commented):
Looks pretty tidy to me. I think metadata being the default packet endpoint makes sense.

> I find the duplication between /packetGroups and /packetGroupSummaries a bit unfortunate. I think these should be merged into a single endpoint. I have left this for a future PR.

Yes, and the separate endpoint for packetGroup `display` seems a bit odd too - but I think that was for technical reasons as it requires a call to outpack server. But it feels like that should eventually be rolled into summaries, and that summaries should probably be the main packetGroups endpoint. 

> The /packets and /packetGroups endpoints are a bit special in that they are used by the frontend way in a generic way to set scoped permissions

You mean the same bit of code in the front end uses the GET responses from both endpoint to find items which permissions can be scoped over? It's probably a good idea for these to be consistent anyway! But I guess we could put some reformatting translation in the front end for that purpose if we needed to.  

> together with the /tag endpoint

Isn't `/tag` on hold at the moment as we aren't entirely sure what to do with it?

api/app/src/main/kotlin/packit/controllers/PacketGroupController.kt
plietar: Yes, even if it isn't technically ambiguous today, conceptually I think it is cleaner if we reserve the `/packetGroups` route to only have packet names as the next segment. It seems likely that we'll have a `/packetGroups/<name>` in the near future (eg. by renaming the `/display` route) or maybe even a `/packetGroupSummaries/<name>`, in which case there will be an actual ambiguity.

plietar:
> > The /packets and /packetGroups endpoints are a bit special in that they are used by the frontend way in a generic way to set scoped permissions
> 
> You mean the same bit of code in the front end uses the GET responses from both endpoint to find items which permissions can be scoped over? It's probably a good idea for these to be consistent anyway! But I guess we could put some reformatting translation in the front end for that purpose if we needed to.

Yes that's what I meant. I have no problem with it, just thought it was useful to mention.

One oddity implied by this is that the `/packetGroups` endpoint returns an "id" field, which is a numerical index that is internal to our database. I don't think this really belongs in the REST api, but removing it would break the FE code that assumes that "permission scopes" (ie. packets/packetGroups/tags) all have an ID. We can easily remove that assumption from the FE is we really wanted to remove the ID.

> > together with the /tag endpoint
> 
> Isn't `/tag` on hold at the moment as we aren't entirely sure what to do with it?

The endpoint implementation exists and it does a database lookup, but that always returns an empty result since the table isn't populated. But yeah it is basically a placeholder and not really something we should worry too much about in its current form.

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
```diff
@@ -13,78 +12,40 @@ import org.springframework.http.HttpStatus
...
-                    0.0,
-                    0.0,
-                    0.0
-                ),
-            )
-        )
-        packetGroups = packetGroupRepository.saveAll(
-            packets.map { it.name}.distinct().map{PacketGroup(it)}
-        )
+        packetService.importPackets()
```
EmmaLRussell, line 41: Sorry, I can't quite remember how these integration tests were set up. Do the ids defined above happen to exist in the demo dataset so just doing importPackets here suffices rather than stuffing the packets into the repo as before?

EmmaLRussell (approved):
LGTM, just one question.

api/app/src/test/kotlin/packit/integration/controllers/PacketControllerTest.kt
plietar: Yes. This test was doing a mix of using the files that are on disk and inserting a bunch of fake entries in the database.

Any test that leads to requests to outpack server will need the files to exist in the repository anyway, so it seems simpler to just use that consistently and not try to setup these fake packets that don't exist on disk. Ideally we'd create those packets programmatically in the test setup instead of checking them in to Git, but there is nothing in Packit that is capable of doing that.

PR #174, opened 2025-02-04
Use hard-coded email and password when creating the super user.

plietar:
There's no good reason for our local development workflow to depend on Vault secrets. We can use a hardcoded email and password for the admin account.

It's unclear to me whether we can do the same thing for GitHub OAuth client ID / secret, ie. use a dedicated test OAuth app and commit its client secret into the repository, for use in development and testing only. I've avoided doing so for now.


M-Kusumgar (approved):
awesome thank you!

PR #175, opened 2025-02-10
mrc-6126 update interface to orderly runner

M-Kusumgar:
Updating the interface based on https://github.com/mrc-ide/orderly.runner/pull/17

### DIFF
--- a/api/app/src/main/kotlin/packit/model/dto/RepositoryFetch.kt
+++ b/api/app/src/main/kotlin/packit/model/dto/RepositoryFetch.kt
@@ -1,3 +1,8 @@
 package packit.model.dto
 
-data class RepositoryFetch(val ssh_key: String?)
+import com.fasterxml.jackson.annotation.JsonProperty
+
+data class RepositoryFetch(
+    @JsonProperty("ssh_key")
+    val sshKey: String?
+)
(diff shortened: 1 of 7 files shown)


plietar (approved):

api/app/src/main/kotlin/packit/model/dto/RepositoryFetch.kt
```diff
@@ -1,3 +1,8 @@
 package packit.model.dto
 
-data class RepositoryFetch(val url: String)
+import com.fasterxml.jackson.annotation.JsonProperty
+
+data class RepositoryFetch(
+    @JsonProperty("ssh_key")
```
plietar, line 6: Can we not just tell jackson to snake case everything?

plietar: https://www.baeldung.com/jackson-deserialize-snake-to-camel-case#use-jsonnaming-annotation

I think I prefer using this tbh

M-Kusumgar: sure i can maybe add it here, i cant add it to the submit run one because it deserialises `hash` as `commitHash` so doesnt conform to that, i guess we could change this potentially, the annoying this is that the frontend also has to use snake case in javascript right? ugly af man

plietar: Yeah I'm happy to do it some other time

PR #176, opened 2025-02-11
Use streaming to download single files

david-mears-2:
This is required in order to download larger files without running out of memory.

I replace the non-streaming `/packets/:id/file` endpoint with a streaming version.

#### How to manually test

You want to verify that the server does not run out of memory when transferring large files. To do this, follow the instructions in `demos/README.md` in the section `Using the 'massive-files' report for testing how the app copes with very large files`


david-mears-2:
Todo, rename the component as per https://github.com/mrc-ide/packit/pull/161#discussion_r1954252965

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -62,6 +62,29 @@ class PacketController(private val packetService: PacketService)
             .body(response.first)
     }
 
+    @GetMapping("/{id}/file/{hash}/stream")
+    @PreAuthorize("@authz.canReadPacket(#root, #id)")
+    fun streamFile(
```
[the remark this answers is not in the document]
david-mears-2: Well, they `return` different things (a ResponseEntity vs nothing), and that's a clue that what they do is fairly different.

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -152,25 +153,23 @@ class BasePacketService(
     override fun getFileByHash(hash: String, inline: Boolean, filename: String): Pair<ByteArrayResource, HttpHeaders>
```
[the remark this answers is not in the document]
david-mears-2: We do use non-streaming file downloads for a few things:
- previewing reports on a packet's summary page
- previewing image files on the downloads page
- displaying image files on the page for doing that

api/app/src/test/kotlin/packit/unit/controllers/PacketControllerTest.kt
```diff
@@ -76,22 +78,30 @@ class PacketControllerTest
...
+            getFileByHash(
+            anyString(),
+            any<OutputStream>(),
+            any<(HttpHeaders) -> Unit>()
+        )
+        } doAnswer { invocationOnMock ->
+            val outputStream = invocationOnMock.getArgument<OutputStream>(1)
+            outputStream.write("mocked output content".toByteArray())
+            val callback = invocationOnMock.getArgument<(HttpHeaders) -> Unit>(2)
+            callback(testHeaders)
```
plietar, line 95:
```suggestion
            val callback = invocationOnMock.getArgument<(HttpHeaders) -> Unit>(2)
            callback(testHeaders)
            val outputStream = invocationOnMock.getArgument<OutputStream>(1)
            outputStream.write("mocked output content".toByteArray())
```

I don't know how much it matters in this unit test, but I think it makes sense to match the order these things happen in the real implementation. In general you can't write HTTP headers (including sending the content length) once the response has been ["committed"](https://docs.oracle.com/javaee/7/api/javax/servlet/ServletResponse.html#isCommitted--)

api/app/src/test/kotlin/packit/unit/controllers/PacketControllerTest.kt
```diff
@@ -76,22 +78,30 @@ class PacketControllerTest
...
+        }
         on { getPacketsByName(anyString()) } doReturn packets
 
         packetMetadata.forEach {
             on { getMetadataBy(it.id) } doReturn it
         }
     }
 
+    private val response = MockHttpServletResponse()
```
plietar, line 104: Doesn't this making a global value cause problems with it being shared across tests. For example on test runs, passes it to a controller method that calls `setHeader` on it, the next test runs and it inherits sees the header. eg. does the response in the `streamZip` test have a left over content length from the `streamFile` tests?

I think it would be safer to create one in each test function.

(Usually we get away with global mocks and SUTs because they are immutable objects, but this is not)

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -69,8 +70,11 @@ class PacketController(private val packetService: PacketService)
         @RequestParam paths: List<String>,
         response: HttpServletResponse
     ) {
-        response.contentType = "application/zip"
-        response.setHeader("Content-Disposition", "attachment; filename=$id.zip")
+        val headers = HttpHeaders().apply {
+            contentType = MediaType.valueOf("application/zip")
+            contentDisposition = ContentDisposition.parse("attachment; filename=$id.zip")
```
plietar, line 75: contentDisposition = ContentDisposition.attachment().filename("$id.zip").build()

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -69,8 +70,11 @@ class PacketController(private val packetService: PacketService)
...
+        }
+        headers.map { response.setHeader(it.key, it.value.first()) }
```
plietar, line 77: What was the motivation behind this change? Creating a headers map just to iterate over it afterwards seems more complicated than the existing version.

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -43,23 +38,29 @@ class PacketController(private val packetService: PacketService)
...
-        val response = packetService.getFileByHash(hash, inline, filename)
-        return ResponseEntity
-            .ok()
-            .headers(response.second)
-            .body(response.first)
+        val disposition = if (inline) "inline" else "attachment"
+        val mediaType = getMediaType(filename).orElse(MediaType.APPLICATION_OCTET_STREAM)
+        val headers = HttpHeaders().apply {
+            contentType = MediaType.valueOf(mediaType.toString())
+            contentDisposition = ContentDisposition.parse("$disposition; filename=$filename")
```
plietar, line 57:
```suggestion
            contentDisposition = ContentDisposition.builder(disposition).filename(filename).build()
```

This will behave better if the filename contains odd characters (including eg. a quotation mark or a semicolon)

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -43,23 +38,29 @@ class PacketController(private val packetService: PacketService)
...
 
-        val response = packetService.getFileByHash(hash, inline, filename)
-        return ResponseEntity
-            .ok()
-            .headers(response.second)
-            .body(response.first)
+        val disposition = if (inline) "inline" else "attachment"
+        val mediaType = getMediaType(filename).orElse(MediaType.APPLICATION_OCTET_STREAM)
+        val headers = HttpHeaders().apply {
+            contentType = MediaType.valueOf(mediaType.toString())
```
plietar, line 56:
```suggestion
            contentType = mediaType
```

or am I missing something?

api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -43,23 +38,29 @@ class PacketController(private val packetService: PacketService)
...
+            contentDisposition = ContentDisposition.parse("$disposition; filename=$filename")
+        }
+        headers.map { response.setHeader(it.key, it.value.first()) }
```
plietar, line 59: Honestly I would find it more readable if you just had two calls to `response.setHeader`, rather than build a map and iterate over it.

plietar (commented):
Overall design seems good to me now. Just some more minor comments

api/app/src/main/kotlin/packit/controllers/PacketController.kt
david-mears-2: As far as possible, I wanted to make use of the libraries written by other people, that ensure the strings are parsed as valid http headers, rather than pass any old string into `response.setHeader`. Underlyingly, `contentType = ` makes use of `HttpHeaders.setContentType` (which performs a few validation checks).

We can instead use `response.contentType = `.

The other reason is the typing of `response.setHeader` makes it pretty ugly with `toString` everywhere. The full change would read as:

```kt
response.contentType = MediaType.valueOf(mediaType.toString()).toString()
response.setHeader("Content-Disposition", ContentDisposition.attachment().filename("$id.zip").build().toString())
```

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -139,7 +135,7 @@ class BasePacketService(
             ZipOutputStream(output).use { zipOutputStream ->
                 hashesByPath.forEach { (filename, hash) ->
                     zipOutputStream.putNextEntry(ZipEntry(filename))
-                    outpackServerClient.getFileByHash(hash, zipOutputStream)
+                    outpackServerClient.getFileByHash(hash, zipOutputStream) {}
```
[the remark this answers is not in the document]
david-mears-2: I think the callback has to be defined in the controller, because it makes use of the response object - it would be annoying to have to pass this response object through the layers as well.

david-mears-2:
The codecov complaint is because I marked a test as pending. I created a ticket for fixing that test: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6223/Fix-flakey-test

EmmaLRussell (commented):
Downloading massive csv file now works (with limited memory), but I'm finding that downloading the massive zip file isn't working - spins forever and doesn't appear in browser list of downloads. Is it working for you? I may have got my local packit into a weird state.

david-mears-2:
> Downloading massive csv file now works (with limited memory), but I'm finding that downloading the massive zip file isn't working - spins forever and doesn't appear in browser list of downloads. Is it working for you? I may have got my local packit into a weird state.

Downloading the file itself took about 10 seconds. The zip took about 60 seconds. (Zip file size was about half uncompressed file size.) I got the same results whether I used reduced memory or not.

You won't see the files in the browser list of downloads until the download is complete. This is possibly because the browser isn't paying attention to the Content-Length header in the case of streaming? I don't know.

I guess the extra processing of compressing the file is causing the relative slow-down in the case of zip downloads. Is the zip feature even useful given how much longer it takes to transfer the file? Maybe if the network speed is a bottleneck, zipping is preferable, because there's about half as much data to transfer.

I tried simulating a slow network using Chrome dev tools, but even trying to download a single massive file using the so-called 'Fast 4G' preset took over 3 minutes before I gave up.

I can imagine an iteration where the choice of whether to zip the files in the response is determined by the total file size rather than by whether there is more than one file. But then if you click 'download all' you might get lots of separate little files instead of a neatly wrapped up archive folder.

api/app/src/main/kotlin/packit/service/GenericClient.kt
```diff
@@ -89,19 +89,19 @@ object GenericClient
...
                 serverResponse.body.use { input ->
                     IOUtils.copy(input, output)
                 }
             }
         } catch (e: HttpStatusCodeException)
         {
-            throw GenericClientException(e)
-        } catch (e: Exception) {
-            throw PackitException("couldNotStream", HttpStatus.INTERNAL_SERVER_ERROR)
+            log.error("Error response: {}", e.statusText)
```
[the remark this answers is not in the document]
david-mears-2: I see we have a function `handlePackitException` annotated with `@ExceptionHandler(PackitException::class)`, but I'm not sure if it does any logging. It doesn't seem to log anything to the logs of the packit-api container, while `log.error` does. Should we add logging to the `handlePackitException` class so that we can inspect any errors using the logs of the packit api?

api/app/src/main/kotlin/packit/service/GenericClient.kt
```diff
@@ -89,19 +89,19 @@ object GenericClient
         }
     }
 
-    fun streamingGet(url: String, output: OutputStream) {
+    fun streamingGet(url: String, output: OutputStream, copyHeaders: (HttpHeaders) -> Unit) {
         try
         {
             restTemplate.execute(URI(url), HttpMethod.GET, {}) { serverResponse ->
+                copyHeaders(serverResponse.headers)
```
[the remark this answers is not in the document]
david-mears-2: I feel averse to passing the response object through all the other layers (which, for separation of concerns reasons, should not know about http responses) -- but perhaps passing a lambda expression through those layers is just as bad? At least with using a lambda, it only has to be passed down in the case of downloading files; if we made the response a parameter, it would also need to be passed down through the packetService.streamZip function, in order to call outpackServerClient.getFileByHash.

(There's also a slight benefit to having all the code that sets headers be done in the controller method.)

But as you say, this lambda is only being used for one purpose, which means its pretence at generality is only a pretence.

I think I would like to preserve packetService.streamZip from knowing about responses, so I prefer using a lambda. I'll make it a bit more truly generic by renaming it to `preStream` and making it take a ClientHttpResponse parameter rather than a HttpHeaders parameter.

david-mears-2:
> I guess the extra processing of compressing the file is causing the relative slow-down in the case of zip downloads.

Seems likely, because zipping the file using the Ubuntu interface takes just over 50 seconds.

plietar:
> You won't see the files in the browser list of downloads until the download is complete.

I think this will be because of the blob URL thing:
https://github.com/mrc-ide/packit/blob/2383016d14350d62ececb318dbafdf21311a9134/app/src/lib/download.ts#L30-L39

The download happens entirely within the JS, and only when that is complete it gets put into an "Object URL" and then browser gets told to download that (which will happen almost immediately).

> This is possibly because the browser isn't paying attention to the Content-Length header in the case of streaming?

Even without a Content-Length (eg. in the case of downloading a zip), browsers usually show the download. They just won't give you a progress bar for it.

plietar:
> The download happens entirely within the JS, and only when that is complete it gets put into an "Object URL" and then browser gets told to download that (which will happen almost immediately).

The reason we do that is because we need to pass the bearer token as a header, and we can't put headers in download URLs.

The typical way around it is have a POST request that can create a very short lived download token, then pass that token in the download URL, eg. `/packets/xxx/file?hash=yyy&token=zzz`. The server would accept that request without a header bearer token.

I'm pretty sure I saw an implementation of that in montagu actually, I'll try to dig it up.

plietar:
The main docs for the montagu implementation: https://github.com/vimc/montagu-api/blob/c0a655d87955acc059c336c197deddbcdaf4a1f9/docs/spec/OneTimeToken.md?plain=1#L1

Chasing around the implementation:
https://github.com/vimc/montagu-api/blob/c0a655d87955acc059c336c197deddbcdaf4a1f9/src/app/src/main/kotlin/org/vaccineimpact/api/app/controllers/OnetimeTokenController.kt
https://github.com/vimc/montagu-api/blob/c0a655d87955acc059c336c197deddbcdaf4a1f9/src/app/src/main/kotlin/org/vaccineimpact/api/app/security/OneTimeTokenGenerator.kt#L18-L24
https://github.com/vimc/montagu-api/blob/c0a655d87955acc059c336c197deddbcdaf4a1f9/src/security/src/main/kotlin/org/vaccineimpact/api/security/WebTokenHelper.kt#L53-L67

plietar (approved):
This LGTM. We should fix the browser bit of it separately.

I don't mind too much the `log.error` in `GenericClient`, although at some point I think we should re-think the whole logging of the API (eg. print access logs for all incoming and outgoing requests).

david-mears-2:
Thanks for that information @plietar - I've made a new ticket for the browser downloading [here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6224/Use-browsers-native-download-instead-of-JS-blob).

api/app/src/main/kotlin/packit/service/GenericClient.kt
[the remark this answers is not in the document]
EmmaLRussell: Yeah I guess the GenericClient in theory doesn't care what type of OutputStream it's writing to, so it could be writing to a file or whatever. Fair point on sep of concerns, the controller wouldn't usually pass the response around. I think it just comes down to whether you're more grossed out by callbacks or layer pollution!

api/app/src/main/kotlin/packit/service/GenericClient.kt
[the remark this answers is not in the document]
EmmaLRussell: Hm, sorry I assumed we were logging from the exception handler. Yeah, we should be I think!

EmmaLRussell (approved):
Yes, it did download eventually, I just needed to be more patient!

PR #177, opened 2025-02-12
Add needed env vars for packit deploy

M-Kusumgar:
Just adding a couple of env vars that packit-infra (packit deploy via nix) can use to override the orderly runner related application properties


absternator (approved):
LGTM!

plietar:
You don't need those: https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.typesafe-configuration-properties.relaxed-binding.environment-variables

Setting `ORDERLY_RUNNER_ENABLED` will work already, or better even you can use the `properties` option in the Nix `packit-api` module. See for example how I set up the `auth.service` properties.

PR #178, opened 2025-02-13
add sub url depth var and remove public url and namespace env vars

M-Kusumgar:
BAAM, relative homepage path in package.json gets picked up in react scripts build! now we have relative paths and this works


app/package.json
```diff
@@ -76,7 +76,7 @@
     "format:check": "prettier --check .",
     "format:write": "prettier --write ."
   },
-  "homepage": "/",
+  "homepage": "./",
```
M-Kusumgar, line 79: :heart:

plietar:
I don't think this works?

If I run `npm run build`, I get an `index.html` file that starts like this:
```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8"/>
    <link rel="icon" href="./favicon.ico"/>
    [...]
    <script defer="defer" src="./static/js/main.239faae1.js"> </script>
    <link href="./static/css/main.feb55f2f.css" rel="stylesheet">
  </head>
```

All of the assets are relative paths, as expected given the value of `homepage`.

This will work fine if the browser loads `packit.dide.ic.ac.uk/reside/`, but if the user follows a link to a nested page eg. `/reside/foo/20241022-150903-5d41a8ad`, then the browser will load the JS file at `/reside/foo/static/js/main.239faae1.js`, which will not give use what we expected.

If we want to have client side routing based on URL sub-directories, we need PUBLIC_URL (or homepage, they are essentially the same thing) to be an absolute path since index.html ends up being rendered at arbitrary paths.

(I'm sure there are awful tricks we could do in nginx to get this working, but I'd rather not go down that route)

M-Kusumgar:
closing this PR, it indeed does not work, there are work arounds we can put in place but its not worth the grossness introduced, just putting the ideas we talked about here:
* have a `proconfigure.js` script that parses the url and appends the appropriate script tag into the html
* have nginx change the html on the fly

PR #179, opened 2025-02-17
mrc-4333 e2e tests

EmmaLRussell:


app/e2e/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+  return section;
+};
+
+export const selectPacketPageTab = async (content: Locator, tab: string) => {
+  const nav = await content.getByRole("navigation");
+  await nav.getByRole("link", { name: tab }).click();
+};
+
+export const getPackitInstance = (baseURL: string) => {
```
plietar, line 51: At first I was a bit confused why this would be needed, and then I saw the code that uses it to find the instance-relative path. I would suggest changing this function to doing that operation here instead, especially since it will come up in more places.

```js
export const getRelativePath = (baseURL: string, href: string) {
    const basePath = new URL(baseURL).pathname;
    return href.startsWith(basePath) ? basePath.slice(path.LENGTH) : href;
}
```

app/playwright.config.ts
```diff
@@ -0,0 +1,83 @@
+import process from "node:process";
+import { defineConfig, devices } from "@playwright/test";
+import path from "path";
+
+const setupProject = { name: "setup", testMatch: /.*\.setup\.ts/ };
+
+const outputDir = "test-results";
+const authStorageStateFile = path.join(__dirname, outputDir, `auth.setup.ts-authenticate-setup/auth.json`);
+const baseURL = process.env.PACKIT_E2E_BASE_URL || "http://localhost:3000";
+const testMatch = baseURL.startsWith("http://localhost") ? /.*\.(local|spec)\.ts/ : /.*\.spec\.ts/;
```
plietar, line 10: I think this should be an environment variable (possibly with a fall back based on the URL).
We can arrange for an instance on `packit-dev` to mirror the contents of this repository, in which case we'll be able to run those tests there too.

app/e2e/auth.setup.ts
```diff
@@ -0,0 +1,29 @@
+import { test as setup, expect } from "@playwright/test";
+
+const authMethod = process.env.PACKIT_E2E_AUTH_METHOD || "basic";
+const basicUser = process.env.PACKIT_E2E_BASiC_USER || "resideUser@resideAdmin.ic.ac.uk";
```
plietar, line 4:
```suggestion
const basicUser = process.env.PACKIT_E2E_BASIC_USER || "resideUser@resideAdmin.ic.ac.uk";
```

there's a typo there on the I being lowercase instead of uppercase

app/e2e/packetPage.local.ts
```diff
@@ -0,0 +1,100 @@
...
+    test("can see dependencies", async ({ page }) => {
+      await page.goto(`/depends/${dependsPacketId}`);
+      const content = await getContentLocator(page);
+      const dependsDiv = await getPacketPageAccordionSection(content, "Dependencies");
+      const depItems = await dependsDiv.getByRole("listitem");
+      await expect(depItems).toHaveCount(2);
+      const dep1Name = "explicit";
+      const dep1Id = "20240729-154639-25b955eb";
+      await expect(await depItems.first()).toHaveText(`${dep1Name}${dep1Id}`);
+      await expect(await depItems.first().getByRole("link")).toHaveAttribute("href", `/${dep1Name}/${dep1Id}`);
```
plietar, line 36: It would be nice to use `getPackitInstance` (or equivalent) consistently so we can run this against `packit-dev` as well.

scripts/run-e2e-tests-on-server
```diff
@@ -0,0 +1,38 @@
...
+export PACKIT_E2E_AUTH_METHOD=$AUTH_METHOD
+
+echo "Server is $BASE_URL. Auth method is $AUTH_METHOD."
+
+if [[ $AUTH_METHOD == "basic" ]]; then
+  echo "Please provide user credentials for running e2e tests. User will need to have admin permissions in Packit."
+  read -p 'Username: ' BASIC_USER
+  read -sp 'Password (will be hidden):' BASIC_PASSWORD
+  echo
+  export PACKIT_E2E_BASiC_USER=$BASIC_USER
```
plietar, line 28:
```suggestion
  export PACKIT_E2E_BASIC_USER=$BASIC_USER
```

scripts/run-e2e-tests-on-server
```diff
@@ -0,0 +1,38 @@
...
+export PACKIT_E2E_BASE_URL=$BASE_URL
+export PACKIT_E2E_AUTH_METHOD=$AUTH_METHOD
+
+echo "Server is $BASE_URL. Auth method is $AUTH_METHOD."
+
+if [[ $AUTH_METHOD == "basic" ]]; then
+  echo "Please provide user credentials for running e2e tests. User will need to have admin permissions in Packit."
+  read -p 'Username: ' BASIC_USER
+  read -sp 'Password (will be hidden):' BASIC_PASSWORD
+  echo
```
plietar, line 27: Is this necessary, especially given that `auth.setup.ts` applies some defaults to it?

scripts/run-e2e-tests-on-server
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
plietar, line 1:
```suggestion
#!/usr/bin/env bash
set -eu
```

always a good idea to have these

app/playwright.config.ts
plietar: Alternatively we could use tags for this: https://playwright.dev/docs/test-annotations#tag-tests

This has two benefits: it allows us to specify whether or not to enable them on the command line instead of guessing based on the hostname, and it means we don't have to partition our tests into different files (especially when we introduce tests that cause mutations, which will add yet another axis along which tests are partitioned).

plietar:
This all looks pretty reasonable, mostly just some suggestions.

.gitignore
```diff
@@ -1,3 +1,6 @@
 .idea
 .vscode
-packit.iml
\ No newline at end of file
+.token
+packit.iml
+app/test-results
+app/playwright-report
```
plietar, line 6: Would these be useful to upload as artefacts of the CI?

eg. https://github.com/mrc-ide/packit/blob/377f8e487a894825f04a579f51e25d5e557f6e40/.github/workflows/backend-test-and-build.yml#L72-L81

app/e2e/index.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    // expect to have at least one packet group remaining, and expect all to have filter term  as a substring
+    expect(await filteredGroups.count()).toBeGreaterThan(0);
+    for (const packetGroup of await filteredGroups.all()) {
+      expect(await packetGroupNameFromListItem(packetGroup)).toContain(firstPacketGroupName);
+    }
+  });
+
+  test("can navigate from packet group name link to packet group page", async ({ page }) => {
+    const firstPacketGroupName = await navigateToFirstPacketGroup(content);
+    const displayName = getDisplayString(firstPacketGroupName);
```
david-mears-2, line 47: I'm confused by this. It looks like we are programmatically deriving a display name from a name. But display names can be anything.

app/e2e/packetGroupPage.local.ts
```diff
@@ -0,0 +1,73 @@
...
+    await expect(parameterTexts[0]).toBe("a:  1");
+    await expect(parameterTexts[1]).toBe("b:  2");
+    await expect(parameterTexts[2]).toBe("c:  3");
+  });
+
+  test("can search packet rows by id", async () => {
+    // Enter packet id in input
+    const packetInput = await rows.first().getByRole("cell").first().getByPlaceholder("Search...");
+    await packetInput.fill(packetId);
+    // should only have two rows left and last should have the id of the entered term
```
david-mears-2, line 41: Why two rows? I would expect a single row, for the one matching packet. So what is the first row for?

app/e2e/packetGroupPage.local.ts
```diff
@@ -0,0 +1,73 @@
...
+    const parameterCheckBox = page.getByRole("menuitemcheckbox", { name: "a" });
+    await expect(parameterCheckBox).toBeEnabled();
+    await expect(parameterCheckBox).not.toBeChecked();
+    await parameterCheckBox.check();
+    await expect(lastRowCells).toHaveCount(3); // should have a new column for parameter "a"
+    await expect(rows.first().getByRole("cell").last()).toHaveText("a"); //header
+    await expect(lastRowCells.last()).toHaveText("1");
+    const parametersCell = (await lastRowCells.all())[1];
+    const parameterPills = await parametersCell.locator(".rounded-md").all();
+    await expect(parameterPills.length).toBe(2); // a should have been removed
```
david-mears-2, line 71: I don't see at what step 'a' is removed?

app/e2e/packetGroupPage.local.ts
david-mears-2: OK, the first row is headers. Could you add a comment to say so please?

app/e2e/packetPage.local.ts
```diff
@@ -0,0 +1,100 @@
+import { test, expect, Locator } from "@playwright/test";
+import { getContentLocator, getPacketPageAccordionSection, selectPacketPageTab } from "./utils";
+
+// Test which are only run against localhost, where we can assume we have the demo dataset packets
```
david-mears-2, line 4:
```suggestion
// Tests which are only run against localhost, where we can assume we have the demo dataset packets
```

app/e2e/packetPage.local.ts
```diff
@@ -0,0 +1,100 @@
...
+test.describe("Local packet page", () => {
+  const parametersPacketId = "20240829-185440-781be0f3";
+  const dependsPacketId = "20250120-073637-b8c1e707";
+  const artefactTypesPacketId = "20240729-155513-1432bfa7";
+
+  test("can see packet group name and packet id", async ({ page }) => {
+    await page.goto(`./parameters/${parametersPacketId}`);
+    const content = await getContentLocator(page);
+    await expect(await content.getByRole("heading", { name: "parameters", level: 2 })).toHaveText("parameters");
+    await expect(await content.getByText(parametersPacketId)).toHaveText(parametersPacketId);
```
david-mears-2, line 14:
```suggestion
    await expect(await content.getByRole("heading", { name: "parameters", level: 2 })).toBeVisible();
    await expect(await content.getByText(parametersPacketId)).toBeVisible();
```

app/e2e/packetPage.local.ts
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  test.describe("Summary", () => {
+    test("can see parameters", async ({ page }) => {
+      await page.goto(`./parameters/${parametersPacketId}`);
+      const content = await getContentLocator(page);
+      const parametersDiv = await getPacketPageAccordionSection(content, "Parameters");
```
david-mears-2, line 21: This applies a few times: since `content` is never used except to be passed into getPacketPageAccordionSection, we can save lines in the tests by passing `page` to getPacketPageAccordionSection instead, and having getPacketPageAccordionSection use getContentLocator itself.

app/e2e/packetPage.local.ts
```diff
@@ -0,0 +1,100 @@
...
+      await expect(remoteItems.first()).toHaveText("https://github.com/mrc-ide/packit.git");
+    });
+
+    test("can see platform", async () => {
+      const platformDiv = await getPacketPageAccordionSection(content, "Platform", true);
+      // get top level list items
+      const listItems = await platformDiv.locator("ul.space-y-1 > li").all();
+      await expect(listItems.length).toBe(3);
+      await expect(listItems[0]).toHaveText("OS" + "Ubuntu 22.04.4 LTS");
+      await expect(listItems[1]).toHaveText("Systemx" + "86_64, linux-gnu");
```
david-mears-2, line 89:
```suggestion
      await expect(listItems[1]).toHaveText("System" + "x86_64, linux-gnu");
```

app/e2e/packetPage.local.ts
```diff
@@ -0,0 +1,100 @@
...
+      await expect(listItems[2]).toHaveText("Language" + "R version 4.4.1 (2024-06-14)");
+    });
+
+    test("can see packages", async () => {
+      const packagesDiv = await getPacketPageAccordionSection(content, "Packages", true);
+      const listItems = await packagesDiv.locator("ul.space-y-1 > li").all();
+      await expect(listItems.length).toBe(21);
+      await expect(listItems[0]).toHaveText("askpass" + "1.2.0");
+    });
+  });
```
david-mears-2, line 99: One of us should add some e2e tests for the downloads tab.

app/e2e/packetPage.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      if ((await noneDiv.count()) > 0) {
+        await expect(noneDiv).toHaveText("This packet has no dependencies on other packets");
+      } else {
+        const depItems = await dependenciesDiv.getByRole("listitem");
+        await expect(await depItems.count()).toBeGreaterThan(0);
+        for (const li of await depItems.all()) {
+          const link = await li.getByRole("link");
+          await expect(link).toBeEnabled();
+          const linkText = await link.innerHTML();
+          expect((await link.getAttribute("href")).split("/").at(-1)).toBe(linkText);
```
david-mears-2, line 54: why is linkText expected to be part of the href? is it a packet id? if so, would be helpful to rename:
```suggestion
          const packetId = await link.innerHTML();
          expect((await link.getAttribute("href")).split("/").at(-1)).toBe(packetId);
```

app/e2e/packetPage.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      // get top level list items
+      const listItems = await gitDiv.locator("ul.space-y-1 > li").all();
+      await expect(await listItems.length).toBe(3);
+      await expect(listItems[0]).toHaveText(/^Branch/);
+      await expect(listItems[1]).toHaveText(/^Commit[\da-f]{40}$/);
+      await expect(await listItems[2].locator("span")).toHaveText("Remotes");
+      const remoteItems = await listItems[2].getByRole("listitem");
+      await expect(await remoteItems.count()).toBeGreaterThan(0);
+    });
+  });
```
david-mears-2, line 97: NB this metadata is optional, so the section may validly not exist: https://github.com/mrc-ide/orderly2/blob/main/inst/schema/outpack/metadata.json

app/e2e/packetPage.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    await expect(listItems[2]).toHaveText(/^Language/);
+  });
+
+  test("can see Packages section", async ({ page }) => {
+    const content = await getContentLocator(page);
+    await selectPacketPageTab(content, "Metadata");
+    const packagesDiv = await getPacketPageAccordionSection(content, "Packages", true);
+    // Might be empty but expect the list to exist!
+    await expect(await packagesDiv.locator("ul.space-y-1")).toBeVisible();
+  });
```
david-mears-2, line 117: NB the 'session' metadata, used here, is required only for orderly packets, not outpack packets (maybe just add this as a comment in case we never in fact have any non-orderly packet)

.gitignore
david-mears-2: Note that `auth.json` is written to the `test-results` dir, so if that is erroneously not deleted (e.g. if playwright dies mid-run), we'd be uploading the github access token for all to see. So `auth.json` might want to be written somewhere else instead.

app/playwright.config.ts
```diff
@@ -0,0 +1,83 @@
...
+    baseURL,
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: "on-first-retry",
+
+    headless: true,
+    screenshot: "only-on-failure"
+  },
+
+  /* Configure projects for major browsers */
```
david-mears-2, line 47: Well, I don't think firefox really counts as a major browser. Safari and Edge are way more popular.

https://en.wikipedia.org/wiki/Usage_share_of_web_browsers#Summary_tables

david-mears-2: I'd be happy with testing on Chrome and Safari alone.

david-mears-2 (approved):

app/e2e/index.spec.ts
EmmaLRussell: Yeah, probably a bad method name - it's not actually the display name of the packet group, but the slightly more human readable form that is sometimes used for packet group (non-display-)names and packets, which simply replaces hyphens with spaces. Could call it `getReadableIdString` or something...

app/e2e/packetGroupPage.local.ts
EmmaLRussell: This is testing adding a new column to the table for the "a" parameter, whose values should now appear in the new column and should be removed from the main "Parameters" column.

app/e2e/packetPage.local.ts
EmmaLRussell: One of us made a ticket at least :)
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6240/E2e-tests-for-downloads

app/playwright.config.ts
EmmaLRussell: sad face

app/e2e/auth.setup.ts
```diff
@@ -0,0 +1,29 @@
+import { test as setup, expect } from "@playwright/test";
+
+const authMethod = process.env.PACKIT_E2E_AUTH_METHOD || "basic";
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah, then we just need the packit instance from the user. And it'll be a bit easier if we roll the auth flow itself into the typescript.

app/playwright.config.ts
EmmaLRussell: Tags are fine but I'm wary of accidentally running tests against servers they shouldn't be run on. so I think we'd still want some sort of sanity test that the base url looks right for the test suite, probably in the suite itself. 
I think the generic vs specific tests do fall naturally into different suites so I don't mind that too much, but yeah, mutation tests will be another dimension again.

scripts/run-e2e-tests-on-server
EmmaLRussell: Only if we had a non-localhost basic auth deployment, which we don't at the moment but  we could do!

app/playwright.config.ts
EmmaLRussell: Actually, it would probably make sense to put this sanity check in auth.setup.ts since this is going to run for every test which needs to authenticate. It has access to the `testInfo` for the test, which includes its tags, so it can check if we're running a test  with the `@mutatesState` tag against non-localhost or non-dev server and bomb out if so.

app/e2e/utils.ts
EmmaLRussell: It will do. Am changing to something similar to this.

.gitignore
EmmaLRussell: It wouldn't be the github access token, it would just be the temporary packit token, which would only apply to the localhost instance we're running on CI, so I think we should be ok unless we start running tests against prod on CI.  I'll try to omit it auth.json though.

app/e2e/packetGroupPage.local.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  // we can't assume new packets won't have been created, but should at least have those from the demo set
+  test("can see packet rows", async () => {
+    await expect(await rows.count()).toBeGreaterThanOrEqual(3);
+    const oldestRow = await rows.last();
+    const firstCell = await oldestRow.getByRole("cell").first();
+    const link = await firstCell.getByRole("link");
+    await expect(link).toHaveText(packetId);
+    await expect(await link.getAttribute("href")).toBe(`/parameters/${packetId}`);
+    const expectedDate = new Date(Date.UTC(2024, 6, 29, 15, 46, 52));
```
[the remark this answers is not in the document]
EmmaLRussell: Ha, that is odd, I was trying to avoid those sorts of issues by using `toLocaleString` but maybe your browser and node have different idea of the locale!..

app/e2e/index.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    const filterInput = await page.getByPlaceholder("Filter packet groups...");
+    await filterInput.fill(firstPacketGroupName);
+    // wait for reset-filter button to become visible
+    await expect(await content.getByLabel("reset filter")).toBeVisible();
+    const filteredGroups = await content.getByRole("listitem");
+    // expect to have at least one packet group remaining, and expect all to have filter term  as a substring
+    expect(await filteredGroups.count()).toBeGreaterThan(0);
+    for (const packetGroup of await filteredGroups.all()) {
+      expect(await packetGroupNameFromListItem(packetGroup)).toContain(firstPacketGroupName);
+    }
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah maybe, just one of those great unexplained e2e oddities..

README.md
```diff
@@ -60,3 +60,46 @@ See [app/README.md](https://github.com/mrc-ide/packit/blob/main/app/README.md)
...
+
+NB If you're testing with a server url which is not a domain root, ensure that you append a "/". In this case it's also 
+important to use `./` as the base route in the Playwright tests, not `/`.
+
+The tests and test script use environment variables to set and determine test configuration values. These are:
+`PACKIT_E2E_BASE_URL`, 
+`PACKIT_E2E_AUTH_METHOD`,  
+`PACKIT_E2E_BASIC_USER`, 
+`PACKIT_E2E_BASIC_PASSWORD`, 
+`PACKIT_E2E_GITHUB_TOKEN`
```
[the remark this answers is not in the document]
EmmaLRussell: This is all working a bit differently following the refactor. 
To use a basic auth server, you can either set the user and password env vars yourself then run playwright, or use the `test:e2e` script which will set the default localhost ones for you. 
The PACKIT_E2E_AUTH_METHOD var is no longer used (the auth playwright setup works out the auth method itself by asking the API), and neither is PACKIT_E2E_GITHUB_TOKEN because authenticating with github and obtaining the token is now done in the playwright setup too. You can still set GITHUB_ACCESS_TOKEN, in which case that will be used for github auth, rather than needing to go through device flow.

README.md
```diff
@@ -60,3 +60,46 @@ See [app/README.md](https://github.com/mrc-ide/packit/blob/main/app/README.md)
...
+hatch shell 
+pip install pyorderly
+```
+
+NB If you're testing with a server url which is not a domain root, ensure that you append a "/". In this case it's also 
+important to use `./` as the base route in the Playwright tests, not `/`.
+
+The tests and test script use environment variables to set and determine test configuration values. These are:
+`PACKIT_E2E_BASE_URL`, 
+`PACKIT_E2E_AUTH_METHOD`,
```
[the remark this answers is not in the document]
EmmaLRussell: This is all a bit different now - I'll update this section of the README.

app/e2e/packetGroupPage.local.ts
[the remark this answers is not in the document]
EmmaLRussell: I'm going to make it check that you either get US or GB format, as I think those ought to cover the likely options between them.

EmmaLRussell:
**Updates following last review**

The major changes were abandoning the `run-e2e-tests-on-server` script and associated python file. Instead, all authentication happens in `auth.setup.ts` which now:

- Makes an auth config call to the server to figure out whether basic or github auth is required
- If basic, requires PACKIT_E2E_BASiC_USER and PACKIT_E2E_BASIC_PASSWORD env vars to be set, and logs in with those
- If github, checks if GITHUB_ACCESS_TOKEN env var is set. If not, does device flow to get a github token using `@octokit/auth-oauth-device`. Uses the github token to login to packit. 

The other main change was to refactor the division of "local" and generic tests, which were previously matching on file name. Now there are tags for `@demoPackets` (tests which use the local demo dataset) and `@stateMutate` (for tests which will change the state of the server by running packets or changing permissions. We don't have any of these yet, but I've added a placeholder test which has this tag). We can include or exclude these tags when running playwright - test runs on dev or prod will probably fail if `@demoPackets` is included. `@stateMutate` should never be included when running against prod. I've added a fixture, `tagCheckFixture` which all tests should use. This checks if we're running tests with unexpected tags given the baseURL, and prints warning messages or throws an error (in the case of `@stateMutate` tag on non-local or non-dev). 

There are now three npm scripts in packit json:
- `test:e2e` which allows playwright to use default localhost server, and sets basic username and password
- `test:e2e-dev` which runs against https://packit-dev.dide.ic.ac.uk/reside-dev/ and excludes the two tags
- `test:e2e-prod` which runs against https://packit.dide.ic.ac.uk/reside/ and excludes the two tags

I was going to include functionality to allow user to enter basic username and password interactively if not provided via environment variables, but stdin doesn't seem to be available in the auth.setup script (although I could get it to read input if I made a universal global setup). I presume something in Playwright is tampering with stdin. If it's useful, I can add that, in global setup, or in a separate node script but it's probably not worth it when we can just ask the user to set the variables!

Various other changes from the review comments. Some detailed in replies, and also:
- Replaced Firefox with Safari, since people insist on using it..
- renamed workflow file and some of the job names
- upload test-results, should delete auth.json (but not a big deal if it doesn't)


**Testing**

The dev and prod scripts script are currently failing because the deployed main branch doesn't include the breadcrumb test id. 

To test running non-interactively with github login, run the `test:e2e-dev` script with env var GITHUB_ACCESS_TOKEN set to your PAT - the tests should run successfully without needing to authenticate interactively: `export GITHUB_ACCESS_TOKEN=[your PAT] && npm run test:e2e-dev`

You can test triggering the `tagCheckFixture` by running without tag exclusion:
`export PACKIT_E2E_BASE_URL=https://packit-dev.dide.ic.ac.uk/reside-dev/ && npx playwright test` or `export PACKIT_E2E_BASE_URL=https://packit.dide.ic.ac.uk/reside/ && npx playwright test`

david-mears-2:
It looks like you can make test fixtures '[automatic](https://playwright.dev/docs/test-fixtures#automatic-fixtures)' so that we don't need to rely on future developers remembering to ensure their import of `test` comes from the right place!

> Automatic fixtures are set up for each test/worker, even when the test does not list them directly.

app/e2e/auth.setup.ts
```diff
@@ -0,0 +1,93 @@
+import { test as setup, expect, Page } from "@playwright/test";
```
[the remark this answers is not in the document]
david-mears-2: I am uncomfortable relying on developers to remember to check that they are importing 'test' from the correct place. Especially since IDEs often write imports for you. We could do something like the following, in a global setup script akin to auth.setup.js that is listed as a dependency for all test projects in playwright config:

```js
function checkImports(filePath) {
  const content = fs.readFileSync(filePath, 'utf-8');
  const importRegex = /import\s+{[^}]*\btest\b[^}]*}\s+from\s+".\/tagCheckFixture";/;
  if (!importRegex.test(content)) {
    throw new Error(`Test file ${filePath} does not import 'test' from 'tagCheckFixture'.`);
  }
}
```

app/e2e/packetPage.demo.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+// Tests which are only run against localhost, where we can assume we have the demo dataset packets
+test.describe("Demo packet page", { tag: TAG_DEMO_PACKETS }, () => {
+  const parametersPacketId = "20240829-185440-781be0f3";
+  const dependsPacketId = "20250120-073637-b8c1e707";
+  const artefactTypesPacketId = "20240729-155513-1432bfa7";
+
+  test("can see packet group name and packet id", async ({ page }) => {
+    await page.goto(`./parameters/${parametersPacketId}`);
+    const content = await getContentLocator(page);
+    await expect(await content.getByRole("heading", { name: "parameters", level: 2 })).toHaveText("parameters");
```
david-mears-2, line 19:
```suggestion
    await expect(await content.getByRole("heading", { name: "parameters", level: 2 })).toBeVisible();
```

app/playwright.config.ts
```diff
@@ -0,0 +1,91 @@
...
+const outputDir = "test-results";
+const authStorageStateFile = path.join(__dirname, outputDir, `auth.setup.ts-authenticate-setup/auth.json`);
+let baseURL = process.env.PACKIT_E2E_BASE_URL || "http://localhost:3000";
+if (!baseURL.endsWith("/")) {
+  baseURL = `${baseURL}/`;
+}
+const testMatch = /.*\.spec\.ts/;
+export default defineConfig({
+  testDir: "./e2e",
+  testMatch,
```
david-mears-2, line 22:
```suggestion
export default defineConfig({
  testDir: "./e2e",
  testMatch: /.*\.spec\.ts/,
```

david-mears-2 (commented):
I haven't re-read all of the test files themselves, only the surrounding infrastructure things. My main comment is to suggest a further bit of defence-in-depth to make triply sure we won't be mutating state on prod.

app/e2e/auth.setup.ts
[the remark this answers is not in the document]
EmmaLRussell: Don't think it really matters for the auth setup test but yeah, for the purposes of setting a good example let's do that!

`checkImports` is a great idea. Actually, I'd be inclined to do that in auth.setup itself (possibly with a rename to reflect more generic concerns) as further defence, since people could still forget to include checkImports.setup.ts as a dependency, but their tests aren't going to get anywhere if they don't include the dep that does auth...

EmmaLRussell:
I've implemented the suggested `checkImports` method in a `globalSetup` script rather than within the auth setup (it runs first before any other project). It uses a generic glob search for test files under the test root dir, then loads all the file contents and checks for the suggested regex match of loading `test` from `tagCheckFixture`. 

The fs glob search is new and experimental in node 22 - have updated the node version in CI. 

I've also updated the dev and prod urls now they're on proper subdomains, and also made sure the BASIC_AUTH variable is set in the dev-start script (it caused an error if running in github mode).

app/e2e/packetGroupPage.demo.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  // we can't assume new packets won't have been created, but should at least have those from the demo set
+  test("can see packet rows", async ({ baseURL }) => {
+    await expect(await rows.count()).toBeGreaterThanOrEqual(3);
+    const oldestRow = await rows.last();
+    const firstCell = await oldestRow.getByRole("cell").first();
+    const link = await firstCell.getByRole("link");
+    await expect(link).toHaveText(packetId);
+    const expectedHref = getInstanceRelativePath(baseURL, `parameters/${packetId}`);
+    await expect(await link.getAttribute("href")).toBe(expectedHref);
+    // We expect servers we're testing on to have either GB or US locale.
```
plietar, line 25:
```suggestion
    // We expect browsers we're testing on to have either GB or US locale.
```

No? I'd be surprised if we did any date rendering server side. The JS should pick up the browser defaults.

app/e2e/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  return section;
+};
+
+export const selectPacketPageTab = async (content: Locator, tab: string) => {
+  const nav = await content.getByRole("navigation");
+  await nav.getByRole("link", { name: tab }).click();
+};
+
+// Get a relative path which includes the packit instance, if any
+export const getInstanceRelativePath = (baseURL: string, path: string) => {
```
plietar, line 74: So as of last Friday we're now running all instances at the root path, meaning we could actually remove this function.

README.md
```diff
@@ -60,3 +62,62 @@ See [app/README.md](https://github.com/mrc-ide/packit/blob/main/app/README.md)
 ## Authentication in Packit
 
 See [docs/auth.md](docs/auth.md)
+
+## e2e Tests
+
+Playwright tests are in `./app/e2e`. There are three npm scripts defined in `app/package.json`:
+- `test:e2e` - tests against `http://localhost:3000`. Can use either github or basic auth (assumes default superuser credentials).
+- `test:e2e-dev` - tests against `https://packit-dev.dide.ic.ac.uk/reside-dev/` excluding demo dataset and state mutating tests
```
plietar, line 70:
```suggestion
- `test:e2e-dev` - tests against `https://reside-dev.packit-dev.dide.ic.ac.uk/` excluding demo dataset and state mutating tests
```

app/e2e/tagCheckFixture.ts
```diff
@@ -0,0 +1,52 @@
+import { test as base } from "@playwright/test";
+
+export const TAG_STATE_MUTATE = "@stateMutate";
+export const TAG_DEMO_PACKETS = "@demoPackets";
+
+const isLocal = (baseURL: string) => baseURL.startsWith("http://localhost");
+const isDev = (baseURL: string) => baseURL.startsWith("https://packit-dev.dide.ic.ac.uk");
```
plietar, line 7: I think this is broken now that the URL has changed. You would need to parse the URL, extract the domain name and check if that part ends in `packit-dev.dide.ic.ac.uk`

plietar (approved):
LGTM, just some minor bits

app/e2e/tagCheckFixture.ts
plietar: Something like
```suggestion
const isDev = (baseURL: string) => new URL(baseURL).hostname.endsWith("https://packit-dev.dide.ic.ac.uk");
```

app/e2e/packetGroupPage.demo.spec.ts
EmmaLRussell: Actually you can set the locale in the playwright config, that should make this more reliable..

app/e2e/utils.ts
EmmaLRussell: Might just leave it in for now just in case we run any old style multi-tenancies again!

plietar (approved):

david-mears-2 (approved):
Thank you!

Perhaps the version of "@types/node" (package.json) wants to be kept in sync with the node version - it's still on 16.

EmmaLRussell:
> Sorry, just noticed app/Dockerfile has node 18 still as well. Also line 6 of app/README.md cites Node 18.

:scream: So many node versions! Great spot!

david-mears-2 (approved):

PR #180, opened 2025-02-21
Mrc-6226 Sort display of parameters on packetgroup page

absternator:
display the parameters in the parameters column sorted by the parameter key.
Can be seen in screen shot below (sorted a , b, c)
image


plietar (approved):
Thanks!

david-mears-2 (approved):

PR #182, opened 2025-02-21
Implement the sd_notify protocol.

plietar:
systemd is the standard service manager used on Linux systems, and theone we use on our production machines.

It is able to track the state of services, including whether a service is "starting" or "started" already. For most services that are quick to start, the difference is minimal but in the case of Packit it can take a handful of seconds or more for the service to be ready to accept connections.

Thankfully systemd provides an easy way to notify it of the service's state. If the service is configured with `Type=notify`, systemd will set a `NOTIFY_SOCKET` environment variable point to a Unix socket. The service can communicate with systemd over that socket, and in particular it can write `READY=1` to change the service's state from "starting" to
"started". See [the systemd documentation][sd_notify] for more details about the protocol and the various messages it supports.

Having systemd know accurately when the service has started allows us to coordinate startup more easily (for example we could execute some provisioning steps automatically once the service has started). It can also make whole-system integration tests easier to write, by providing an easy way to start the service and wait for it to be ready, without having to repeatedly poll the HTTP endpoint.

Testing this change outside of the context of a systemd service is a little tricky. The best way to reproduce this is to run packit as a transient unit with `systemd-run`:

```sh
# Build the app's JAR file
gradle :app:bootJar

# Start Packit as a transient unit, telling systemd we support the
# notify protocol
systemd-run --user --collect --service-type=notify --same-dir \
  --unit=packit -- java -jar app/build/libs/app.jar

# Watch the logs for updates from systemd about the state of the
# service.
journalctl --user --unit packit.service

# Cleanup
systemctl --user stop packit.service
```

The logs show systemd updating the status of the `packit.service` unit
just after Spring has completed initialization.

```
14:35:08 systemd[2970]: Starting packit.service
14:35:09 java[276126]:   .   ____          _            __ _ _
14:35:09 java[276126]:  /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
14:35:09 java[276126]: ( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
14:35:09 java[276126]:  \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
14:35:09 java[276126]:   '  |____| .__|_| |_|_| |_\__, | / / / /
14:35:09 java[276126]:  =========|_|==============|___/=/_/_/_/
14:35:09 java[276126]:  :: Spring Boot ::                (v3.0.2)
[...] Lots more output from spring
14:35:14 java[276126]: Tomcat started on port(s): 8080 (http) with context path ''
14:35:14 java[276126]: Started AppKt in 5.372 seconds (process running for 5.771)
14:35:14 systemd[2970]: Started packit.service
```

Amazingly, systemd has actually added an option just yesterday that makes it easier to start programs that support sd_notify, without needing to create a transient service: https://github.com/systemd/systemd/commit/4389e4c2aea36224e99dae616aff08c8560c94e8
Unfortunately, it will be a little while until we can actually make use of this.

[sd_notify]: https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html


david-mears-2:
This seems useful.

Am I understanding right: this PR introduces the ability to notify systemd that the api has started, but does not make use of the notification yet?

plietar:
# Add a 'wait-notify' script.

This implements the supervisor-side of the sd_notify protocol.

It can be used either with local processes, or as a wrapper around `docker run`:

```sh
./scripts/wait-notify -- java -jar api/app/build/libs/app.jar
./scripts/wait-notify --docker-run -- mrcide/packit-api
```

This has been integrated in the scripts responsible for starting the API container, removing the need to poll and wait for the API to become ready.

Ideally we would also use this to start the database container, and remove the `wait-for-db` script. Unfortunately the Postgres image has some quirk that makes it incompatible, despite Postgres natively supporting sd_notify. The first time the container starts, it runs a temporary server used to initialiase the database storage. This temporary server is not accesible, but because it inherits the NOTIFY_SOCKET environment variable, it notifies the caller and the container appears ready too early. This can be fixed upstream in the postgres image.

plietar:
> Am I understanding right: this PR introduces the ability to notify systemd that the api has started, but does not make use of the notification yet?

Yeah, originally my intention was to not use it anywhere here, but use it in `packit-infra`, by [changing the service type from `simple` to `notify`](https://github.com/reside-ic/packit-infra/blob/85153e79dc47b4b376b3502e57a35e94a80cf45c/modules/packit-api.nix#L186). This tells systemd that our service supports the sd_notify protocol, and to use it to determine readiness.

Following your suggestion I've actually started using the feature in this repository too. There is a wait-notify script that implements the other side of the protocol. It's being used to start the Packit API container now, and in the future it can be used to start the database too (I just need to make a small upstream PR first).

plietar:
Some dummy uses of the new `wait-ready` script:
```sh
# systemd-notify --ready finds the notify socket and write READY=1
$ ./scripts/wait-notify -- systemd-notify --ready
Program is ready and running as PID 852675!

$ ./scripts/wait-notify -- bash -c 'echo "Init"; sleep 5; systemd-notify --ready; sleep 5; echo "After"'
Init
Program is ready and running as PID 852921!
$ # Command returns after five seconds.
After # This is printed by the background task 5 seconds later

$ ./scripts/wait-notify -- true
Program terminated with status 0 before becoming ready.
$ ./scripts/wait-notify -- false
Program terminated with status 1 before becoming ready.

$ ./scripts/wait-notify --docker-run ubuntu true
Program terminated with status 0 before becoming ready.
$ ./scripts/wait-notify --docker-run ubuntu false
Program terminated with status 1 before becoming ready.

# I couldn't find any image with systemd-notify included, but socat can be used to write to the socket instead.
$ ./scripts/wait-notify --docker-run -- --entrypoint sh alpine/socat -c 'echo "READY=1" | socat - UNIX-CLIENT:$NOTIFY_SOCKET'
Container is ready!
```

plietar:
The change to the postgres image was merged upstream but I don't really understand how and when it gets published to Docker Hub. In the meantime I'm happy to get this reviewed and merged, I'll be making a separate PR for remove the `wait-db` script later.

scripts/wait-notify
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
```
david-mears-2, line 1: My laptop doesn't recognise 'python', only 'python3'. Should that imply a change on my end (to alias python3, maybe) or a change here?

david-mears-2:
> It can be used either with local processes, or as a wrapper around docker run:
> ./scripts/wait-notify -- java -jar api/app/build/libs/app.jar
> ./scripts/wait-notify --docker-run -- mrcide/packit-api

I think these example usages would be very helpful to add in either some .md or a comment in the script.

scripts/wait-notify
plietar: Oh wow I didn't realize Ubuntu/Debian had done that. I'll change it to python3. Thanks!

plietar:
> I think these example usages would be very helpful to add in either some .md or a comment in the script.

Good shout. I've added them in the script.

david-mears-2 (approved):
By the way, what's the reason for using a script that says the intended language on line 1, rather than creating a `wait-notify.py` script?

plietar:
The first line is called a shebang. If the file is marked executable but is not recognised as a binary file, when you run `./scripts/wait-notify` the kernel looks for the first line and if it starts with `#!` it will run the rest of the line as an “interpreter” and pass the path to the file as an argument.

The path to the interpreter you specify must be absolute, which means `#!python3` would not work. While it is possible to just use `#!/usr/bin/python3`, that would assume a particular location for the Python binary. On a strict FHS type system that would probably work, but if you have something a bit exotic like virtualenv, easybuild, pyenv or nix, the Python executable you want to use might be elsewhere. Additionally Linux distributions used to distinguish between `/bin` and `/usr/bin`, so you would have to guess in which directory your interpreter was installed (realistically Python would also be `/usr/bin`, but there’s a good chance `sh` was only found in `/bin`) - these days however all mainstream distributions have merged the two directories and `/bin` is a symbolic link to `/usr/bin`.

The custom is therefore to use `#!/usr/bin/env python3`. The env program will search for a python3 binary in every directory listed in your PATH environment variable. Even the most exotic Linux distributions have an executable in `/usr/bin/env` to maintain this interface (on NixOS that is literally the only file that exists within `/usr`).

From the kernel’s perspective, the file extension is irrelevant and is never used. Whether you use an extension or not is a matter of personal preference. My take on it is that here Python is an implementation detail of the script. The caller of the script shouldn’t need to know what language the script is written in, therefore the extension is not needed. Files that are not standalone executables but instead are supposed to be imported should have an extension.

scripts/wait-notify
plietar: For some context on what I meant - back in the Python 2 days, there was only `python`. Then distributions started shipping both Python 2 and Python 3, as the `python` and `python3` executables respectively (and I think some had `python2` if you wanted to be explicit). Eventually they stopped shipping Python 2 and removed the `python` binary.

Some distributions (off the top of my head at least Arch and NixOS) used that opportunity to now make `python` an alias to `python3`. I did not realise that Debian / Ubuntu never made that alias, and therefore `python` may not exist on these systems.

PR #184, opened 2025-03-20
mrc-5140 Support Montagu redirect

EmmaLRussell:
Because the PUBLIC_URL and REACT_APP_PACKIT_URL environment variables are baked into the docker image on build, I've made a separate montagu packit image with the correct `/packit` paths set. This isn't an issue for the new Nix installation because it doesn't use docker and builds the packit app itself with required environment. 

The alternative to this, suggested by Paul, would be to have a top level domain for packit which is separate from the montagu domain, and which montagu proxy is configured to redirect login requests from. It would also mean modifying and configuring the auth logic on the montagu homepage (or use paths there which nginx is configured to redirect).. However, I think this wouldn't work because we save packit user details into local storage, which woudn't be accessible cross-domain.


EmmaLRussell:
Closing as superseded by https://github.com/mrc-ide/packit/pull/187

PR #185, opened 2025-03-26
Mrc-6362 Remove Published

absternator:
There is a published column in packets table since inception. This is redundant and since the way we will implement "published" is via permission manipulation we can safely remove this column. The rest of the changes is remove published from models


EmmaLRussell (approved):
LGTM

PR #187, opened 2025-03-27
Mrc 5140 Preauthentication with trusted headers

EmmaLRussell:
This branch adds support for Montagu authentication in Packit, implemented as a new auth method called "preauth" i.e. preauthenticated. This is also known as "trusted headers" as a login endpoint is implemented which accepts user details in its headers, and trusted that these headers come from a reputable source, and logs the user in (provides a standard packit source). Hence, if this auth method is used, for Montagu or anything else, this endpoint MUST be locked down to external access in the proxy. 

There are also some front end changes to redirect to the server for login and logout if auth method is preauth - we need to go via nginx to get to the Montagu login page in this case, rather than staying in packit front end app using the default React routing.

## Core changes
Backend:
- New endpoint `/login/preauth` defined in LoginController. Throws exception if auth type is not `preauth`. Uses `PreAuthenticatedLoginService` to save user details and issue packit token. 
- `PreAuthenticatedLoginService` - calls `UserService` to save user details and `JwtIssuer` to generate a token
- `UserService` - refactoring to support saving user details for external user originating from both github and preauth logins. The Preauth user will be created in the database if it doesn't already exist - this allows us to pre-populate the user table with users and the permissions which they should have (potentially migrating from OrderlyWeb). 
- add `preauth` auth method to AppConfig. 

Frontend:
- Login form redirects to the server login route
- ProtectedRoute redirects to the server on logout - in this case we pass a "loggingOut" flag so that Montagu knows that it has to log out too (we need to keep Montagu and Packit login state synchronised). 

## Other changes
- new gha workflow, `build_montagu_packit_front_end` and related script `build-and-push-montgu` which builds a front end image which bakes in the front end and back end routes we expect to use for packit under the top level montagu domain (we may change this to use a subdomain in a future ticket). Related Dockerfile changes to pick up these build args. 
- Updated `/api/scripts/run` to support specifying any auth method in arg (not just basic auth), and added optional `--preauth` arg to `scripts/dev-start`
- removed a `navigate("/login")` from `AccountHeaderDropdown` as this navigate was being doubled up by `ProtectedRoute` which has a `useEffect` to notice when logout happens (indicated by user info being cleared from state) and does the necessary navigate to the "login" react page or, now, to the externally configured logout route via the server. The additional login navigate was causing problems in the proxy as it was not passing the 'loggingOut' flag. 
- added a `create-preauth-user` script to the db container which takes user details and role name. This allows the montagu proxy repo to test running with a user who has admin permissions (and can actually see some packets!).  
- I've spruced up the auth doc a little, which was very out of date - but only bits pertaining to config.
- For reasons I cannot fathom, the gh actions stopped running for this branch until I targeted it specifically. No idea why, I'll revert before merging. 

## Testing
To test the new endpoint manually:

Start everything up in preauth mode:
```
 ./scripts/dev-start --preauth
```

Make a preauth login request:
```
curl -H "X-Remote-User: preauth.user" -H "X-Remote-Name: Pre-Auth User" -H "X-Remote-Email: p.a.user@example.com" http://localhost:8080/auth/login/preauth 
```

Check that the user has been created:
```
docker exec packit-db psql -d postgresql://packituser:changeme@localhost:5432/packit -c "SELECT * FROM public.user;"
```
NB It's important to user the public. prefix to the user table here, otherwise you might get results from the system user table. 

To test that everything works manually from the front end, you should use the related montagu proxy branch which includes this packit branch's containers:  https://github.com/vimc/montagu-proxy/pull/97
You should be able to:
- login in Montagu, then click "Packit" to get to Packit in a logged in state
- log out from Packit or from Montgu 
- get redirected to Montagu login if you try to browse directed to Packit when not logged in, then get directed to packit on login

For more context on Montagu for Packit auth, see the [epic](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4753/Montagu-Auth)


david-mears-2:
Are we intending this hole to have been patched yet:

1. Run montagu-proxy as per https://github.com/vimc/montagu-proxy/pull/97
2. Log in to montagu
3. Go to the reporting portal (not Packit, i.e., orderlyweb)
4. Log out of orderlyweb and be redirected to montagu, apparently in a logged-out state (clicking 'Reporting Portal' will now redirect you back to montagu)
5. Click on Packit - lo and behold, you are now logged in to Packit!

So it seems there's a state where Montagu believes you are logged out, but Packit hasn't got the message and still lets you in.

api/app/src/test/kotlin/packit/integration/controllers/LoginControllerTest.kt
```diff
@@ -61,12 +61,91 @@ class LoginControllerTestGithub : IntegrationTest()
...
+        userRepository.save(testUser)
+
+        val result =
+            packit.integration.controllers.LoginTestHelper.getPreauthLoginResponse(
+                userName, userDisplayName, userEmail, restTemplate
+            )
+
+        assertSuccess(result)
+        val packitToken = jacksonObjectMapper().readTree(result.body).get("token").asText()
+        assertThat(packitToken.count()).isGreaterThan(0)
```
david-mears-2, line 137: Did you want to assert that the last-logged-in field was updated?

api/app/src/test/kotlin/packit/integration/controllers/LoginControllerTest.kt
```diff
@@ -61,12 +61,91 @@ class LoginControllerTestGithub : IntegrationTest()
...
+    }
+
+    @Test
+    fun `returns 400 when username header not provided`()
+    {
+        val result =
+            packit.integration.controllers.LoginTestHelper.getPreauthLoginResponse(
+                null, userDisplayName, userEmail, restTemplate
+            )
+        assertEquals(result.statusCode, HttpStatus.BAD_REQUEST)
```
david-mears-2, line 147:
```suggestion
        assertEquals(result.statusCode, HttpStatus.BAD_REQUEST)
        assertThat(packitToken.count()).isEqualTo(0)
```

app/Dockerfile
```diff
@@ -1,12 +1,20 @@
...
 COPY package.json /app/package.json
 COPY package-lock.json /app/package-lock.json
 WORKDIR /app
 RUN npm ci
 
 COPY . /app
 
+# override env var in production file if set
+RUN if [ -n "$REACT_APP_PACKIT_API_URL" ]; then  echo "REACT_APP_PACKIT_API_URL=$REACT_APP_PACKIT_API_URL" >> env.production.local;  fi
+
```
david-mears-2, line 17: I'm not sure why this only applies to production. It makes me uncomfortable to introduce differences between production and other environments, hence why I ask.

app/scripts/build-and-push-montagu
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
david-mears-2, line 1: I'm not sure what this script is for. Why does packit contain a script that builds montagu?

app/src/app/components/routes/ProtectedRoute.tsx
```diff
@@ -13,6 +14,17 @@ export const ProtectedRoute = () => {
   const { pathname } = useLocation();
   const expiryMessage = "You have been signed out because your session expired. Please log in.";
 
+  const navigateToLogin = (qs = "") => {
+    // navigate to logged out screen, either using react routing or external navigation, depending on whether we're
+    // using pre-auth, which may require logout action from auth provider
+    if (authConfig?.enablePreAuthLogin) {
+      const href = `${process.env.PUBLIC_URL}/logout`;
```
david-mears-2, line 21: Is `${process.env.PUBLIC_URL}/logout` an endpoint in montagu? Or where is the logic that logs out the montagu session when packit logs out?

david-mears-2: I was expecting to see something here relating to 

> in this case we pass a "loggingOut" flag so that Montagu knows that it has to log out too

but I don't see how/where we pass this to Montagu.

app/scripts/build-and-push-montagu
EmmaLRussell: Relates to this bit of the PR description:
"new gha workflow, `build_montagu_packit_front_end`. and related script `build-and-push-montagu` which builds a front end image which bakes in the front end and back end routes we expect to use for packit under the top level montagu domain (we may change this to use a subdomain in a future ticket). Related Dockerfile changes to pick up these build args."

This is an unfortunate consequence of assuming that we'd be using the nix deployer from here on out, which builds the front end on deploy rather than using a pre-built docker image..

app/Dockerfile
EmmaLRussell: It's "production" rather than "dev", so production npm build rather than say "production" vs "uat". This is forcing npm to override an environment variable used at build time (when it builds the docker image) - so it's the same issue with needing to build an image which will work for the Montagu deploy to get the right API url.

api/app/src/main/kotlin/packit/controllers/LoginController.kt
```diff
@@ -38,6 +40,25 @@ class LoginController(
...
+    {
+        if (!config.authEnablePreAuthLogin)
+        {
+            throw PackitException("preauthLoginDisabled", HttpStatus.FORBIDDEN)
+        }
+
+        val token = preAuthenticatedLoginService.saveUserAndIssueToken(username, name, email)
+        return ResponseEntity.ok(token)
+    }
+
```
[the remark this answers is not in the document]
EmmaLRussell: I think it would be a fair amount of pain, because that interacts with the GitHub configuration you need to do to redirect back to packit after github authenticates the user. Could be one for another ticket though potentially.

app/src/app/components/routes/ProtectedRoute.tsx
EmmaLRussell: Ah yes, sorry, I put that into the nginx config in the proxy and didn't update the description here. I thought it would be better to have a generic `logout` route from here, and then let the preauthenticating application deal with it appropriately - here by redirecting with "loggingOut=1":
line 155 in nginx.montagu.conf:
https://github.com/vimc/montagu-proxy/pull/97/files#diff-76948d98bd1b1f9548f44ff60e6d989cc95a8aeafe5f52555b2aa7dbf4636059R155

EmmaLRussell:
> So it seems there's a state where Montagu believes you are logged out (and the user does too), but Packit hasn't got the message and still lets you in.

Oops, that should be working! Will investigate.

app/src/app/components/routes/ProtectedRoute.tsx
david-mears-2: Would be good to link there from a comment on this code, so others can follow the breadcrumbs in the future.

app/scripts/build-and-push-montagu
EmmaLRussell: It's run as part of CI, in the `build_montagu_packit_front_end` workflow, which just builds and pushes a slightly different docker image with the required env vars baked in, just for use by montagu. Other deployments don't need to do that because they use the nix deployment, which builds the front end from scratch at the point of deployment. 

It's all a bit unfortunate. Let's chat about it on Monday if still not clear.

api/app/src/main/kotlin/packit/service/PreAuthenticatedLoginService.kt
```diff
@@ -0,0 +1,26 @@
...
+import packit.exceptions.PackitException
+import packit.security.provider.JwtIssuer
+
+@Component
+class PreAuthenticatedLoginService(
+    val jwtIssuer: JwtIssuer,
+    val userService: UserService,
+)
+{
+    fun saveUserAndIssueToken(username: String, name: String, email: String): Map<String, String>
```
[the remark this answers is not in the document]
EmmaLRussell: Yes, the only one we really need should be username header I think - will check!

app/scripts/build-and-push-montagu
david-mears-2: Does this script build the whole of montagu, or just a montagu-flavoured version of packit?

api/app/src/main/kotlin/packit/service/PreAuthenticatedLoginService.kt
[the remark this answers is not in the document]
EmmaLRussell: OK, contrary to what I said in standup, this should be fine. I just hadn't cleaned up everything properly in integration test.

app/scripts/build-and-push-montagu
EmmaLRussell: The latter! I will rename..

docs/auth.md
```diff
@@ -1,14 +1,17 @@
 ## Authentication in Packit
 
 Packit uses [Spring Security](https://spring.io/projects/spring-security) for authentication. For logging in through the front end web app, we currently 
-support GitHub authentication using OAuth2 and a partial demo Basic auth method (to demonstrate supporting multiple auth types). We
-will eventually replace the Basic auth support with Montagu auth. We also support authenticated API access, currently
-with GitHub auth only. 
+support:
+- GitHub authentication using OAuth2
+- Basic auth with users whose details are stored in packit-db
+- Pre-authentication with trusted headers (used for authentication from Montagu - see [Montagu proxy](https://github.com/vimc/montagu-proxy) for more details).
```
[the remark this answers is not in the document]
EmmaLRussell: Done, and have also updated the root README for dev start option.

api/app/src/test/kotlin/packit/unit/service/PreAuthenticatedLoginServiceTest.kt
```diff
@@ -0,0 +1,47 @@
...
+            on { getUserPrincipal(mockUser) } doReturn mockUserPrincipal
+        }
+        val token = "test-token"
+        val mockjwtIssuer = mock<JwtIssuer> {
+            on { issue(mockUserPrincipal) } doReturn token
+        }
+
+        val sut = PreAuthenticatedLoginService(mockjwtIssuer, mockUserService)
+        val result = sut.saveUserAndIssueToken("test.user", "Test User", "test.user@example.com")
+        assertEquals(result, mapOf("token" to token))
```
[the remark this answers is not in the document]
EmmaLRussell: This is a unit test so those things aren't being done directly by the class being tested here. I think from the way the mocks are set up we're testing that the interactions in the class are happening as expected - the `mockJwtIssuer` will only return `token` when passed `mockUserPrincipal` as a parameter, the `mockUserService` will only return `mockUserPrincipal` from `getUserPrincipal` when passed `mockUser` as a parameter, and will only return `mockUser` from `savePreAuthenticatedUser` when passed the expected user details parameters...

api/app/src/test/kotlin/packit/integration/controllers/LoginControllerTest.kt
EmmaLRussell: It won't be an empty list, but I'll test that a token node isn't in the response body when there's an error.

app/src/app/components/routes/ProtectedRoute.tsx
```diff
@@ -13,6 +14,18 @@ export const ProtectedRoute = () => {
   const { pathname } = useLocation();
   const expiryMessage = "You have been signed out because your session expired. Please log in.";
 
+  const navigateToLogin = (qs = "") => {
+    // navigate to logged out screen, either using react routing or external navigation, depending on whether we're
+    // using pre-auth, which may require logout action from auth provider. (See /packit/logout route in montagu nginx
+    // conf for how this works with Montagu: https://github.com/vimc/montagu-proxy/blob/master/nginx.montagu.conf)
```
EmmaLRussell, line 20: This link will be valid to the description once the related proxy branch is merged..

EmmaLRussell:
> > So it seems there's a state where Montagu believes you are logged out (and the user does too), but Packit hasn't got the message and still lets you in.
> 
> Oops, that should be working! Will investigate.

This was a bug in the Montagu proxy branch, should now be working.

api/README.md
```diff
@@ -14,17 +14,20 @@ to start database and `outpack_server` instances.
...
+- **github**: The default dev configuration enables Github auth, but you will need to set environment variables PACKIT_GITHUB_CLIENT_ID
    and
    PACKIT_GITHUB_CLIENT_SECRET with the details of an OAuth app which you can use to log in. You can use the details held in
    the
-   mrc vault at `VAULT:secret/auth/githubclient/id:value` and `VAULT:secret/auth/githubclient/secret:value`.
-2. The other configuration is to use basic auth. This is enabled by setting `auth.method=basic` in `application.properties`.
-3. No auth mode. This is enabled by setting `auth.enabled=false` in `application.properties`.
-   Run up app on command line from the /api directory
-   `./scripts/run-dev`
+   mrc vault at `VAULT:secret/auth/githubclient/id:value` and `VAULT:secret/auth/githubclient/secret:value`.
```
david-mears-2, line 23: I think this should be updated to say

```
You can use the details held in the mrc vault at `secret/packit/githubauth/auth/githubclient`.
```

since the paths listed don't seem to exist.

david-mears-2: Or we can encourage people to use some `dev` details in case the ones I just referred to are tied to a non-dev app - seems risky to mix up prod and dev secrets.

Here are some dev details:

`vault read secret/packit/oauth/dev`

^ but maybe this 'oauth' folder is liable to be deleted at some point given we do not support oauth.

api/README.md
```diff
@@ -14,17 +14,20 @@ to start database and `outpack_server` instances.
...
    the
-   mrc vault at `VAULT:secret/auth/githubclient/id:value` and `VAULT:secret/auth/githubclient/secret:value`.
-2. The other configuration is to use basic auth. This is enabled by setting `auth.method=basic` in `application.properties`.
-3. No auth mode. This is enabled by setting `auth.enabled=false` in `application.properties`.
-   Run up app on command line from the /api directory
-   `./scripts/run-dev`
+   mrc vault at `VAULT:secret/auth/githubclient/id:value` and `VAULT:secret/auth/githubclient/secret:value`. 
+- **basic**: In basic auth, Packit manages user passwords. This is enabled by setting `auth.method=basic` in `application.properties`.
+- **preauth**: Also known as trusted headers, with this method, packit trusts that user details provided in headers are correct.
+             Must be used with an appropriate auth provider and proxy configuration - this is currently expected be Montagu.
```
david-mears-2, line 26:
```suggestion
- **preauth**: Also known as 'trusted headers', with this method, Packit trusts that user details provided in headers are correct.
             Must be used with an appropriate auth provider and proxy configuration - this is currently expected to be Montagu. 
```

api/app/src/test/kotlin/packit/unit/service/PreAuthenticatedLoginServiceTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class PreAuthenticatedLoginServiceTest
+{
+    @Test
+    fun `can save user and issue token`()
+    {
+        val mockUser = mock<User>()
+        val mockUserPrincipal = UserPrincipal("test.uesr", "Test User", mutableListOf(), mutableMapOf())
```
david-mears-2, line 22: This isn't actually a mock. Probably `mock<UserPrincipal>()` is preferable, since these constructor args aren't used by the test (as you can tell from the misspelling `test.uesr`).

api/app/src/test/kotlin/packit/unit/service/PreAuthenticatedLoginServiceTest.kt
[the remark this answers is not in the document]
david-mears-2: I see, this chain of mocks does indeed rely on `savePreAuthenticatedUser` having been called.

It seems like this unit test is very tied to the current implementation of the subject! Every single call and argument the subject does has been mocked out, making the unit test brittle to any change of implementation details.

I guess the trade-off is: the less you mock out, the more likely the test will be broken by changes in external entities (e.g. if `JwtIssuer` stops working, it would break this test too); while the more you mock out, the more the unit test is reduced to being a duplicate of the precise implementation details, and has to be re-written every time the implementation details change.

But maybe sometimes asserting the implementation details is exactly what one is trying to do? Here, you assert that the `JwtIssuer` and `UserService` interfaces are used in a certain way, and that might have value. (Not sure how much value!) I would probably vote to omit this unit test unless there is some articulable value to it.

Anyway. More importantly than testing the use of those interfaces, the main thing I want the test suite to give me confidence in is that those classes, here mocked out for unit testing, will do what they're meant to when used for real - namely they should save/update the user, and issue a token based on that user. That's probably the job of the controller integration test. I see the integration tests do assert that the user is in the db, _but they don't assert anything about the claims in the token_. Do we have any tests anywhere that confirm that tokens correctly contain the user's actual permissions?

app/src/app/components/routes/ProtectedRoute.tsx
david-mears-2: Thanks!

david-mears-2:
> > > So it seems there's a state where Montagu believes you are logged out (and the user does too), but Packit hasn't got the message and still lets you in.
> > 
> > 
> > Oops, that should be working! Will investigate.
> 
> This was a bug in the Montagu proxy branch, should now be working.

I haven't re-tested this manually - trusting that it works now, slash I can see you wrote an e2e test for it in that branch.

api/README.md
EmmaLRussell: yep, good spot, it had been updated in auth.md but not here. We probably don't need this many READMEs about auth!

I made a ticket to consolidate some of this stuff: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6452/

"oauth" here refers to authenticating with GitHub using its oauth path. Could have been better named perhaps. I think at one stage we were imagining supporting other OAuth providers so "oauth" and "github" authenticating got a bit conflated.

api/app/src/test/kotlin/packit/unit/service/PreAuthenticatedLoginServiceTest.kt
EmmaLRussell: Ha, ok, I just remembered why I did it this way - because Mockito can't mock final classes, so you can't do `mock<UserPrincipal>`, and it won't let you cast to the required type either. I will rename the value at least.

david-mears-2:
I see the integration tests do assert that the user is in the db -- but they don't assert anything about the claims in the token. Do we have any tests anywhere that confirm that tokens correctly contain the user's actual permissions?

EmmaLRussell:
> I see the integration tests do assert that the user is in the db -- but they don't assert anything about the claims in the token. Do we have any tests anywhere that confirm that tokens correctly contain the user's actual permissions?

Sorry, somehow managed not to see your long comment. Yep, integration tests are the way to go, neurotic levels of mocking of dubious benefit but can have some use as a regression. Have added a permissions check on the returned token in the integration tests.

PR #188, opened 2025-03-28
Mrc-6369 add/remove read packet permission on roles

absternator:
This PR adds an endpoint called `updatePacketReadPermissionOnRoles`. This will allow users to add/remove packets read permission on roles. The read permission is scoped to packetGroup or packet. This is determined by passing either _packetId_ or _packetGroupId_ as part of the body to the endpoint

Preauthorize to hit endpoint:
Users can only update packet.read if they have user.manage, global packet.manage permission. Or correct scoped packet.mange permission depending on scope of packet.read permission giving. If trying to add permission with packetGroup scope, then packet.manage for the packet group is needed. For a packet scope, then packet.manage for the packet or the packetgroup is necessary. 

Testing:
Setup packit and login as admin. Create some roles and users. 
hit the endpoint (curl or postman etc) with differing roles and check manage access in packit web  to ensure they are updating correctly. Below is example request with body.
`http://localhost:8080/roles/artefact-types/read-permissions`
`{
    "packetId": "20240729-155513-1432bfa7",
    "roleNamesToAdd": ["jojo"],
    "roleNamesToRemove":["jojo"]   
}`


api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -3,48 +3,68 @@ package packit.service
...
 interface RolePermissionService
 {
     fun removeRolePermissionsFromRole(role: Role, removeRolePermissions: List<UpdateRolePermission>): Role
     fun getRolePermissionsToAdd(role: Role, addRolePermissions: List<UpdateRolePermission>): List<RolePermission>
+    fun updatePacketReadPermissionOnRoles(
+        rolesToAdd: List<Role>,
+        rolesToRemove: List<Role>,
+        packetGroupName: String,
+        packetId: String? = null,
+    )
```
plietar, line 20: Seems like there's a lot of duplication between the existing methods and the new one.

The duplication might make some sense (the old method does one role many permissions, whereas the new one does many roles one permission), but the structure of the two should at least be coherent:

- The existing methods' logic is split between the `RoleService` and `RolePermissionService`. `RoleService.updatePermissionsToRole` does the overall update, calling the two methods of `RolePermissionService`. The `RolePermissionService` creates the roles, but `RoleService` is the one that saves them to the role.

- The new method has all of its logic in the `RolePermissionService` class. There's a bit of code in `RoleService`, but that's just some preprocessing. In particular `RoleService` does no database manipulation.

I think the new structure is much clearer - splitting the logic across the two classes is more confusing than anything else. The `getRolePermissionsToAdd` method doesn't make a whole lot of sense as a standalone method. I would move all of the `RoleService.updatePermissionsToRole` logic into `RolePermissionService`, and have a uniform API for the two kinds of updates.

While we're at it, I think you might as well generalize `updatePacketReadPermissionOnRoles` to any permission (not just `packet.read`), in which case the arguments are basically a `UpdateRolePermission` object.

The `RolePermissionService` interface then becomes:
```kotlin
interface RolePermissionService
{
  fun updateRolesOnPermission(
    permission: UpdateRolePermission,
    rolesToAdd: List<Role>,
    rolesToRemove: List<Role>,
  )
  fun updatePermissionsOnRole(
    role: Role,
    permissionsToAdd: List<UpdateRolePermission>,
    permissionsToRemove: List<UpdateRolePermission>,
  )
}
```

(The above doesn't quite work because `UpdateRolePermission` has integer IDs to refer to the tag and packet group, but IMO that's a mistake and we should never expose database IDs on the API)

api/app/src/main/kotlin/packit/service/RoleService.kt
```diff
@@ -180,4 +178,35 @@ class BaseRoleService(
...
+            rolesToUpdate.filter { it.name in updatePacketReadRoles.roleNamesToRemove },
+            packetGroupName,
+            updatePacketReadRoles.packetId,
+        )
+    }
+
+    internal fun getUniqueRoleNamesForUpdate(roleNamesToAdd: Set<String>, roleNamesToRemove: Set<String>): Set<String>
+    {
+        // Calculate symmetric difference (roles that appear in only one of the sets)
+        val roleNamesToUpdate = (roleNamesToAdd - roleNamesToRemove) + (roleNamesToRemove - roleNamesToAdd)
```
plietar, line 203: Might be simpler to just make it a 400 error to have a non-empty intersection?

Also I think this should just all be in `RolePermissionService`. There's no benefit in splitting the logic like this.

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -81,4 +101,90 @@ class BaseRolePermissionService(
...
+            packetGroup
+        )
+        addPermissionToRoles(
+            rolesToAdd,
+            permission,
+            packet,
+            packetGroup
+        )
+        roleRepository.saveAll(rolesToAdd)
+    }
```
plietar, line 135: How atomic is this whole process? If an error is thrown by `addPermissionToRoles`, would the removal still take place? Can we use a transaction to make sure this is an all-or-nothing? (I have no experience with JPA so I don't know what's atomic and what isn't)

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -81,4 +101,90 @@ class BaseRolePermissionService(
...
+            val rolePermission = RolePermission(
+                role = role,
+                permission = permission,
+                packet = packet,
+                packetGroup = packetGroup,
+                tag = tag,
+            )
+            if (role.rolePermissions.any { it == rolePermission })
+            {
+                throw PackitException("rolePermissionAlreadyExists", HttpStatus.BAD_REQUEST)
```
plietar, line 156: Alternatively, we could make this method idempotent and do nothing if the entry already exists and not throw an error. I'd be fine with either approach.

api/app/src/main/kotlin/packit/security/AuthorizationLogic.kt
```diff
@@ -15,6 +15,7 @@ class AuthorizationLogic(
     fun canReadPacket(operations: SecurityExpressionOperations, packet: Packet): Boolean =
         // TODO: update with tag when implemented
         operations.hasAnyAuthority(
+            "user.manage",
```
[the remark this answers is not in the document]
absternator: after working on FE, realised it makes sense as this permission allows users to manage roles and users. So they could just go in and edit their own role and give themselves read permission. Also leads on from the fact that if have packet.manage then you can read packets, thus it makes sense to allow if you have user.manage to read packets too

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: yeah that may be an option. Im personally more of a fan of failing just because the user will know if they have messed something up, and not get confused like thinking they have added a role or something.

api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -78,4 +75,21 @@ class RoleController(private val roleService: RoleService, private val userRoleS
         val role = roleService.getRole(roleName)
         return ResponseEntity.ok(role.toDto())
     }
+
+    @PutMapping("/{packetGroupName}/read-permissions")
```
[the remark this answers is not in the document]
absternator: yeah agreed that does seem a bit weird, I have updated to just `/read-permissions` and added packetGroupName to body of request..

I initially had it in the packetGroupController, but honestly made more sense to keep it here as the resources you are altering are all to do with roles and permissions (not packets or packet groups).

api/app/src/main/kotlin/packit/service/RoleService.kt
absternator: sorry what do ya mean? it does throw a 400 already when its empty? 

also split was just a nice internal function to help reading a bit easier

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: ahh yup good catch !!! actually it wasnt atomic I've added the @transactional so that should rollback if errors occcur

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -81,4 +101,90 @@ class BaseRolePermissionService(
...
+                tag = tag,
+            )
+            val matchedRolePermission = role.rolePermissions.find { it == rolePermission }
+                ?: throw PackitException("rolePermissionDoesNotExist", HttpStatus.BAD_REQUEST)
+
+            rolePermissionsToRemoveIds.add(matchedRolePermission.id!!)
+            role.rolePermissions.remove(matchedRolePermission)
+        }
+
+        rolePermissionRepository.deleteAllByIdIn(rolePermissionsToRemoveIds)
```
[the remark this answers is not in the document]
absternator: omg you are legend!!! im sure i tried this initially when I made the other methods but it wasn't working but that was because I didn't make it @transactional with the catching the exception!!! so now looks way cleaner thanks

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: yeah agreed seemed a bit odd, ive made both to just update now. I've also added passed in permission and made it more generic. but couldn't pass a UpdateRolePermission because of ids etc.. I've made ticket to come back and revisit -> https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6393/Rethink-permission-payloads

api/app/src/main/kotlin/packit/service/RoleService.kt
plietar: My understanding of what this tries to do is that if you try to do `{"add": ["alice", "bob"], "remove": ["alice", "charlie"]}` then it will simplify that as `{"add": ["bob"], "remove": ["charlie"]}` right?

My suggestion was to treat that kind of input as an error instead of trying to sanitize it. If `roleNamesToAdd & roleNamesToRemove` is non-empty (eg. in this case it is "alice") then you throw an error. Once you've done that you don't need to worry about filtering.

No strong opinions about it, happy to go with either semantics.

> also split was just a nice internal function to help reading a bit easier

I don't mind it being split into a separate function, but I think it should be in the `RolePermissionService` class. It's not really clear why `updatePacketReadPermissionOnRoles` needs to be in this class at all, the controller could call the `RolePermissionService` directly.

api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -78,4 +75,22 @@ class RoleController(private val roleService: RoleService, private val userRoleS
         val role = roleService.getRole(roleName)
         return ResponseEntity.ok(role.toDto())
     }
+
+    @PutMapping("/read-permissions")
```
plietar, line 79: This is still problematic, since you could have a role called `read-permissions` which would clash with this. We don't currently have a PUT on `/{roleName}`, but we might. There's also a could chance you will want to have a `GET /read-permissions`, in which case this route pattern definitely becomes an issue since there is a `GET /{roleName}` already.

 I honestly think putting this on the PacketGroupController makes the most sense. You are changing the properties of a packet group, so it should belong there.

You could even have it on both the PacketController and PacketGroupController:
- `PUT /packetGroups/{name}/read-permissions`
- `PUT /packets/{id}/read-permissions`

and maybe one day
- `PUT /tags/{id}/read-permissions`

I don't know if having separate routes makes the FE more or less complicated.

Would it make sense to broaden this to `/permissions` while we're at it? And have the permission as an argument to the body? You could definitely imagine a scenario where someone wants to grant `packet.manage` permission, or they want to grant `packet.run` permission this way (it's okay if we don't implement that in the FE yet, but it feels weird to special case the routes and function names in here).

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -2,83 +2,182 @@ package packit.service
...
+    {
+        val permission = permissionService.getByName(permissionName)
+        val packet = packetId?.let { packetService.getPacket(it) }
+        val packetGroup = if (packetId == null)
+        {
+            packetGroupService.getPacketGroupByName(packetGroupName)
+        } else
+        {
+            null
+        }
```
plietar, line 67: What happens here if the packet doesn't match the packetGroup name? I think we should check that and fail.

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
```diff
@@ -2,83 +2,182 @@ package packit.service
...
+                role = role,
+                permission = permission,
+                packet = packet,
+                packetGroup = packetGroup,
+                tag = tag,
+            )
+            val matchedRolePermission = role.rolePermissions.find { it == rolePermission }
+                ?: throw PackitException("rolePermissionDoesNotExist", HttpStatus.BAD_REQUEST)
+
+            role.rolePermissions.remove(matchedRolePermission)
```
plietar, line 180:
```suggestion
            if (!role.rolePermissions.remove(rolePermission)) {
                throw PackitException("rolePermissionDoesNotExist", HttpStatus.BAD_REQUEST)
            }
```

List's remove gives you back a boolean: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-mutable-list/remove.html

Same applies to the `removeRolePermissionsFromRole` method

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
plietar: LGTM. I think having names that are a bit more different than just the position of an `s` would be better, but up to you.

plietar: In a way this is actually quite similar to the structure of `UserRoleService` methods now

api/app/src/main/kotlin/packit/service/RoleService.kt
```diff
@@ -94,10 +95,9 @@ class BaseRoleService(
         val role = roleRepository.findByName(roleName)
             ?: throw PackitException("roleNotFound", HttpStatus.BAD_REQUEST)
```
plietar, line 96: I think we should just remove this wrapper from this class and have the controller call `RolePermissionService.updatePermissionsOnRole` directly.

The `RoleController` calls `UserRoleService.updateRoleUsers` directly for example.

api/app/src/main/kotlin/packit/controllers/RoleController.kt
plietar: Looking at the user-role relation, we have
- `PUT /roles/{roleName}/users`: add/remove many users to a single role
- `PUT /user/{username}/roles`: add/remove many roles to a single user

So a pretty similar structure, and the controller responsible for the route is the "single" object one.

api/app/src/main/kotlin/packit/security/AuthorizationLogic.kt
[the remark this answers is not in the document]
plietar: > after working on FE, realised it makes sense as this permission allows users to manage roles and users. So they could just go in and edit their own role and give themselves read permission

By that logic, anywhere that does any permission check should also allow `user.manage`. Just because the user could add themselves doesn't mean we should do it preemptively.

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: yeah i just updated because was confusing

api/app/src/main/kotlin/packit/service/RolePermissionService.kt
absternator: good catch!

api/app/src/main/kotlin/packit/service/RoleService.kt
absternator: i feel it just splits the logic up better... preprocess any role related behaviour in the role service, so the rolepermission service can just deal with their relationship

api/app/src/main/kotlin/packit/service/RoleService.kt
absternator: yeah i feel ya but ive already gone with the do nothing approach in other endpoints so will just leave for now (wont really happen in real life as the FE wont allow this).
also its split here because its role related code like pre-filtering of sorts before doing actual logic to do with rolepermissions

api/app/src/main/kotlin/packit/controllers/RoleController.kt
absternator: as per chat have put into 1 endpoint in `packetGroupController` `PUT /packetGroups/{name}/read-permissions`

plietar (approved):
I still find the code structure a little wonky, but I'm happy to agree to disagree and merge this as is.

Was the switch from plural permissions to singular in the REST API intentional?

api/app/src/main/kotlin/packit/service/RoleService.kt
absternator: sorry by do nothing i meant like not throw error.. i kinda do get your point but it just handles silliness from client if they were to just hit the API (not like its heavy filtering).

Also I kinda do get the approach you are coming from but its just  a different approach may be the correct one... but the current approach is still really clear as the rolepermissionservice is almost being used as an utility service that can call out to other services to do processing and get the data it needs for role permissions...

But still I've created a ticket to revisit this logic and go with your approach suggested --> https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6397/Rethink-RolePermissions-and-role-service

absternator:
> I still find the code structure a little wonky, but I'm happy to agree to disagree and merge this as is.
> 
> Was the switch from plural permissions to singular in the REST API intentional?

Have made ticket to revisit structure as per comments.
And yes it was since its updating single read permission.And pluralising made `read` sound more like a verb

PR #189, opened 2025-04-03
Mrc-6385 Manage Read Access for packet groups

absternator:
This PR adds the ability to manage read permissions on the packet groups page (home page) as seen on [figma](https://www.figma.com/design/eqmmVN15v6e7HGziiUZPWe/Packit-%22Publish%22?node-id=0-1&p=f&t=kjw5OqsCgZUgHHHj-0). There is slight difference in the modal dialog, the pills being different color currently not possible with multi select we use, thus have decided to color code the drop down itself and add a literal "Role" | "User" to help decide. This can all be changed when/if we choose to overhaul the permission forms as per figma discussions, [ticket here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6391/Update-Manage-Permission-forms).

This is branched off the API for managing permissions, thus the endpoint itself may change slightly (shouldn't affect this review)

Below are the changes and assumptions made about how the logic is working:

Note: Permission logic can be seen in `hasPermission.ts`, should make things clearer than me explaining.

- On the page, show the update icon only if the user has packet. manage permission
- On modal, a user can only grant/remove packet.read scoped to packetGroup (eg. packet.read:packetGroup:groupA)
  - To grant permission, user/role must currently not be able to read that packetGroup. If they have global permission then the user/role wont show up in select dropdown(as they already have permission to read).
  - To remove permission, user/role must have the read permission scoped to the packet group. If they have global access they will not show up (as we are only allowed to remove scoped to the permission group)

Testing:
Sign in with admin and create users and roles. Play around adding permissions and removing permissions in the user manage page and then also in the home page modal with managing read access. Ensure the behaviour is as expected. (as in roles/users should be added/removed and show up as expected in dropdowns)

Below are images of changes:

image
image
image


app/src/app/components/contents/home/PacketGroupSummaryListItem.tsx
```diff
@@ -1,47 +1,63 @@
...
+  rolesResponse?: ReturnType<typeof useGetRolesWithRelationships>;
 }
 
-export const PacketGroupSummaryListItem = ({ packetGroup }: PacketGroupSummaryListItemProps) => {
+export const PacketGroupSummaryListItem = ({ packetGroup, rolesResponse }: PacketGroupSummaryListItemProps) => {
   const { unit, value } = getTimeDifferenceToDisplay(packetGroup.latestTime)[0];
+  const { user } = useUser();
 
   return (
-    <li key={packetGroup.latestId} className="p-4 flex flex-col border-b">
```
absternator, line 14: just put inside div to add button no changes

app/src/lib/auth/hasPermission.ts
absternator, line 1: good file to understand permission logic

app/src/app/components/contents/home/utils/getRolesAndUsersCantReadGroup.ts
```diff
@@ -0,0 +1,47 @@
...
+    (user) =>
+      !userHasDirectReadPermission(user, packetGroupName) && !userHasReadPermissionViaRole(user, roles, packetGroupName)
+  );
+
+  // sorted by roles first
+  return [...rolesCantRead, ...usersCantRead];
+};
+
+/**
+ * Checks if a user has direct read access to the given packet group
```
absternator, line 32: below are just helper funcs for top one

app/src/app/components/contents/home/utils/getRolesUsersWithReadGroupPermission.ts
```diff
@@ -0,0 +1,26 @@
...
+import { hasPacketReadPermissionForGroup } from "../../../../../lib/auth/hasPermission";
+import { mapPermissionsToNames } from "../../../../../lib/constructPermissionName";
+import { RoleWithRelationships } from "../../manageAccess/types/RoleWithRelationships";
+import { UserWithRoles } from "../../manageAccess/types/UserWithRoles";
+
+/**
+ * This function retrieves roles and users that have read permissions for a specific packet group.
+ * It does NOT include those with global read permissions, manage permissions, etc.
+ */
+export const getRolesUsersWithReadGroupPermission = (
```
absternator, line 12: is used in remove permissions multi-select

app/src/app/components/contents/home/utils/getRolesAndUsersCantReadGroup.ts
```diff
@@ -0,0 +1,47 @@
...
+import { canReadPacketGroup } from "../../../../../lib/auth/hasPermission";
+import { mapPermissionsToNames } from "../../../../../lib/constructPermissionName";
+import { RoleWithRelationships } from "../../manageAccess/types/RoleWithRelationships";
+import { UserWithRoles } from "../../manageAccess/types/UserWithRoles";
+
+/**
+ * This function retrieves roles and users that cannot read a specific packet group.
+ * It includes those with global read permissions, manage permissions, etc.
+ */
+export const getRolesAndUsersCantReadGroup = (
```
absternator, line 13: is used in grant  read permissions multi-select

app/src/lib/constructPermissionName.ts
```diff
@@ -1,17 +1,39 @@
...
+    return `${permission}:tag:${tag}`;
   }
 
   return permission;
 };
+
+const validateScopedPermissionArgs = (packetGroupName?: string, packetId?: string, tag?: string) => {
+  if ([packetGroupName, tag].filter((x) => x !== undefined).length > 1) {
+    throw new Error("Only one of packetGroupName or tag can be provided");
+  }
```
EmmaLRussell, line 35: Should also be an error if both tag and packetId are defined?

app/src/lib/auth/hasPermission.ts
```diff
@@ -1,5 +1,35 @@
-import { UserState } from "../../app/components/providers/types/UserTypes";
+import { buildScopedPermission } from "../constructPermissionName";
```
plietar, line 1: The frontend really should not have this much authorization logic in it. Doing a bit of authorization on the current user to decide which bits to turn on and off in the UI may be okay, but doing authorization on other users is pretty strange.

IMO the backend should have a `GET /read-permission` that tells you which roles have read access to the packet group, and then you'd use that to filter the list of roles that can be added/removed. The response to the GET could also tell you whether they have scoped access to the packet group (and therefore can be revoked) or if they have a more global access to it (eg. they have `packet.read`).

In #189 you wrote
> - allow reading of all roles in packet.manage permission (this will be needed by FE so they can decide what roles to add/remove)

In retrospect I think this was maybe not a good direction. Someone with `packet.manage.foo` should only be able to list all the roles and get the read-permission of `foo`, but they shouldn't be able to list all the permissions that all the roles have.

app/src/app/components/contents/home/UpdatePacketReadPermissionForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    try {
+      await fetcher({
+        url: `${appConfig.apiUrl()}/packetGroups/${packetGroupName}/read-permission`,
+        body: {
+          roleNamesToAdd: values.roleNamesToAdd,
+          roleNamesToRemove: values.roleNamesToRemove
+        },
+        method: "PUT"
+      });
```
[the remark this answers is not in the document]
absternator: yup beacuse we have username roles.... 

sounds good with API changes when we do single select

app/src/app/components/contents/home/utils/getRolesUsersWithReadGroupPermission.ts
```diff
@@ -0,0 +1,26 @@
...
+  users: UserWithRoles[],
+  packetGroupName: string
+): (RoleWithRelationships | UserWithRoles)[] => {
+  const rolesWithRead = roles.filter((role) =>
+    hasPacketReadPermissionForGroup(mapPermissionsToNames(role.rolePermissions), packetGroupName)
+  );
+
+  const usersWithRead = users.filter((user) =>
+    hasPacketReadPermissionForGroup(mapPermissionsToNames(user.specificPermissions), packetGroupName)
+  );
```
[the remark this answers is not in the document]
absternator: ohh yesss good catch updated...

app/src/app/components/contents/home/utils/getRolesAndUsersCantReadGroup.ts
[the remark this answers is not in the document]
absternator: i intentionally didnt make it consistent because they are not really opposite.. used `cantReadGroup` here because it means exactly that they cant read group at all (could be due to global permissions).. if I used `withoutReadGroup` that would be different as they could still read with global permissions without specifically having the read Group permission

app/src/app/components/contents/home/utils/getRolesAndUsersCantReadGroup.ts
```diff
@@ -0,0 +1,47 @@
...
+  users: UserWithRoles[],
+  packetGroupName: string
+): (RoleWithRelationships | UserWithRoles)[] => {
+  const rolesCantRead = roles.filter(
+    (role) => !canReadPacketGroup(mapPermissionsToNames(role.rolePermissions), packetGroupName)
+  );
+
+  const usersCantRead = users.filter(
+    (user) =>
+      !userHasDirectReadPermission(user, packetGroupName) && !userHasReadPermissionViaRole(user, roles, packetGroupName)
```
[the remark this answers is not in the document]
absternator: good idea ive added the first line... didn't add link to manage access because they may not have user.manage permission page to visit link...

app/src/app/components/contents/home/utils/getRolesUsersWithReadGroupPermission.ts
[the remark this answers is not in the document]
absternator: have updated to `getRolesAndUsersWithOnlyReadGroupPermission`

app/src/lib/auth/hasPermission.ts
absternator: lets have a chat about this tommroow

app/src/app/components/contents/home/utils/getRolesAndUsersCantReadGroup.ts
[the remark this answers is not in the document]
EmmaLRussell: Ah ok, makes sense. I think I'd still prefer `getRolesAndUsersWhoCantReadGroup` as it's slightly more grammatical but up to you! :laughing:

app/e2e/auth.setup.ts
```diff
@@ -11,8 +11,8 @@ const authMethodIsBasic = async (apiURL: string) => {
 };
 
 const getBasicCredentials = () => {
-  const user = process.env.PACKIT_E2E_BASIC_USER;
-  const password = process.env.PACKIT_E2E_BASIC_PASSWORD;
+  const user = "resideUser@resideAdmin.ic.ac.uk";
+  const password = "password";
```
EmmaLRussell, line 15: Why change these? These are using environment variables so that you can potentially run e2e test against non-local packit instances that potentially don't have these packit users. Currently our instances use github auth, but it would be good to be able to support non-local basic auth e2e tests.

app/e2e/index.spec.ts
```diff
@@ -60,4 +61,45 @@ test.describe("Index page", () => {
     const displayPacketGroupName = getReadableIdString(packetGroupName);
     await expect(await getBreadcrumbLocator(page)).toHaveText(`home${displayPacketGroupName}${displayPacketId}`);
   });
+
+  test("can add update read permission on packet group", async ({ page }) => {
```
EmmaLRussell, line 65: You'll need to mark this test as not runnable on deployed instances since it changes the state of the system. You can do this using the `@stateMutate` [tag](@stateMutate) - there's a bit about this in the e2e section of the readme.

app/src/lib/constructPermissionName.ts
```diff
@@ -1,17 +1,39 @@
...
+  if (tag) {
+    return `${permission}:tag:${tag}`;
   }
 
   return permission;
 };
+
+const validateScopedPermissionArgs = (packetGroupName?: string, packetId?: string, tag?: string) => {
+  if (tag && (packetGroupName || packetId)) {
+    throw new Error("tag cannot be provided with packetGroupName or packetId");
```
EmmaLRussell, line 34:
```suggestion
    throw new Error("Tag cannot be provided with packetGroupName or packetId");
```

EmmaLRussell (commented):
Looks good, just a couple of things to bear in mind with the e2e tests.

app/e2e/auth.setup.ts
absternator: OMG i wasn't supposed to commit this!!!!! good catch thanks I've reerted

app/src/lib/auth/hasPermission.ts
absternator: as per discussions this will be done as another PR

EmmaLRussell (approved):

PR #190, opened 2025-04-08
Mrc-6394 Fix issue of creating name with spaces

absternator:


api/app/src/main/kotlin/packit/controllers/RoleController.kt
```diff
@@ -24,7 +24,7 @@ class RoleController(private val roleService: RoleService, private val userRoleS
     {
         val role = roleService.createRole(createRole)
 
-        return ResponseEntity.created(URI.create("/roles/${role.name}")).body(role.toDto())
+        return ResponseEntity.created(URI.create("/roles/${role.id}")).body(role.toDto())
```
plietar, line 27: I'm confused by this. The `getRoleByName` endpoint requires a name, not a numerical ID. How does this fix anything? Why do spaces actually cause an issue? Do we need to add a role "slug"? Should we ban spaces from role names?

plietar: I think what you want is something like:
```
UriComponentsBuilder.fromUriString("/roles/{}").build(role.name)
```

That will properly encode the argument into the URI string. Spaces in the name should become `%20` and slashes should become `%2F` (and we should make sure requests to `/roles/foo%2fbar` actually works the way we want it to).

Or we could just not return any URI here. I don't think the frontend uses it anyway.

absternator: ohh yeah i thought we had /role/{id} endpoint... but you are right and we don't use URL anways so have removed... also good point with BE validation too so have added that

api/app/src/main/kotlin/packit/model/dto/CreateRole.kt
```diff
@@ -1,9 +1,14 @@
 package packit.model.dto
 
+import jakarta.validation.constraints.Pattern
 import org.jetbrains.annotations.NotNull
 
 data class CreateRole(
     @field:NotNull
+    @field:Pattern(
+        regexp = "^[a-zA-Z0-9]+(\\s[a-zA-Z0-9]+)*\$",
```
absternator, line 9: same regex as FE

plietar: The `\$` at the end seems suspicious. It should just be `$` I think.

You could use a character class (eg. `\w` or `\p{Alnum}`) instead of `[a-zA-Z0-9]`, though that’s just a matter of taste.

Finally I reckon using a literal space rather than `\s` may be better to avoid tabs and newlines and whatnot

https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html?is-external=true

plietar (approved):

PR #191, opened 2025-04-16
Cut excess verbiage on home page

david-mears-2:
I found the existing text a bit excessive.

I modelled the replacement text on what orderlyweb has (see [here](https://montagu.vaccineimpact.org/reports/)), which is a section called "Pinned reports" (forthcoming) and a section below it called "Find a report".

# Before

Screenshot from 2025-04-16 12-20-10

# After

Screenshot from 2025-04-16 12-19-37

### DIFF
--- a/app/src/app/components/contents/home/Home.tsx
+++ b/app/src/app/components/contents/home/Home.tsx
@@ -1,4 +1,5 @@
 import { useState } from "react";
+import { Search } from "lucide-react";
 import { FilterInput } from "../common/FilterInput";
 import { PacketGroupSummaryList } from "./PacketGroupSummaryList";
 import { PAGE_SIZE } from "../../../../lib/constants";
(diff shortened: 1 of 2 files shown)


app/src/app/components/contents/home/Home.tsx
```diff
@@ -9,17 +10,17 @@ export const Home = () => {
 
   return (
     <div className="flex justify-center">
-      <div className="h-full flex flex-1 flex-col space-y-8 p-8 max-w-5xl">
-        <div>
-          <h2 className="text-2xl font-bold tracking-tight">A place to view and manage packets</h2>
-          <p className="text-muted-foreground">Here&apos;s a list of packet groups</p>
-        </div>
+      <div className="h-full flex flex-1 flex-col space-y-6 p-8 max-w-5xl">
+        <h2 className="text-2xl font-bold tracking-tight">Find a report</h2>
```
EmmaLRussell, line 14:
```suggestion
        <h2 className="text-2xl font-bold tracking-tight">Find a packet group</h2>
```

EmmaLRussell (commented):
Yeah, probably don't need quite so much blurb, but I don't think we should mix in "report" with the packet group terminology ("report" is really something from OrderlyWeb times which some users iirc explicitly objected to as it meant something else to them... )

EmmaLRussell (approved):

PR #193, opened 2025-04-23
Add scripts for pinning packets and update homepage to show pinned packets

david-mears-2:
As our first iteration of the pinning feature, we are just pinning packets directly, not queries-for-packets. (The latter is a separate ticket.)

The changes comprise the following:

### 1. Add a 'Pinned packets' section to the home page

image

And with more pinned packets than can fit on a row, it will start using a horizontal scroll:

image

### 2. Add some scripts that create pins

* `db/bin/create-packet-pin` which can be used to handily create a pin on a specified packet. This is the intended way for us to configure an instance to pin a certain packet (there is no UI for this; it's an infrequent enough thing for an org to do that we didn't want to bother building a whole form for this). Remember to rebuild the db image before trying to use this new script.
* its counterpart `db/bin/delete-packet-pin`
* `scripts/dev-create-pins`, so named because it's intended for use in dev situations (local or perhaps packit-dev). This assumes the demo packets exist, and pins 3 of them by hard-coded id.

I updated `scripts/dev-start` so that by default it runs `scripts/dev-create-pins`, unless a flag `--create-pins-skip` is passed.

### 3. Less central changes

Some of the changes in this branch relate to the download buttons. Previously, the zip downloads button relied on some packet-related query parameters. But now that I want to reuse that component on the home page, we have to derive the packet id another way. So it becomes a component prop.

`artefactFiles.ts` is an extraction of some logic from the Artefacts component so that it can be reused by the PinnedPacket component.


api/app/src/main/kotlin/packit/controllers/PinController.kt
```diff
@@ -0,0 +1,29 @@
...
+{
+    @GetMapping("/packets")
+    fun getPinnedPackets(): ResponseEntity<List<PacketMetadata>>
+    {
+        val pins = pinRepository.findAll()
+
+        val packets = pins.map { pin ->
+            packetService.getMetadataBy(pin.packetId)
+        }.sortedByDescending { it.time.start }
+
```
absternator, line 26: for abstraction from repository maybe best to create a `PinService` that does this logic and calls repository and packet service

api/app/src/main/kotlin/packit/model/Pin.kt
```diff
@@ -0,0 +1,16 @@
...
+import jakarta.persistence.*
+import java.util.UUID
+
+@Entity
+@Table(name = "`pin`")
+class Pin(
+    @Id
+    @Column(name = "id", nullable = false)
+    @GeneratedValue(strategy = GenerationType.UUID)
+    var id: UUID? = null,
```
absternator, line 12: What was the reason for using UUID over auto-incrementing int?

david-mears-2: Habit! Is there a reason to prefer auto-incrementing int?

absternator: Unless you need UUID, then default should probably be auto-increment because it is easier to read, requires less storage space(4 bytes vs 16), and is consistent with the rest of the DB.

plietar: Heh, I think we might want to need a way to refer to pins in the API eventually (today you can uniquely identify a pin by packet ID, but once we have pinned queries we'll need a way to refer to those).

I really don't like having numerical auto-increment IDs in the API. We do it in a couple of places already but we should avoid it as much as possible. Using a UUID is much better.

If you think UUIDs look ugly then you can encode them using more characters:
- https://tedspence.com/api-design-friendly-id-numbers-1a78d37261c6
- https://dev.to/stripe/designing-apis-for-humans-object-ids-3o5a

api/app/src/main/kotlin/packit/model/dto/PinDto.kt
```diff
@@ -0,0 +1,5 @@
+package packit.model.dto
+
+data class PinDto(
```
plietar, line 3: I don't see this being used anywhere?

api/app/src/main/kotlin/packit/controllers/PinController.kt
```diff
@@ -0,0 +1,29 @@
...
+)
+{
+    @GetMapping("/packets")
+    fun getPinnedPackets(): ResponseEntity<List<PacketMetadata>>
+    {
+        val pins = pinRepository.findAll()
+
+        val packets = pins.map { pin ->
+            packetService.getMetadataBy(pin.packetId)
+        }.sortedByDescending { it.time.start }
```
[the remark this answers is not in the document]
david-mears-2: D'oh. I think so yes.

api/app/src/main/kotlin/packit/model/Pin.kt
plietar: > is consistent with the rest of the DB.

I don't think anything would be consistent.
- Roles, Packet Groups, Permissions, Tags and Role Permissions use auto-increment numeric IDs
- Users use UUID
- Packets use orderly IDs
- Task Info use ID generated by the orderly runner, that ultimately come from https://reside-ic.github.io/ids/reference/random_id.html

app/src/app/components/contents/home/hooks/useGetPinnedPackets.ts
```diff
@@ -0,0 +1,23 @@
+import useSWR from "swr";
+import appConfig from "../../../../../config/appConfig";
+import { fetcher } from "../../../../../lib/fetch";
+import { PacketMetadata } from "../../../../../types";
+
+console.log("url " + `${appConfig.apiUrl()}/pins/packets`);
```
david-mears-2, line 6: Need to remove this line. (Why didn't linting catch it?)

api/app/src/main/kotlin/packit/model/Pin.kt
plietar: It leaks implementation details, which is both an issue in terms of security (it reveals chronological ordering, it makes it possible to trivially guess past/future IDs and enumerate resources that aren’t supposed to be enumerated) and means clients have a tendency to make assumptions that make future changes more difficult. With random IDs that’s less likely to happen.

The sequential IDs are also merely unique to a single database rather than globally unique, which can make future migration/merging of two systems more difficult.

Lastly in a distributed setting, you want to avoid global coordination to generate non conflicting IDs. With sequential IDs you need a single source of truth for “the next index” whereas random IDs can easily be produced independently (if they are large enough to avoid collisions).

Now these are all general issues, which may or may not apply in our context, but it’s enough to be a general API design principle that’s worth sticking to.

david-mears-2: thanks for explaining!

david-mears-2: I particularly like your points about scaling

david-mears-2:
Emma points out that there needs to be a way to un-pin too.

EmmaLRussell (commented):
Looks great, works a treat. I was able to successfully delete a pinned packet using the script. 

Just some tiny suggestions/comments/questions inline which you can take or leave.

Also:

The download artefacts link is not working for pinned packets with a single artefact, I think because it's saving as a zip file but not actually zipping them. I got this error when I tried to open the downloaded "zip file" for "Packet with description"

image

But when I renamed the ".zip" to ".png" I found that it was actually a valid png image!


Not sure if this is a result of your styling changes in the index page, but the "n packets" label now seems to be squashed into two lines. I'm not seeing this with main branch on uat:
image

api/app/src/main/kotlin/packit/service/Scheduler.kt
```diff
@@ -13,7 +13,7 @@ class Scheduler(
     private val outpackServerClient: OutpackServerClient,
 ) {
 
-    @Scheduled(fixedDelay = 10000)
+    @Scheduled(fixedDelay = 10000, initialDelay = 0)
```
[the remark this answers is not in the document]
david-mears-2: I think what happened here was:
* I put in a 'create the pins!' command into the dev-start script
* This didn't work because when it was called, there weren't any packets in the database yet
* I initially solved this by telling dev-start script to wait for 10 seconds for packets to exist
* Then realised we can obviate the need for 10 seconds of wait by skipping the initial delay in the above way.

api/app/src/main/resources/db/migration/V20250303__create_pin_table.sql
```diff
@@ -0,0 +1,8 @@
+DROP TABLE IF EXISTS "pinned_packet_global";
+
+CREATE TABLE IF NOT EXISTS "pin"
+(
+    "id"        UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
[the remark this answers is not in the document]
david-mears-2: I feel a fear of stepping into the unknown by missing out an id column...

Beyond that primal fear, I can think of a concrete reason we'll want an id column: when (/if) we iterate the pinning feature to 'pin by query' rather than by packet (or in addition to 'by packet'), we'll no longer be able to rely on picking up a unique packet id to identify the pins.

david-mears-2:
> Not sure if this is a result of your styling changes in the index page, but the "n packets" label now seems to be squashed into two lines. I'm not seeing this with main branch on uat

I'm not seeing this locally. I wonder if it got fixed when I merged in main :thinking:

david-mears-2:
> The download artefacts link is not working for pinned packets with a single artefact, I think because it's saving as a zip file but not actually zipping them. I got this error when I tried to open the downloaded "zip file" for "Packet with description"

This has now been fixed by the commit
[Allow single-files to be downloaded as zip](https://github.com/mrc-ide/packit/pull/193/commits/efcc441e25061b02b2220d6f8dd909ace28d2230)

david-mears-2:
I've re-requested your review @EmmaLRussell, but docker has reached its limit for today so we won't know if CI passes...

david-mears-2:
TODO:
- merge in main
- then update the date of the migration so it's after the new migration on main.

EmmaLRussell (approved):
This looks good to me, pending build being fixed!
Weird css issue on Firefox not a showstopper, not sure if it's even originating on this branch...

david-mears-2:
Thanks @plietar - I will look into this after the next daedalus meeting, which is on Tuesday.

PR #194, opened 2025-04-23
Mrc-6386 Manage Read Permissions on packet

absternator:
This is a follow-up PR from https://github.com/mrc-ide/packit/pull/189. This is the work for managing read permissions on a specific packet.

Below is the list of work done and decisions: 
- show read-permissions side bar if user has permissions to manage packet
- show table of roles that can read the packet with associated users. 
- show list of users that have specific read access
- Add global filter to datatable. it is used in manage users, roles tables.
- On modal, a user can only grant/remove packet.read scoped to packet (eg. packet.read:packet:groupA:idA)
 To grant permission, the user/role must not currently be able to read that packet. If they have global permission, then the user/role will not show up in the select dropdown(as they already have permission to read).
   - To remove permission, user/role must have the read permission scoped to the packet. If they have global access they will not show up (as we are only allowed to remove scoped to the permission group)

Testing:
Sign in with admin and create users and roles. Play around adding permissions and removing permissions in the user manage page and then also read permissions for a packet. Ensure the behaviour is as expected. (as in roles/users should be added/removed and show up as expected in dropdowns)

Below are images of changes:

image
image


app/src/app/components/routes/Router.tsx
```diff
@@ -39,6 +40,7 @@ export const Router = () => {
               <Route path="/:packetName/:packetId" element={<PacketDetails />} />
               <Route path="/:packetName/:packetId/metadata" element={<Metadata />} />
               <Route path="/:packetName/:packetId/downloads" element={<Downloads />} />
+              <Route path="/:packetName/:packetId/read-permissions" element={<PacketReadPermission />} />
```
EmmaLRussell, line 43:
```suggestion
              <Route path="/:packetName/:packetId/read-access" element={<PacketReadPermission />} />
```

app/src/tests/components/contents/PacketReadPermission/PacketReadPermission.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <Route element={<PacketOutlet packetId={mockPacket.id} />}>
+              <Route path="/:packetName/:packetId/read-permissions" element={<PacketReadPermission />} />
+            </Route>
+          </Routes>
+        </MemoryRouter>
+      </UserProvider>
+    </SWRConfig>
+  );
+
+describe("PacketReadPermission", () => {
```
EmmaLRussell, line 35: Could test that it works with both global and specific packet manage permission?

app/src/tests/components/contents/PacketReadPermission/utils/getRolesUsersWithOnlyReadPacketPermission.test.ts
```diff
@@ -0,0 +1,91 @@
...
+const mockCanReadPacketGroup = jest.fn();
+const mockCanManagePacket = jest.fn();
+jest.mock("../../../../../lib/constructPermissionName", () => ({
+  mapPermissionsToNames: () => mockMapPermissionsToNames()
+}));
+jest.mock("../../../../../lib/auth/hasPermission", () => ({
+  hasPacketReadPermissionForPacket: () => mockHasPacketReadPermissionForPacket(),
+  canReadPacketGroup: () => mockCanReadPacketGroup(),
+  canManagePacket: () => mockCanManagePacket()
+}));
```
EmmaLRussell, line 17: Because all of these permission utils are quite tightly dependent on each other  I wonder if we shouldn't mock these `hasPermission` methods here. Like, I was wondering if we should test the case here of a user with global read permission... which of course should be covered by `canReadpacketGroup`, but that's being mocked.. Maybe a lot of this is going away in the refactor to the back end though, so it might not be worth it!

app/src/tests/components/contents/downloads/Downloads.test.tsx
```diff
@@ -24,7 +24,7 @@ describe("download component", () => {
       <SWRConfig value={{ dedupingInterval: 0 }}>
         <MemoryRouter initialEntries={[`/${packet.name}/${packet.id}/downloads`]}>
           <Routes>
-            <Route element={<PacketLayout />} path="/:packetName/:packetId">
+            <Route element={<PacketOutlet packetId={packet.id} />}>
```
EmmaLRussell, line 27: Why this change?

EmmaLRussell (commented):
Looks good!

Should maybe have a "none" label or something if there aren't any specific users with read access, rather than having an empty box. Could remove the section entirely, but I think it's useful to have it there so the user can see that this is a feature.

app/e2e/utils.ts
```diff
@@ -81,8 +81,11 @@ export const createEmptyTestRole = async (page: Page, testRoleName: string) => {
   await page.getByRole("button", { name: "Add Role" }).click();
   await page.getByRole("textbox", { name: "Name" }).fill(testRoleName);
   await page.getByRole("button", { name: "Add" }).click();
-  if (await page.locator("text=Role already exists").isVisible()) {
+  try {
+    await page.locator("text=Role already exists").waitFor({ timeout: 2000 });
     await page.getByRole("button", { name: "Close" }).click();
+  } catch {
+    // Role was created successfully
```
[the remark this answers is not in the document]
absternator: so its to create to a role. If we create the role for the first time, then `await page.locator("text=Role already exists").waitFor({ timeout: 2000 });` throws an error. and we have to catch but do nothing because it means role was created successfully

app/src/app/components/contents/PacketReadPermission/PacketReadPermission.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const PacketReadPermission = () => {
+  const { packetId, packetName } = useParams();
+  const { packet } = usePacketOutletContext();
+  const { user } = useUser();
+  const rolesResponse = canManagePacket(user?.authorities, packetName ?? "", packetId ?? "")
+    ? useGetRolesWithRelationships()
+    : null;
+
+  if (rolesResponse?.error?.status === HttpStatus.Unauthorized || rolesResponse === null) return <Unauthorized />;
```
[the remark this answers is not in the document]
absternator: in this case no becuase the `rolesResponse == null` only when `canManagePacket(user?.authorities, packetName ?? "", packetId ?? "")` is false. thus the user is unauthorized to view this page. 

note: logic will change a bit in new PR I'm working on atm with refactoring

app/src/app/components/contents/home/UpdatePacketReadPermissionForm.tsx
```diff
@@ -55,7 +53,8 @@ export const UpdatePacketReadPermissionForm = ({
         url: `${appConfig.apiUrl()}/packetGroups/${packetGroupName}/read-permission`,
```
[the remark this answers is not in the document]
absternator: yeah you are right it is a bit weird.. will address in next ticket as part of refactor

app/src/tests/components/contents/PacketReadPermission/PacketReadPermission.test.tsx
absternator: Didn't think that was needed as the logic for checking if user can manage packet  is in `canManagePacket`,

app/src/tests/components/contents/PacketReadPermission/utils/getRolesUsersWithOnlyReadPacketPermission.test.ts
absternator: yeah good idea!! il keep that in mind when testing the refactor code.... i wont change here as this function will be removed in refactor

app/src/tests/components/contents/downloads/Downloads.test.tsx
absternator: because packetlayout wraps packetoutlet. packetlayout needs more dependencies now (user) as well thus if we are testing `Downloads.test.tsx` don't need to test the layout as well, outlet should suffice

absternator:
> the text rather than "packet group" but it still seems a bit misleading/confusing.. It woul

omg yes sorry!!! I thought I had changed before but missed completly

EmmaLRussell (approved):
Thanks! :shipit:

PR #195, opened 2025-04-28
wip: custom branding

david-mears-2:
TODO - instead of doing a request through to the back-end for the app name, derive it from the index.html? (And store in appstore) This would be to reduce the time it takes for the app name to load initially. Alternatively use Anmol's suggestion of a Provider, which should block loading until the information is present


app/src/app/components/header/NavMenu.tsx
```diff
@@ -0,0 +1,81 @@
+import { hasPacketRunPermission, hasUserManagePermission } from "../../../lib/auth/hasPermission";
```
david-mears-2, line 1: This is a merge of LeftNav (which is no longer on the left, hence, 'NavMenu'), with NavMenuMobile. Both of those were making use of the same logic to process 'NavItems' into a list of nav items that the user had permission to see, so I thought it would make more sense to handle that just once, in a component that concerns itself with that logic, and then the markup follows from that.

PR #196, opened 2025-05-01
Move FE permissions + role logic to backend

absternator:
This is a follow-up ticket from managing packet read permissions. As per comments, the logic has been moved from the frontend to the backend with multiple cleanups. Since this PR is pretty big,  There is a separate PR #194 for the tests.

For consistency of logic can check the code in the deleted FE files referenced that match the new BE code. 

Below is the list of things changed/ refactored:

- Add `PermissionChecker.kt` that copies over logic from `hasPermission.ts`.This is used by authorizationLogic and other  services to help check permissions now, rather than relying on strings. 
- rename FE interface `UserWithRoles` to `UserWithPermissions`
- Remade the whole `RoleController` to require 'user.manage' permission.
- Update `useGetRolesWithRelationships.ts` to `useGetRolesAndUsersWithPermissions.ts` so we can get back all the roles and users that we need. Users are no longer derived from roles (`getUsersWithRoles()`). The new endpoint is at `UserRoleController.getRolesAndUsers()`. 
- Split up the initial endpoint  `PacketGroupController.updatePacketReadPermissionOnRoles` that was used to update read permissions on roles/users into 2 endpoints. 1 for packet group and 1 for packet. The packet one lives in `PacketController.updatePacketReadPermissionOnRoles`

Lists Needed for packet and packet group read permission update. The core replaced logic can be found in `UserRoleFilterService` which is called by `UserRoleService`.
- For updating packetGroup read permissions we no longer construct lists that will be displayed on dropdowns in FE. it is done in the BE at endpoint `PacketGroupController.getRolesAndUsersForReadPermissionUpdate`. It actually gets all roles&users for every packet group the user can manage.(this is so the FE can just make 1 request rather than 1 for each packet group).  The logic for what is returned replaces the files `getRolesAndUsersCantReadGroup.ts` and `getRolesAndUsersWithOnlyReadGroupPermission.ts`.
- Similar work is done for updtading permissions on sing,e packet. The endpoint is at `PacketController.getRolesAndUsersForReadPermissionUpdate`. As well as lists for dropdowns which replace code at `getRolesAndUsersCantReadPacket.test.ts` & `getRolesUsersWithOnlyReadPacketPermission.test.ts`.The endpoint also returns roles and specific users that can read the packet (replaces logic in file in `PacketReadPermission.tsx), this is what is displayed on the table of roles and list of users. 

Testing:

Launch up app and play around with managing roles and users. Then, try to update read permissions for packets and packet groups. Ensure the lists are as expected.


db/bin/create-super-user
```diff
@@ -31,6 +31,10 @@ psql -U packituser -d packit \
 INSERT INTO "user" (id, username, display_name, password, email, disabled, user_source, last_logged_in)
 VALUES (:'uuid', :'email', 'Super Admin', :'password', :'email', FALSE, 'basic', NOW());
 
+INSERT INTO "role" (name, is_username)
+VALUES (:'email', TRUE);
+
 INSERT INTO "user_role" (user_id, role_id)
-SELECT :'uuid', role.id FROM role WHERE role.name = 'ADMIN';
+SELECT :'uuid', role.id FROM role WHERE role.name IN ('ADMIN', :'email');
```
absternator, line 38: for consitency with rest of users

api/app/src/main/kotlin/packit/model/dto/UpdateRolePermission.kt
```diff
@@ -7,12 +7,22 @@ data class UpdateRolePermissions(
...
+data class RolesAndUsersToUpdateRead(
+    val cantRead: RolesAndUsersWithPermissionsDto,
+    val withRead: RolesAndUsersWithPermissionsDto
+)
+
+data class RolesAndUsersForPacketReadUpdate(
+    val cantRead: RolesAndUsersWithPermissionsDto,
+    val withRead: RolesAndUsersWithPermissionsDto,
+    val canRead: RolesAndUsersWithPermissionsDto,
+)
```
plietar, line 24: I don't fully understand the difference between those two types and the meaning of the fields. I think this at least deserves a comment to clarify it.

- `cantRead` is for people who cannot read the resource
- `withRead` is for people who have a read permission targeting that specific resource (ie. it can be revoked)
- `canRead` is for people who can read the resource, either thanks to a specific permission, or via an indirect broader permission

Why is it that PacketGroups use the two field `RolesAndUsersToUpdateRead` but not `RolesAndUsersForPacketReadUpdate`? If a user has the global `packet.read`, then they should be in the `canRead` list right?

I guess the frontend is currently missing the UI to display a packet group's `canRead` list, but having the backend expose a consistent API seems beneficial. I would use a single class for both endpoints.

----

Why do these types need to return so much information? I can sort of understand including the members of the role with it, but I don't see why we need to return all the permissions of a user/role with it.

---

I think this structure has some non-obvious redundancy and unwritten invariants that hurts understandability a bit. Unless I am mistaken, `cantRead` and `canRead` are mutually exclusive, `withRead` is always a subset of `canRead`, right?

A alternative approach would be to transpose the structure a bit, and do something like

```kt
data class AccessList {
    val users: List<UserWithAccessKind>,
    val roles: List<RolesWithAccessKind>,
}
data class UserWithAccessKind {
    val username: String,
    val access: AccessKind,
}
// RolesWithAccessKind is similar to UserWithAccessKind

enum class AccessKind {
    NoAccess,

    // User has the exact permission referred to, eg. `packet.read:foo:<id>`
    DirectAccess,

    // User has a broader permission, eg. it has global `packet.read`, or it has `packet.manage:foo:<id>`
    // Could also be called "implied access"
    IndirectAccess,

    // User can access a packet thanks to a permission that was granted to one of the non-username roles they are part of, rather than to themselves.
    InheritedAccess,
}
```

Here we map from user to their access kind. You could also merge the two lists as a general "roles" list that includes both username roles and non-username roles.

I'm not too fussy about using this particular structure, but I think it is worth considering.

api/app/src/main/kotlin/packit/model/dto/RolesAndUsers.kt
```diff
@@ -0,0 +1,14 @@
+package packit.model.dto
+
+import packit.model.Role
+import packit.model.User
+
+data class RolesAndUsers(
```
plietar, line 6: Given that Users are Roles, can't we just use a `List<Role>` here that encompasses both of them?

The whole benefit of having all Users be Roles is to avoid code duplication, we should take advantage of that.

api/app/src/main/kotlin/packit/service/UserRoleFilterService.kt
```diff
@@ -0,0 +1,153 @@
...
+
+interface UserRoleFilterService
+{
+    fun getRolesAndUsersWithSpecificReadPacketGroupPermission(
+        roles: List<Role>,
+        users: List<User>,
+        packetGroupName: String
+    ): RolesAndUsers
+
+    fun getRolesUsersWithSpecificReadPacketPermission(
```
plietar, line 18:
```suggestion
    fun getRolesAndUsersWithSpecificReadPacketPermission(
```

This matches the other function names better.

api/app/src/main/kotlin/packit/service/UserRoleFilterService.kt
```diff
@@ -0,0 +1,153 @@
...
+        packetGroupName: String
+    ): RolesAndUsers
+
+    fun getRolesAndUsersCantReadPacket(
+        roles: List<Role>,
+        users: List<User>,
+        packet: Packet
+    ): RolesAndUsers
+
+    fun getRolesAndSpecificUsersCanReadPacket(
```
plietar, line 36: I've lost track of what "Specific" means in this context. In the earlier functions "specific" means eg. `packet.read:foo:<id>` for the `<id>` packet, but here it is applied to the User? Does it mean non-inherited via a non-username role?

api/app/src/main/kotlin/packit/service/UserRoleFilterService.kt
```diff
@@ -0,0 +1,153 @@
+package packit.service
+
+import org.springframework.stereotype.Service
+import packit.model.Packet
+import packit.model.Role
+import packit.model.User
+import packit.model.dto.RolesAndUsers
+import packit.security.PermissionChecker
+
+interface UserRoleFilterService
```
plietar, line 10: The explosion of methods on here really makes me think that the `AccessKind` pattern described in my other comments would simplify things:

This is all that this service would really need to do:
```kt
fun getRoleAccessForPacketGroup(role: Role, name: String) -> List<AccessKind>
fun getRoleAccessForPacket(role: Role, packet: Packet) -> List<AccessKind>
```

- Role can be either a user or a non-username role
- The AccessKind is an enum describing the three possible states a role may have access to something.

api/app/src/main/kotlin/packit/model/dto/RolesAndUsers.kt
```diff
@@ -0,0 +1,14 @@
...
+    val roles: List<Role> = listOf(),
+    val users: List<User> = listOf()
+)
+
+data class RolesAndUsersWithPermissionsDto(
+    val roles: List<RoleDto> = listOf(),
+    val users: List<UserWithPermissions> = listOf()
```
plietar, line 13: Same comment about "users are roles" applies here I guess, isn't this just a list of roles? Does the roles list include usernames?

api/app/src/main/kotlin/packit/controllers/PacketGroupController.kt
```diff
@@ -62,17 +64,22 @@ class PacketGroupController(
         return ResponseEntity.ok(packetGroupService.getPacketGroupSummaries(payload, filter))
     }
 
+    @PreAuthorize("@authz.canManageAnyPacket(#root)")
+    @GetMapping("packetGroups/read-permission")
```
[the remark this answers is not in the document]
absternator: good catch have updated to `/packetGroups/user-role/read-permission`

api/app/src/main/kotlin/packit/service/UserRoleFilterService.kt
absternator: Yeah second point... Specific means its permissions are particular to that user... i.e not inherited via role

api/app/src/main/kotlin/packit/model/dto/RolesAndUsers.kt
absternator: That is what was returned originally... but the FE was required to mangle the data to get a list of roles and users (this refactor is trying to address this issue)... thus it made more sense to create a class that encompasses both of them

api/app/src/main/kotlin/packit/model/dto/RolesAndUsers.kt
absternator: as per above comment.

api/app/src/main/kotlin/packit/model/dto/UpdateRolePermission.kt
absternator: added comments, made consistent to return canRead For PacketGroup, returned less data now (basic data only)

api/app/src/main/kotlin/packit/service/UserRoleFilterService.kt
absternator: as per above comment

api/app/src/main/kotlin/packit/controllers/PacketGroupController.kt
```diff
@@ -62,17 +64,22 @@ class PacketGroupController(
         return ResponseEntity.ok(packetGroupService.getPacketGroupSummaries(payload, filter))
     }
 
+    @PreAuthorize("@authz.canManageAnyPacket(#root)")
+    @GetMapping("packetGroups/user-role/read-permission")
```
plietar, line 68: This is still not great and is prone to mess things up if a packetGroup is called `user-role`.

How about `packetGroups/_/read-permission`, where the underscore denotes that the endpoint is for the whole collection? I think it is reasonable to assume that no packet group will ever have that name.

(I still don't think this endpoint should even exist, but whatever)

api/app/src/main/kotlin/packit/controllers/UserRoleController.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.http.ResponseEntity
+import org.springframework.security.access.prepost.PreAuthorize
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import packit.model.dto.RolesAndUsersWithPermissionsDto
+import packit.service.UserRoleService
+
+@Controller
+@RequestMapping("/user-role")
```
plietar, line 12: Why is this needed? Having a dedicated endpoint that returns two things sounds like a red flag to me.

api/app/src/main/kotlin/packit/model/dto/UpdateRolePermission.kt
```diff
@@ -7,12 +7,22 @@ data class UpdateRolePermissions(
...
 
-data class UpdatePacketReadRoles(
-    val packetId: String? = null,
+data class UpdateReadRoles(
     val roleNamesToAdd: Set<String> = setOf(),
     val roleNamesToRemove: Set<String> = setOf()
 )
 
+/**
+ * @param cantRead Roles and users that are explicitly restricted from reading the resource.
```
plietar, line 16: > Roles and users that are **explicitly** restricted from reading the resource.

I don't think "explicitly" is the right word here. That would imply that they got blacklisted or banned from the resource, which is not the case.

Maybe more like "roles and users which have no permission to read the resource, directly or indirectly".

Can you clarify how these lists relate to one another? Are they mutually exclusive? Does withRead imply canRead?

api/app/src/main/kotlin/packit/model/dto/RolesAndUsers.kt
plietar: I think we should make two requests, one to `/users` and one to `/roles`. Not mash the two together.

api/app/src/main/kotlin/packit/model/dto/UpdateRolePermission.kt
plietar: > but it would require the FE to handle filter logic once again

Some simple filtering logic is fine if it means the API is easier to understand. It's the complex domain-specific reasoning that we had before that was an issue.

plietar (approved):
Honestly I still think this is way more difficult to understand than it should be, but I don't really want to delay this indefinitely. Please do change the name of the `packetGroups/user-role/read-permission` endpoint before merging.

In general I disagree with the design philosophy of building the API based on the exact needs of the UI. The API should make sense on its own, not just in the context of the frontend.

There's a lack of consistency across endpoints that should be similar. We have a `GET /packets/<id>/read-permissions`, a `PUT /packets/<id>/read-permissions`, a `PUT /packetGroups/<name>/read-permissions` but no `GET /packetGroups/<name>/read-permissions`. Instead the way to get the existing read permissions of a packet group is to fetch the read-permissions of all packet groups, and filter that down. The only justification for this inconsistency is that is how the UI is structured.

If the frontend needs two bits of information it can make two API calls. We shouldn't have an endpoint that returns two things (eg. `/user-role`).

The user vs role vs non-username role situation is still a mess that makes the logic so much more painful to reason about.

absternator:
> Honestly I still think this is way more difficult to understand than it should be, but I don't really want to delay this indefinitely. Please do change the name of the `packetGroups/user-role/read-permission` endpoint before merging.
> 
> In general I disagree with the design philosophy of building the API based on the exact needs of the UI. The API should make sense on its own, not just in the context of the frontend.
> 
> There's a lack of consistency across endpoints that should be similar. We have a `GET /packets/<id>/read-permissions`, a `PUT /packets/<id>/read-permissions`, a `PUT /packetGroups/<name>/read-permissions` but no `GET /packetGroups/<name>/read-permissions`. Instead the way to get the existing read permissions of a packet group is to fetch the read-permissions of all packet groups, and filter that down. The only justification for this inconsistency is that is how the UI is structured.
> 
> If the frontend needs two bits of information it can make two API calls. We shouldn't have an endpoint that returns two things (eg. `/user-role`).
> 
> The user vs role vs non-username role situation is still a mess that makes the logic so much more painful to reason about.

1. thanks I've updated the URL 
2. yeah I kind of agree and as a purest REST API I would think the same. But it still does make sense on its own. The base methods are all still there and adding extra endpoints specifically to make the FE more clean is a good idea IMO. Eg. the user-role, makes sense for FE to get both bits in 1 hit. there is no draw back of having that endpoint in the API.
3. The consistency is a bit of an issue, but I can easily just add the `GET /packetGroups/<name>/read-permissions` if thats an issue? it just wont be used currently but yeah can easily be added.
4. yeah that user vs role vs non-username role is a bit annoying. I'm tempted to just create a new user_permissions table and get rid of this hacky username role

PR #198, opened 2025-05-06
Provide base for absolute URLs for download URL utils

david-mears-2:
These functions that calculate URLs for requesting download-related entities were broken when deployed, because they were assuming that the configured 'api url' is absolute (which it is in development/testing).

The interface of URL requires absolute URLs, so we need to pass in a base for it to relativise the path against.

Tested by spinning up a packit-deploy constellation locally and downloading successfully (this correctly failed before I made the change)

### DIFF
--- a/app/.env.test
+++ b/app/.env.test
@@ -1,3 +1,3 @@
 # Tests don't actually hit this endpoint, but we
 # need the environment variable to be defined.
-REACT_APP_PACKIT_API_URL="http://localhost:8080"
+REACT_APP_PACKIT_API_URL="/packitApiUrl"
--- a/app/.env.test
+++ b/app/.env.test
@@ -1,3 +1,4 @@
-# Tests don't actually hit this endpoint, but we
-# need the environment variable to be defined.
+# Tests don't actually hit this endpoint, but we need the environment variable to be defined.
+# We also need it to be a relative URL, as it often is for production environments, to make
+# tests more realistic: sometimes a relative one can cause errors that an absolute one doesn't.
 REACT_APP_PACKIT_API_URL="/packitApiUrl"
--- a/app/src/tests/components/contents/runner/run/PacketRunForm.test.tsx
+++ b/app/src/tests/components/contents/runner/run/PacketRunForm.test.tsx
@@ -6,7 +6,6 @@ import { server } from "../../../../../msw/server";
 import { rest } from "msw";
 import { MemoryRouter, Route, Routes } from "react-router-dom";
 import { getTimeDifferenceToDisplay } from "../../../../../lib/time";
-import appConfig from "../../../../../config/appConfig";
 import { absoluteApiUrl } from "../../../../../helpers";
 
 const renderComponent = () => {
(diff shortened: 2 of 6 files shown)


app/src/tests/components/contents/runner/run/PacketRunForm.test.tsx
```diff
@@ -6,6 +6,9 @@ import { server } from "../../../../../msw/server";
 import { rest } from "msw";
 import { MemoryRouter, Route, Routes } from "react-router-dom";
 import { getTimeDifferenceToDisplay } from "../../../../../lib/time";
+import appConfig from "../../../../../config/appConfig";
+
+const testBaseUrl = "http://localhost";
```
plietar, line 11: Would it make sense for `${testBaseUrl}${appConfig.apiUrl()}` to be defined as a common constant somewhere? It might even make more sense to define it as `new URL(appConfig.apiUrl(), testBaseUrl)` instead.

I don't really know much about the frontend tests, how/where is `window.location.href` defined in this case? Can we use that source of truth instead of defining it ourselves?

app/.env.test
```diff
@@ -1,3 +1,3 @@
 # Tests don't actually hit this endpoint, but we
 # need the environment variable to be defined.
```
plietar, line 2: Seems worth updating this comment if the value is now important (or at least its format)

plietar:
Thanks! I've deployed this to packit-dev.dide.ic.ac.uk and the downloads are working!

app/src/tests/components/contents/runner/run/PacketRunForm.test.tsx
david-mears-2: I don't know where window.location gets defined - seems hard to find out! So I tried to find ways to set the location ourselves. I've given up trying to use a base other than 'localhost' in the test, but I _have_ found a way to mock the location's paths, which is helpful for testing that the current path doesn't get fed through to the url string.

david-mears-2: > It might even make more sense to define it as new URL(appConfig.apiUrl(), testBaseUrl) instead.

I think that would reduce my confidence in the tests - I want to test that, when we use this `URL` module, the resulting string is what we expect. So I want the util under test to handle the building of the URL, and I don't want to use any URL building utils in the test, just plain string interpolation.

david-mears-2: I've made the absolute api url into a common test helper, and the test base url into a common constant.

david-mears-2:
Thanks

app/src/tests/components/contents/runner/run/PacketRunForm.test.tsx
```diff
@@ -6,6 +6,8 @@ import { server } from "../../../../../msw/server";
 import { rest } from "msw";
 import { MemoryRouter, Route, Routes } from "react-router-dom";
 import { getTimeDifferenceToDisplay } from "../../../../../lib/time";
+import appConfig from "../../../../../config/appConfig";
```
plietar, line 9:
```suggestion
```

the linter says this isn't needed

plietar (approved):

PR #200, opened 2025-05-07
Validate packet group name in front-end urls

david-mears-2:
At the moment you can go to https://reside.packit.dide.ic.ac.uk/this-is-not-a-packet-group-name/20250214-210242-0856dd0b and the app does not complain about the incorrect packet group name in the url - it just accepts it at face value and presents the packet as being a member of this non-existent group.

Instead we should match the name in the packet metadata with the name in the url, and display a generic looking error if they don't match.

In this branch I also refactor PacketOutlet to use the url params rather than be passed props. This removes the need for the wrapper layout PacketBlankLayout as we can just use PacketOulet directly now.


app/src/app/components/contents/common/hooks/useGetPacketById.ts
```diff
@@ -2,13 +2,18 @@ import useSWR from "swr";
 import appConfig from "../../../../../config/appConfig";
 import { fetcher } from "../../../../../lib/fetch";
 import { PacketMetadata } from "../../../../../types";
+import { PacketNameError } from "../../../../../lib/errors";
 
-export const useGetPacketById = (packetId: string | undefined) => {
+export const useGetPacketById = (packetId: string | undefined, packetName: string | undefined) => {
```
absternator, line 7: shouldnt need to pass in packetName here

app/src/app/components/main/PacketOutlet.tsx
```diff
@@ -1,16 +1,29 @@
...
+  try {
+    const result = useGetPacketById(packetId, packetName);
+    packet = result.packet;
+    isLoading = result.isLoading;
+    error = result.error;
+  } catch (e) {
+    if (e instanceof PacketNameError) {
+      return <ErrorComponent error={e} message="Error fetching packet details" />;
+    }
+  }
```
absternator, line 26: easier to check if packetName from the URL is different packet.name then show error component 
`if (packet.name !== packetName) return <ErrorComponent error={e} message="Error fetching packet details" />;`

app/src/lib/errors.ts
```diff
@@ -9,3 +9,10 @@ export class ApiError extends Error {
     this.status = status; // Set the status
   }
 }
+
+export class PacketNameError extends Error {
```
absternator, line 13: can be removed

absternator (commented):
Nice bug fix much needed. have added comments to simplify logic...

app/src/lib/errors.ts
david-mears-2: What error would you pass to ErrorComponent instead?

absternator: oh yip you right

absternator (approved):
LGTM

PR #202, opened 2025-05-08
Custom branding - front end

david-mears-2:
TODO - instead of doing a request through to the back-end for the app name, derive it from the index.html? (And store in appstore) This would be to reduce the time it takes for the app name to load initially. Alternatively use Anmol's suggestion of a Provider, which should block loading until the information is present


app/src/app/components/header/NavMenu.tsx
```diff
@@ -0,0 +1,77 @@
+import { hasPacketRunPermission, hasUserManagePermission } from "../../../lib/auth/hasPermission";
```
david-mears-2, line 1: This is a merge of LeftNav (which is no longer on the left, hence, 'NavMenu'), with NavMenuMobile. Both of those were making use of the same logic to process 'NavItems' into a list of nav items that the user had permission to see, so I thought it would make more sense to handle that just once, in a component that concerns itself with that logic, and then the markup follows from that.

(I also merged in the separate NavigationLink component as it was only used here and didn't really do anything.)

app/src/app/components/providers/hooks/useGetBrandName.ts
```diff
@@ -0,0 +1,14 @@
+// Extract the brand name from the page title in index.html
+
+import { useMemo } from "react";
+
+export const useGetBrandName = () => {
+  const brandName = useMemo(() => {
+    if (process.env.NODE_ENV === "test") {
+      return "App Title";
+    }
```
david-mears-2, line 9: Couldn't for the life of me find a way to mock document values that worked, so this is my way of mocking it

app/src/styles/globals.css
```diff
@@ -24,73 +24,73 @@
```
david-mears-2, line 24: Moved color types like `hsl` and `rgb` from tailwind.config.js to here. This seemed to be necessary for the sake of having variables override / fall back to each other; it's also nicer to have the color channels defined next to the color types.

app/src/app/components/providers/hooks/useGetBrandName.ts
M-Kusumgar: you can put
```ts
document.title = "App title";
```
in `setupTests.ts` file and that should do it, document is just another object like window so can override its values like that

M-Kusumgar: also `useMemo` is used to cache expensive calculations, this value is very cheap, and also completely static, no point using any react machinery here, can just do basic typescript, can literally use `document.title` in react code, but if you want it to communicate that this is the brand name, can easily just have a document vars file and put `export const brandName = document.title;`

M-Kusumgar (commented):
Looks good, the tidy ups are nice too, had a couple of comments!

app/src/app/components/providers/BrandingProvider.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+const BrandingContext = createContext<BrandingContextType | undefined>(undefined);
+
+export const BrandingProvider = ({ children }: { children: ReactNode }) => {
+  const [logoConfigIsLoaded, setLogoConfigLoaded] = useState<boolean>(false);
+
+  const { logoConfig, error } = useGetLogoConfig();
+
+  useEffect(() => setLogoConfigLoaded(!!logoConfig), [logoConfig]);
```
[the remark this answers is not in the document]
david-mears-2: `isLoading` only starts being truthy once the request is made. `logoConfigIsLoaded` is meant to be false the whole time until the config is loaded.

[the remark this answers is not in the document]
M-Kusumgar: oh fair enough, i thought that `isLoading` would just be set to `true` straight away by `useSWR`, but also i realised, it might be simpler to just check `!!logoConfig` in `Header.tsx`, not sure if a `useState` and a `useEffect` is really needed but not fussed about this, up to you!

api/app/src/main/kotlin/packit/controllers/ConfigurationController.kt
```diff
@@ -0,0 +1,21 @@
+package packit.controllers
+
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+import packit.AppConfig
+import packit.model.dto.LogoDto
+
+@RestController
+@RequestMapping("/configuration")
```
EmmaLRussell, line 9: I think we want to try to be consistent between the existing `auth/config` endpoint and this. Perhaps this gives the wrong impression that all fetchable configuration will be under the `configuration` route whereas we've already got a competing pattern of `[FEATURE TYPE]/config` endpoints. So perhaps this endpoint should be `logo/config` with corresponding inversion of controller name and method.

api/app/src/test/kotlin/packit/integration/controllers/ConfigurationControllerTest.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import org.junit.jupiter.api.Test
+import packit.integration.IntegrationTest
+
+class ConfigurationControllerTest : IntegrationTest() {
+    @Test
+    fun `can get config`() {
+        val result = restTemplate.getForEntity("/configuration/logo", String::class.java)
+
+        assertSuccess(result)
```
EmmaLRussell, line 11: Might as well check we get expected content

app/public/index.html
```diff
@@ -2,11 +2,20 @@
...
+      'favicon.ico' with the filename of the configured favicon (if supplied).
+    -->
     <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Web interface for outpack" />
+    <!--
+      The deploy tool replaces the content of the (first) title tag with
+      a title tag containing the configured brand name (if supplied).
+    -->
```
EmmaLRussell, line 16: Ah, nice, really useful!

app/src/app/App.tsx
```diff
@@ -1,10 +1,9 @@
-import active from "../config/active.json";
 import { Header } from "./components/header";
 import { Main } from "./components/main";
 
 export const App = () => {
   return (
-    <div data-testid="app" className={active.instance}>
+    <div data-testid="app">
```
EmmaLRussell, line 6: Was this just not having any effect?

app/src/app/components/contents/runner/run/PacketRunPacketGroupField.tsx
```diff
@@ -23,18 +23,21 @@ export const PacketRunPacketGroupField = ({ form, packetGroups }: PacketRunPacke
...
           <Popover open={open} onOpenChange={setOpen}>
             <PopoverTrigger asChild>
               <FormControl>
                 <Button
                   variant="outline"
                   role="combobox"
                   aria-expanded={open}
-                  className={cn("w-[350px] sm:w-[450px] justify-between", !field.value && "text-muted-foreground")}
+                  className={cn(
+                    "hover:bg-background hover:text-inherit w-[350px] sm:w-[450px] justify-between",
```
EmmaLRussell, line 35: This is to stop the grey tint appearing on hover?

EmmaLRussell (commented):
As suggested earlier, I think it would be better to retain the user name initial view for informational purposes. 

Other than that, the changes looks great. We might eventually find other places to put the accent colour - as there aren't that many primary buttons, and the banner header border is quite subtle. But this seems like a great starting point. 

I wonder if it would be nice to have a little summary somewhere, either here in packit, or in the deploy repo, of how each part of the custom branding is implemented and deployed, as it is a bit of a mixture, and each part isn't necessarily going to be obvious to someone who might need to update this functionality.  

Is the idea that we can use the provided example logos to play with by setting relevant env vars  manually? There isn't a scripts for running up with custom logo? Probably don't need two...?

Might have been better to do the cosmetic and casing  tidy ups in another branch. Didn't see anything else obviously controversial.

EmmaLRussell:
Sorry,  jumped the gun on review. More incoming.

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -3,7 +3,7 @@ import { LocalStorageKeys } from "../../../lib/types/LocalStorageKeys";
 import { Theme, ThemeProviderProps, ThemeProviderState } from "./types/ThemeTypes";
 
 const ThemeProviderContext = createContext<ThemeProviderState>({
-  theme: "system",
+  theme: "light",
```
EmmaLRussell, line 6: Don't tell Anmol!
Not sure why we shouldn't still support user system preference, this seems like a regression. It's up to the deployment whether it provides dark-optimised values. Or could make light the default only if custom dark accent not provided?

app/src/app/components/header/Header.tsx
```diff
@@ -4,37 +4,36 @@ import { AccountHeaderDropdown } from "./AccountHeaderDropdown";
...
-            <div className="ml-auto flex items-center space-x-4">
-              {/* <NavigationLink to="/accessibility" className="mx-6 hidden md:flex">
-                Accessibility
-              </NavigationLink> */}
-              {hasUserManagePermission(user?.authorities) && (
-                <NavLink to="/manage-roles" className={cn(buttonVariants({ variant: "ghost" }), "justify-start")}>
-                  Manage Access
+      <div data-testid="header">
+        <div className="flex items-center h-20 border-b shadow-sm border-accent">
+          {logoConfigIsLoaded && brandName && (
```
EmmaLRussell, line 19: `brandName` should always be set shouldn't it, since it comes from the document title? Or does it get snarled up in some async via the provider?
I suppose someone could very foolishly set their custom brand name to be an empty string. They will have a bad time anyway..

app/src/app/components/header/NavMenu.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Menu } from "lucide-react";
+import { NavLink } from "react-router-dom";
+import { Button, buttonVariants } from "../Base/Button";
+
+export const NavItems: { [key: string]: string } = {
+  runner: "Runner",
+  "manage-roles": "Manage Access"
+  // "run-workflow": "Workflow Runner",
+  // documentation: "Documentation",
+  // accessibility: "Accessibility",
```
EmmaLRussell, line 14: I think we can probably retire our expectations of having a Workflow Runner any time soon, and Documentation is going to be handled elsewhere. Accessibility page remains an aspiration.

EmmaLRussell (commented):
Looks good! 

I wonder if it would be easyish to put in an e2e test of the custom branding, either here or in packit-deploy...

app/src/app/App.tsx
david-mears-2: Yeah it didn't have any obvious purpose

app/src/app/components/providers/ThemeProvider.tsx
david-mears-2: Yes, consciously making this feature second-class, as it isn't a requirement from VIMC, and makes life somewhere between once and twice as complicated. Pretty sure we've discussed this with Anmol present... (and Rich).

The main idea here is that organisations may not have the resources, or not care enough about dark mode, to have designed a logo that will work in both modes (which would often entail different colouring). If we follow the user's system preference, we are assuming that the organisation is equally happy with their branding resources in both cases. That assumption will often break. With dark mode fully supported, we would also be demanding more work from them to set up their instance: they would need to choose twice as many colours, and potentially twice as many logos and favicons. So we thought it would be preferable to make light mode the default, and then if a user still finds dark mode usable, they can make the switch. With light mode as default, orgs can reasonably provide only a single set of accent colours and logos.

Making light mode the default requires less work from us when we put in custom branding, since we can avoid having to write logic for the choice of which logo/favicon to display here, and simplifies the deploy tool logic too.

david-mears-2:
> Might have been better to do the cosmetic and casing tidy ups in another branch.

FYI I have a draft PR on casing waiting in the wings: https://github.com/mrc-ide/packit/pull/203

david-mears-2:
> Is the idea that we can use the provided example logos to play with by setting relevant env vars manually? There isn't a scripts for running up with custom logo? Probably don't need two...?

I included two because there seem to be two kinds of use case we should have in mind when developing: the known case VIMC, and the unknown generic case. Including two example logos is an attempt to prevent people baking in unwarranted assumptions that depend on VIMC's case.

(Adding the below to some documentation):

```md
The logo files are inserted by the deploy tool into the [public/img](app/public/img) folder. This repo may contain example logo files in the same folder, for development.

Developers can change the logo config either by setting the packit.branding properties in the back end ([here](api/app/src/main/resources/application.properties) or by setting the environment variables), or by manipulating the value returned by [useGetLogoConfig](app/src/app/components/providers/hooks/useGetLogoConfig.ts) in the front end.
```

david-mears-2:
> I wonder if it would be easyish to put in an e2e test of the custom branding, either here or in packit-deploy...

Did you spot [this test](https://github.com/mrc-ide/packit-deploy/blob/custom-branding/tests/test_integration.py#L170) over on packit-deploy?

david-mears-2:
> I wonder if it would be nice to have a little summary somewhere, either here in packit, or in the deploy repo, of how each part of the custom branding is implemented and deployed, as it is a bit of a mixture, and each part isn't necessarily going to be obvious to someone who might need to update this functionality.

We now have documentation on this repo too.

api/app/src/main/kotlin/packit/controllers/ConfigurationController.kt
david-mears-2: I'll rename these routes and controllers to 'branding configuration'.

app/src/app/components/providers/hooks/useGetBrandName.ts
david-mears-2: > document.title = "App title";

My god, you've done it! (Could've sworn I tried it... but it works now!)

app/public/css/custom.css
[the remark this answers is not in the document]
david-mears-2: Good point, I've done both the commented out variables and pointed devs to docs/custom-branding.md

app/src/app/components/providers/BrandingProvider.tsx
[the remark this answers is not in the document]
david-mears-2: Such a good point

app/public/img/exampleLogo.png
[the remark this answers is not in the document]
david-mears-2: google images :sunglasses:

api/app/src/test/kotlin/packit/integration/controllers/ConfigurationControllerTest.kt
david-mears-2: This depends on the env vars, so have updated the test env vars properties file for this purpose

app/src/app/components/header/Header.tsx
david-mears-2: Removed check

M-Kusumgar (approved):
LGTM!

PR #205, opened 2025-05-13
mrc-6470 endpoints to supports migrating users from orderly web

EmmaLRussell:
This branch does a couple of things to support migration of permissions from OrderlyWeb:

1. Adds a `/users/` endpoint to get all non-service users. I'm surprised we didn't have this already! As mentioned, I did add a separate controller for this as the existing `UserController` covers the singular `/user/` route, but we could combine those if you prefer. 
2. Add an endpoint  (`POST /user/external`)to create a user for auth types other than basic. I'm calling his creating an "external" user to cover both github and preauth. The idea here is that we'll usually want to create a user and its permissions before the user actually logs in via the external provider. It is worth having a separate endpoint as we need a different Dto for non-basic e.g. doesn't make sense to include a password, and email is optional. 

I also put in some cleanup code in the github auth integration tests, as this had been leaving users behind, which then caused unexpected results for the test of the new `/users` endpoint.


api/app/src/main/kotlin/packit/controllers/UsersController.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import packit.model.dto.UserDto
+import packit.model.toDto
+import packit.service.UserService
+
+@Controller
+@PreAuthorize("hasAuthority('user.manage')")
+@RequestMapping("/users")
```
absternator, line 14: heads up there is already a `UserController` so can add to that :smile:

EmmaLRussell: I saw that, but this is for `/users` plural rather than `/user` singular, so I thought maybe a separate controller would be in order. But was going to ask you about that, fine to combine them if you'd rather!

api/app/src/main/kotlin/packit/AppConfig.kt
```diff
@@ -36,6 +36,7 @@ class AppConfig(private val environment: Environment)
     val dbPassword: String = requiredEnvValue("db.password")
     val authJWTSecret: String = requiredEnvValue("auth.jwt.secret")
     val authRedirectUri: String = requiredEnvValue("auth.oauth2.redirect.url")
+    val authMethod: String = requiredEnvValue("auth.method")
```
EmmaLRussell, line 39: This is useful so we can just put the method into the `userSource` db column.

api/app/src/main/kotlin/packit/controllers/UsersController.kt
absternator: yeah definitely just stick it in the existing controller. It was my fault when I made it must called it `user` instead of `users` by mistake.. because to get singular you would just do `/users/{id}`

api/app/src/main/kotlin/packit/controllers/UsersController.kt
```diff
@@ -0,0 +1,24 @@
...
+class UsersController(private val userService: UserService,) {
```
absternator, line 15: rogue `,` at `UserService,`

api/app/src/main/kotlin/packit/service/UserService.kt
```diff
@@ -28,6 +31,7 @@ interface UserService
     fun checkAndUpdateLastLoggedIn(username: String)
     fun getServiceUser(): User
     fun getUserPrincipal(user: User): UserPrincipal
+    fun getAllNonServiceUsers(): List<User>
```
absternator, line 34: i have actually made this in my refactor PR.. would be great if you can replace the code please with below!!! 

UserService
```
override fun getAllNonServiceUsers(): List<User>
    {
        return userRepository.findByUserSourceNotOrderByUsername("service")
    }
```
UserRepository
```
fun findByUserSourceNotOrderByUsername(userSource: String): List<User>
```

absternator (commented):
LGTM!! just  a couple tiny comments

absternator (approved):
LGTM

PR #207, opened 2025-05-15
Enforce more formatting rules in Detekt

plietar:
Our detekt configuration disabled a number of rules, including indentation and spacing around braces.

This had a tendency of introducing noise in pull requests diffs, as different people's IDEs would use format things differently. Enabling those rules enforces consistency and avoids this issue in the future.


absternator:
superseded by #209

PR #208, opened 2025-05-15
mrc-5478 create-preauth-user helper adds user role

EmmaLRussell:
The helper script `create-preauth-user` helper script, was adding a user with a requested role (usually ADMIN) but confusingly was not also adding a user role for the new user!


absternator (approved):
LGTM

PR #212, opened 2025-05-16
Fix No auth mode

absternator:
Fix a bug where no auth mode would not work. The main culprit was `@EnableMethodSecurity` as this would override the no authentication in the config. these are the methods @PreAuthorize, @PostAuthorize etc.....

The issue of conditionally disabling was a bit annoying, and thus, the solution is the simplest way to make this possible. Pretty much we add a `SecurityEnabledConfig` with security annotations If `auth.enabled=true`

Testing: 
1. Try running with no auth and see if you can access endpoints 
2. Try run with auth enabled and you should need to be authenticated and authorized


api/app/src/main/kotlin/packit/security/WebSecurityConfig.kt
```diff
@@ -26,8 +29,18 @@ import packit.security.oauth2.OAuth2UserService
 import packit.security.provider.JwtIssuer
 import packit.service.BasicUserDetailsService
 
+@ConditionalOnProperty(prefix = "auth", name = ["enabled"], havingValue = "true", matchIfMissing = true)
+@Configuration
 @EnableWebSecurity
 @EnableMethodSecurity
+class SecurityEnabledConfig {
```
plietar, line 36: Would it make sense to apply the `ConditionalOnProperty` to the existing `WebSecurityConfig` Configuration class instead of creating a new one?

For example the `WebSecurityConfig.HttpSecurity.handleSecurePaths` method below has an `config.authEnabled` conditional, presumably that can go away?

api/app/src/main/kotlin/packit/security/WebSecurityConfig.kt
```diff
@@ -26,8 +29,18 @@ import packit.security.oauth2.OAuth2UserService
 import packit.security.provider.JwtIssuer
 import packit.service.BasicUserDetailsService
 
+@ConditionalOnProperty(prefix = "auth", name = ["enabled"], havingValue = "true", matchIfMissing = true)
```
plietar, line 32: Not a huge fan of having two different places that interpret the `auth.enabled` property, once here and once in `AppConfig`.

I believe something like this should work instead:
```kt
@ConditionalOnExpression("@appConfig.authEnabled")
```

This way the `AppConfig` remains the only source of truth for whether auth is enabled or not.

PR #213, opened 2025-05-20
Begin fixing broken dependencies links

david-mears-2:
Haven't tested manually (or programmatically) yet but my theory is that the 'query' property is not guaranteed to have a 'name' component. E.g. it can list just the id. See https://mrc-ide.github.io/orderly2/articles/query.html

Todo - try it out in dev and write back-end tests.


api/app/src/main/kotlin/packit/controllers/PacketController.kt
```diff
@@ -46,6 +46,12 @@ class PacketController(
         return ResponseEntity.ok(packetService.getMetadataBy(id))
     }
 
+    @GetMapping("/{id}/dependencies")
+    @PreAuthorize("@authz.canReadPacket(#root, #id)")
+    fun getPacketDependencies(@PathVariable id: String): ResponseEntity<List<PacketDto>> {
+        return ResponseEntity.ok(packetService.getDependencies(id).map { it.toDto() })
```
absternator, line 52: can return `BasicPacketDto` here as name, id is all you need in FE

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -85,6 +82,15 @@ class BasePacketService(
             .orElseThrow { PackitException("packetNotFound", HttpStatus.NOT_FOUND) }
     }
 
+    override fun getDependencies(id: String): List<Packet> {
+        val dependencyIds = getMetadataBy(id).depends.map() { it.packet }
+        return packetRepository.findAllById(dependencyIds)
+    }
```
absternator, line 88: this may not be needed. as if we cant really trust the dependencies what we get back from the api, it may make sense to update the `getMetadataBy` and then update the depends field to be a List<`BasePacketDto`>....

app/src/app/components/contents/packets/PacketDependencies.tsx
```diff
@@ -1,12 +1,12 @@
 import { ListTree } from "lucide-react";
-import { PacketDepends } from "../../../../types";
 import { AccordionContent, AccordionItem, AccordionTrigger } from "../../Base/Accordion";
-import { PacketDependency } from "./PacketDependency";
+import { Link, useParams } from "react-router-dom";
+import { useGetPacketDependencies } from "./hooks/useGetPacketDependencies";
+
+export const PacketDependencies = () => {
+  const { packetId } = useParams();
+  const { dependencies } = useGetPacketDependencies(packetId);
```
absternator, line 8: should get as part of getPacket call so dont have to make this extra request

absternator (commented):
Makes sense good fix... but seems as though it maybe best to attach on the `packetDependencies` or fully replace the `depends` on the `PacketMetadata` payload when calling `getMetadataById`..... this will save having to make the another same request to get the dependencies seperatly

api/app/src/main/kotlin/packit/service/PacketService.kt
david-mears-2: > if we cant really trust the dependencies what we get back from the api

Why can't we trust them?

david-mears-2: Oh do you mean that I'm deriving the dependencies from the proxy db, which might be out of date?

david-mears-2:
> Makes sense good fix... but seems as though it maybe best to attach on the packetDependencies or fully replace the depends on the PacketMetadata payload when calling getMetadataById..... this will save having to make the another same request to get the dependencies seperatly

Is the idea that the PacketService should always fetch dependencies info from outpack server? We'd still need to make multiple requests then - between the packit api and outpack server. And we'd be doing that even when we didn't need the dependency data.

absternator (commented):
replied to comment... also may want to pull in main again

app/src/app/components/contents/packets/PacketDependencies.tsx
```diff
@@ -18,10 +20,20 @@ export const PacketDependencies = ({ depends }: PacketDependenciesProps) => {
         </AccordionTrigger>
         <AccordionContent>
           <ul className="space-y-1 overflow-y-auto max-h-80">
-            {depends.length === 0 ? (
-              <div className="italic text-sm">This packet has no dependencies on other packets</div>
+            {dependencies && dependencies.length === 0 ? (
+              <div className="italic text-sm">None</div>
```
david-mears-2, line 24: Updated text to match that for the other sections on this page (parameters, and something else I forgot) which show 'None' when empty.

david-mears-2:
> Another approach could be to leave getMetadataById as is and create a new endpoint getPacketsById that is POST and takes in the body of a list of packet IDS. Then, the frontend can call this getPacketsById with the list of packet dependency IDS instead of getPacketDependencies. This will save calling getMetadataById again (which is much slower than reading from fs), and we can fetch packets directly from the DB.

I've gone with this design. Nice catch in saving calling getMetadataById again.

PR #216, opened 2025-05-29
mrc-5702 device flow back end

EmmaLRussell:
Implements device flow endpoints in Packit API. 

See write-up in docs/auth.md - this implementation of device flow is based, slightly loosely, on [the spec](https://datatracker.ietf.org/doc/html/rfc6749).

This implementation adds env vars for the device url (where users need to go to enter user code) and device flow expiry time in seconds) - there's a ticket [here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6518/Updates-to-packit-deploy-for-device-flow) to add these to deployer and configs. 

### Manual testing
The easiest way to test this manually is using the new `testDeviceFlow` node app, in combination with the /device page, implemented in [the next PR](https://github.com/mrc-ide/packit/pull/221):
In branch `mrc-5702-device-flow-front-end`:
1. Run local packit with basic auth:
```
./scripts/dev-start --super-user
```
2. Run the test app
```
cd testDeviceFlow
npm i
npm run test-local
```
This will show you a url to browse to and a user code to enter. When you enter the user code at the device url, you should see a token returned in the console. It also fetches from the packet group endpoint using the access token to check that the token actually works. 

### Testing with Montagu
You can also test with montagu auth if you run up [this branch](https://github.com/vimc/montagu-proxy/pull/99) of the montagu proxy:
```
./scripts/dev.sh
```
Then run from testDeviceFlow:
```
npm run test
```
and enter `https://localhost/packit/api` as the Packit API url. Login as test.user@example.com (password). If you run `./scripts/packit-create-test-user.sh` before you do so you'll have read access to packets so will see non-zero number of packet groups from the testDeviceFlow token check. 

(Testing local montagu is why there's the code in testDeviceFlow to allow self-signed certs.)


api/app/src/main/kotlin/packit/controllers/DeviceAuthController.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @PostMapping("/token")
+    fun fetchToken(
+        @RequestBody @Validated deviceAuthFetchToken: DeviceAuthFetchToken
+    ): ResponseEntity<DeviceAuthTokenDto> {
+        if (deviceAuthFetchToken.grantType != expectedGrantType) {
+            throw DeviceAuthTokenException("unsupported_grant_type")
+        }
+        val validatingUser = deviceAuthRequestService.useValidatedRequest(deviceAuthFetchToken.deviceCode)
```
plietar, line 63:
```suggestion
    @PostMapping("/token", consumes = [MediaType.APPLICATION_FORM_URLENCODED_VALUE])
    fun fetchToken(
        @RequestParam("grant_type") grantType: String,
        @RequestParam("device_code") deviceCode: String,
    ): ResponseEntity<DeviceAuthTokenDto> {
        if (grantType != expectedGrantType) {
            throw DeviceAuthTokenException("unsupported_grant_type")
        }
        val validatingUser = deviceAuthRequestService.useValidatedRequest(deviceCode)
```

I've had to change this to get this to work with httr2's client implementation. OAuth2 is a bit weird and has form-encoded requests but JSON-encoded responses throughout. The `/deviceAuth` request should also be form encoded (we don't use the body, but I think it would be better to at least specify a `consumes` on it).

For some reason I couldn't get `@RequestBody` to work with form encoding. Go figure.

The following will run the flow and print the full exchange to the console.
```r
client <- httr2::oauth_client(
  id = "orderly2",
  name = "orderly2",
  token_url = "http://localhost:8080/deviceAuth/token"
)

httr2::with_verbosity(verbosity=2, rlang::with_interactive(value = FALSE, {
  httr2::oauth_flow_device(client = client, auth_url = "http://localhost:8080/deviceAuth")
})
```

api/app/src/main/kotlin/packit/controllers/DeviceAuthController.kt
```diff
@@ -0,0 +1,67 @@
...
+            deviceAuthRequest.deviceCode.value,
+            deviceAuthRequest.userCode.value,
+            appConfig.authDeviceFlowVerificationUri,
+            appConfig.authDeviceFlowExpirySeconds
+        )
+        return ResponseEntity.ok(response)
+    }
+
+    @PostMapping("/validate", consumes = ["text/plain"])
+    fun validateDeviceAuthRequest(
```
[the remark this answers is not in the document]
EmmaLRussell: If we just wanted to send a single string as JSON, it should be wrapped in quotes I think, so i thought this would be simpler, but could make it `application/json` with the quotes to be consistent with content types in other requests if you'd rather. Or could make the request something like `{"userCode": "whatever"}`

api/app/src/main/kotlin/packit/AppConfig.kt
```diff
@@ -44,6 +44,8 @@ class AppConfig(private val environment: Environment) {
     val authEnabled: Boolean = requiredEnvValue("auth.enabled").toBoolean()
     val authGithubAPIOrg: String = requiredEnvValue("auth.githubAPIOrg")
     val authGithubAPITeam: String = requiredEnvValue("auth.githubAPITeam")
+    val authDeviceFlowExpirySeconds: Long = requiredEnvValue("auth.deviceFlowExpirySeconds").toLong()
+    val authDeviceFlowVerificationUri: String = requiredEnvValue("auth.deviceFlowVerificationUri")
```
plietar, line 48:
```suggestion
    val authDeviceFlowExpirySeconds: Long = requiredEnvValue("auth.deviceFlow.expirySeconds").toLong()
    val authDeviceFlowVerificationUri: String = requiredEnvValue("auth.deviceFlow.verificationUri")
```

I think this is clearer, but don't feel too strongly either.

If you make there a `@ConfigurationProperties` class instead, you can make the type a `Duration` and then we can use units in the property value. See for example the ServiceLoginConfig class.

api/app/src/main/kotlin/packit/service/DeviceAuthRequestService.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        if (request == null || request.validatedBy != null) {
+            throw DeviceAuthTokenException("access_denied")
+        }
+        // remove request from list if it is expired
+        if (isExpired(request)) {
+            removeRequest(request)
+            throw DeviceAuthTokenException("expired_token")
+        }
+        request.validatedBy = user
```
plietar, line 72: This needs to be synchronized. Two threads could be trying to validate the same request object at the same time and hit a data race on this.

Tbh I would add a `@Synchronized` annotation to the all the public methods and be done with it. Semantically it also makes sense for operations to be one transaction rather than one synchronize to read and one to update the validatedBy.

```kt
@Synchronized
override fun validateRequest(userCode: String, user: User) {
...
}
```

api/app/src/main/kotlin/packit/AppConfig.kt
plietar, line 48:
```suggestion
    val authDeviceFlowExpirySeconds: Long = requiredEnvValue("auth.deviceFlow.expirySeconds").toLong()
    val authDeviceFlowVerificationUri: String = requiredEnvValue("auth.deviceFlow.verificationUri")
```

I think this is cleaner, but don't feel too strongly either.

Also if you make there a `@ConfigurationProperties` class instead, you can make the type a `Duration` and then we can use units in the property value. See for example the ServiceLoginConfig class.

api/app/src/main/kotlin/packit/service/DeviceAuthRequestService.kt
```diff
@@ -0,0 +1,101 @@
...
+        return req
+    }
+
+    override fun cleanUpExpiredRequests() {
+        synchronized(requests, {
+            requests.removeAll { isExpired(it) }
+        })
+    }
+
+    override fun findRequest(deviceCode: String): DeviceAuthRequest? {
```
plietar, line 52: Does this need to be public?

api/app/src/main/kotlin/packit/service/DeviceAuthRequestService.kt
plietar: Maybe we should just use a database table and not have to worry about data races. We already have something very similar with the one_time_token that David added.

PR #217, opened 2025-06-05
Add script to promote a user

richfitz:
As tested on uat. I've wired it up against email as that is what is displayed in the interface

### DIFF
--- a/db/bin/promote-user
+++ b/db/bin/promote-user
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+set -eu
+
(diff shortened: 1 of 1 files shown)


EmmaLRussell (approved):
The script looks fine to me - I think the ADMIN role will always have id 1 though I suppose for clarity you could select that from the role table by name.

The front end tests are failing, but that seems to be something that crept into main  a couple of days ago - I think probably a couple of the smaller merges might have disagreed with each other.

PR #219, opened 2025-06-06
Iterate dark/light mode (theme)

david-mears-2:
Twinned with https://github.com/mrc-ide/packit-deploy/pull/18 - can review both PRs together.

1) Allow either light/dark mode, or both, to be enabled, using environment variables
2) If only one is enabled, remove the toggle
3) Respect 'system' preference for light/dark mode


api/app/src/test/kotlin/packit/integration/controllers/LoginControllerTest.kt
```diff
@@ -140,7 +140,6 @@ class LoginControllerTestPreAuth : IntegrationTest() {
     @Test
     fun `can login with existing user`() {
         val adminRole = roleRepository.findByName("ADMIN")!!
-        val adminPerms = adminRole.rolePermissions.map { it.permission.name }
```
david-mears-2, line 143: Unused variable

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -15,31 +25,70 @@ export const useTheme = () => {
...
-    root.classList.add(theme);
   }, [theme]);
 
   const value = {
+    availableThemes,
     theme,
-    setTheme: (theme: Theme) => {
-      localStorage.setItem(LocalStorageKeys.THEME, theme);
+    setCurrentTheme: (theme: Theme) => {
+      verifyThemeIsAvailable();
```
absternator, line 89: `setTheme` is being called 2 times if it is called in `verifyThemeIsAvailable`.... call once only 1

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -15,31 +25,70 @@ export const useTheme = () => {
...
-
-    root.classList.add(theme);
   }, [theme]);
 
   const value = {
+    availableThemes,
     theme,
-    setTheme: (theme: Theme) => {
-      localStorage.setItem(LocalStorageKeys.THEME, theme);
+    setCurrentTheme: (theme: Theme) => {
```
absternator, line 88: what does current theme mean? sorry a bit confused with wording? why would someone want to set to their current theme?

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -15,31 +25,70 @@ export const useTheme = () => {
...
+    return "light";
+  } else {
+    return privilegedTheme;
+  }
+};
+
+export const ThemeProvider = ({ children, ...props }: ThemeProviderProps) => {
+  const { brandingConfig, isLoading, error } = useGetBrandingConfig();
+
+  if (error) return <ErrorComponent message={error.message} error={error} />;
```
absternator, line 41: good to structure so easy to find things.... useStates at top, then variables, then functions, then useEffects and then return statements

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -1,10 +1,20 @@
 import { createContext, useContext, useEffect, useState } from "react";
 import { LocalStorageKeys } from "../../../lib/types/LocalStorageKeys";
 import { Theme, ThemeProviderProps, ThemeProviderState } from "./types/ThemeTypes";
+import { useGetBrandingConfig } from "./hooks/useGetBrandingConfig";
+import { ErrorComponent } from "../contents/common/ErrorComponent";
+
+// defaultTheme is to be used if both themes are available and no user preference is set in local storage
```
absternator, line 7: constants should probably be capital seprated by unserscore... eg. `DEFAULT_THEME`... also can these constants and `getSystemsTheme` function be moved to another file to help reduce clutter in this file

absternator (commented):
Nice work looks great!!! just small comments about strucuture

app/src/app/components/providers/ThemeProvider.tsx
david-mears-2: The word 'current' isn't really making a meaningful distinction. This re-name was entirely to help me reason about the code, because we had this `setTheme` property of the theme provider's exposed context, which was mirroring an internal `setTheme` variable (from useState), and I found it confusing to tell which one I was looking at when.

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -15,31 +25,70 @@ export const useTheme = () => {
...
+  if (error) return <ErrorComponent message={error.message} error={error} />;
+
   const [theme, setTheme] = useState<Theme>(
     () => (localStorage.getItem(LocalStorageKeys.THEME) as Theme) || defaultTheme
   );
 
+  const [availableThemes, setAvailableThemes] = useState<Theme[]>([]);
+
+  useEffect(() => {
+    if (!isLoading && brandingConfig) {
```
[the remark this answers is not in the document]
david-mears-2: Not sure about this - when only one mode is enabled, the toggle isn't needed, so you'd see a loader and then a blank space.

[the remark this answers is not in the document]
absternator: this is in the theme provider though right? nothing to do with the toggle button.... if the internet was slow the user would see a blank screen until the branding config is retrieved from the server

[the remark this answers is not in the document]
david-mears-2: oh right yes good point.

I think a loader might be overkill for this PR since none of the other providers use loaders, and they should probs be consistent. So I have made a ticket to add loaders to all providers: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6538/Add-a-loader-to-all-Providers

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -15,34 +19,60 @@ export const useTheme = () => {
...
-export const ThemeProvider = ({ children, defaultTheme = "light", ...props }: ThemeProviderProps) => {
+export const ThemeProvider = ({ children, ...props }: ThemeProviderProps) => {
+  const { brandingConfig, isLoading, error } = useGetBrandingConfig();
+
   const [theme, setTheme] = useState<Theme>(
-    () => (localStorage.getItem(LocalStorageKeys.THEME) as Theme) || defaultTheme
+    () => (localStorage.getItem(LocalStorageKeys.THEME) as Theme) || DEFAULT_THEME
   );
 
+  const availableThemes = useMemo((): Theme[] => {
```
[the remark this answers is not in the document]
david-mears-2: But it's a computed variable dependent on the response from the useGetBrandingConfig - so it needs to be updateable.

It's not wrapped up in the other useEffect because we want to expose it in the provider context, so it needs to have a more global scope. (The original design used a useState for this purpose.)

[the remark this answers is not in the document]
david-mears-2: anmol says consts are reactive in React!

app/src/app/components/providers/ThemeProvider.tsx
```diff
@@ -26,17 +26,14 @@ export const ThemeProvider = ({ children, ...props }: ThemeProviderProps) => {
...
-      return ["dark"];
-    } else if (brandingConfig.lightModeEnabled && !brandingConfig.darkModeEnabled) {
-      return ["light"];
-    } else {
-      return DEFAULT_AVAILABLE_THEMES;
-    }
-  }, [brandingConfig, isLoading]);
+  let availableThemes = DEFAULT_AVAILABLE_THEMES;
+  if (isLoading || !brandingConfig) {
+    availableThemes = [];
```
absternator, line 31: Can you please pull out into a function so we don't have to make availableThemes mutable and reassign.... 
`const availableThemes = getAvailableThemes(brandingConfig)`

absternator (approved):
Ive just left 1 more comment but approved so feel free to merge after thats addressed

PR #221, opened 2025-06-09
mrc-5702 device flow front end

EmmaLRussell:
Device flow page where logged-in user enters user code to be validated. The API part of device flow is in the parent branch and will be reviewed separately, this branch is just the front end part (so endpoint access may need to be tweaked following review of the back end). Includes an e2e test of the whole process. 

Uses a third party component `react-verification-input` which is configured to accept only letters and dashes. We upper-case any lower case letters which are entered, to match the format of the user codes. 

To test the full device flow manually:

1. Run local packit with basic auth:
```
./scripts/dev-start --super-user
```
2. Run the device flow test app
```
cd testDeviceFlow
npm i
npm run test-local
```
This will show you a url to browse to and a user code to enter. When you enter the user code at the device url, you should see a token returned in the console. It also fetches from the packet group endpoint using the access token to check that the token actually works.


app/src/app/components/routes/Router.tsx
```diff
@@ -22,6 +22,9 @@ export const Router = () => {
         </Route>
         {/* <Route path="accessibility" element={<Accessibility />} /> */}
         <Route path="redirect" element={<Redirect />} />
+        <Route element={<ProtectedRoute />}>
```
david-mears-2, line 25: confused why a login route would be protected

app/e2e/apiUtils.ts
absternator, line 1: is this file needed? can we use existing API util in `fetcher` in fetch.ts

app/package.json
```diff
@@ -52,6 +52,7 @@
     "react-paginate": "^8.2.0",
     "react-router-dom": "^6.12.1",
     "react-scripts": "5.0.1",
+    "react-verification-input": "^4.2.2",
```
absternator, line 55: as per comment we shouldnt need this

app/src/app/components/login/DeviceLogin.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const USER_CODE_LENGTH = 9;
+
+  const [userCode, setUserCode] = useState("");
+  const [resultStatus, setResultStatus] = useState<HttpStatus | null>(null);
+
+  const handleChange = (newValue: string) => {
+    setUserCode(newValue.toUpperCase());
+  };
+
+  const handleSubmit = async () => {
```
absternator, line 19: should use the  existing  util used across the app `fetcher` in `fetch.ts`...

app/src/app/components/login/DeviceLogin.tsx
```diff
@@ -0,0 +1,83 @@
+import VerificationInput from "react-verification-input";
+import { Button } from "../Base/Button";
+import appConfig from "../../../config/appConfig";
+import { HttpStatus } from "../../../lib/types/HttpStatus";
+import { useState } from "react";
+import { getAuthHeader } from "../../../lib/auth/getAuthHeader";
+import { Check } from "lucide-react";
+
+export const DeviceLogin = () => {
```
absternator, line 9: best to use react-hook-form and exisitng form components with https://ui.shadcn.com/docs/components/input-otp as fields... you can check other forms such as `AddRoleForm.tsx` for inspiration.

this will clean a lot of manual form handling logic up and worrying about state.. also will be consistent with rest of app

app/src/app/components/routes/Router.tsx
absternator: Login is a confusing term here.. maybe rename to github equivalent `DeviceActivation.tsx`

app/src/styles/globals.css
```diff
@@ -111,3 +111,13 @@ iframe {
   width: 1em;
   height: 1em;
 }
+
+.verification-input-selected-char {
```
absternator, line 115: shouldnt need if we remove this component... and if in the future you do need styling bes to use tailwind like rest of app

absternator (commented):
Nice work love the e2e test also!!!!! Just a few comments and some design qualms

- agree with paul we should not have to add "-" manually... 
- the UI doesn't fit and I don't see need for react-verification-input. can we use shadcn component please https://ui.shadcn.com/docs/components/input-otp
- Also the title and subtitle seem very specific and not quite correct... we should just copy what  github shows as per pauls screenshot

EmmaLRussell:
> the UI doesn't fit and I don't see need for react-verification-input. can we use shadcn component please https://ui.shadcn.com/docs/components/input-otp

omg, how did I not find that? Yep, will do, thanks!

> Also the title and subtitle seem very specific and not quite correct... we should just copy what github shows as per pauls screenshot

I thought this would be less confusing as I'm not sure a lot of orderly users are going to get what the "device" terminology means. "Packit API login" was Rich's suggestion. As per David's suggestion, could maybe call it "Packit API Activation...?

EmmaLRussell:
ught this would be less confusing as I'm not sure a lot of orderly users are going to get what the "device" terminology means. "Packit API login" was Rich's suggestion. As per David's suggestion, could maybe call it "Packit API Activation...?


Oh sorry, that was your suggestion @absternator !

app/src/styles/globals.css
EmmaLRussell: Tailwind wasn't really up to the job here because of everything being buried in the verification input, so another great reason to ditch that.

plietar:
Tried this again and the redirect on login doesn't seem to work as it should:
- I am logged out of Packit
- I initiate a device flow using httr2
- It presents a link to http://localhost:3000/device and the user code
- I open the link in the browser, type in my username / password, hit enter
- I am redirected to the Packit frontpage and have nowhere to type in the code

To make it work I have to go to the original URL again to be able to type the code.

I have not tried GitHub based auth yet, but I suspect the issue is the same there.

EmmaLRussell:
Some issues still to fix with build but just to note:

- I've moved over to using zod/react-hook-form
- Also, using fetcher which is easy now that request content is content type json
- It deals with the code containing a hyphen but that not being available in the control (since the otp separator does not have a correponding character) - it removes hyphen from pasted code, ignores in entered key strokes and adds it on submit.
- This isn't that important, but it would be nice to have lower case letters entered automatically upper-cased in the visible form. I've got this happening on submit at the moment, but I couldn't really see an easy way to do it on entry and the value be reflected in the control. Can you think of a straightforward way of doing it? If not, I think it's fine just to transform on submit.

app/src/app/components/login/UserAuthForm.tsx
```diff
@@ -32,7 +32,7 @@ export const UserAuthForm = ({ className, ...props }: UserAuthFormProps) => {
         windowNavigate(loginRoute);
       }
     }
-  }, [user?.token]);
+  }, []);
```
EmmaLRussell, line 35: This change is to fix redirect on login, which was not working. The BasicAuthForm was redirecting to the requested redirect location on login but this component was picking up the user token change and also redirecting - to "/". I'm not sure if this a race condition that has recently appeared? I *think* this change is ok because the idea is just that anyone browsing to "/login" should get redirected to root immediately if they're already logged in - so this code should only run once, on component load. 

For other auth methods - both github and montagu auth methods should mean we get redirected back to /redirect on login, so they shouldn't be affected by this change.

EmmaLRussell:
Should be ready for re-review now...

app/src/styles/globals.css
```diff
@@ -117,3 +117,13 @@ iframe {
...
+  color: var(--foreground);
+  outline-color: var(--muted-foreground);
+  outline-width: thin;
+}
+
+.verification-input-inactive-char {
+  background-color: var(--muted);
+}
```
absternator, line 129: can these be deleted now? since we are using shadcn OTP component

absternator (commented):
Nice work!!!!! LGTM just left a few comments.. haven't approved yet because of failing test, just ping me when they are sorted

absternator (approved):
Niceeeee LGTM!!

PR #222, opened 2025-06-13
Mrc-6515 Add display name + email to manage users page user column

absternator:
The following PR adds display name and email to the manage users table in the users column. It also sorts out some other cleanups.
- Make the max height of rows to be h-20 over h-14. as h-14 was too small and was hard to scroll when there were lots of permissions. 
- Remove scroll area and just `overflow-auto max-h`, as this allows for variable height rows now, which is a nicer UI experience
- fix bug where scroll bar was not showing
- revert accidental base accent change

OLD:

image

NEW:

image


app/.eslintrc.json
```diff
@@ -21,13 +21,7 @@
     "max-len": [2, 120, 4],
     "eol-last": 2,
     "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/no-unused-vars": [
```
absternator, line 24: changed to warn as this makes a horrible dev experience as the page errors and doesn't load when editing and changing a file and you don't care about imports at that point

app/src/styles/globals.css
```diff
@@ -44,8 +44,8 @@
 
     --info: rgb(59 130 246); /* tailwind blue-500 */
 
-    --accent: hsl(228, 60%, 31%);
-    --accent-foreground: hsl(210, 40%, 96.1%);
+    --accent: hsl(210 40% 96.1%);
```
absternator, line 47: revert back to original.. assuming accidental change in branding pr

app/.eslintrc.json
absternator: we do get warnings when we compile and github will show warnings but wont fail pipelines.

david-mears-2: I see a single instance of that in the codebase - line 75 in packetColumns.tsx:
```ts
filterFn: (row, _columnId, filterValue) =>
```

app/src/app/components/contents/manageAccess/utils/manageRolesColumns.tsx
```diff
@@ -24,44 +24,46 @@ export const roleDefaultColumns = (scrollAreaHeightClass = "h-14") => [
...
+          className={cn(
+            "flex flex-wrap gap-1 italic text-xs pl-0.5 overflow-auto [&::-webkit-scrollbar]:w-2",
+            rowMaxHeight
+          )}
+        >
+          {users?.length === 0
+            ? "None"
+            : users.map((user, index) => (
+                <div key={user.id}>{index === users.length - 1 ? user.username : user.username + ", "}</div>
+              ))}
```
david-mears-2, line 37: do we need them in separate keyed divs? If not then this would be easier, and can also be applied to `setupManageRolesColumns` here, and (in manageUsersColumns.tsx) `columnHelper.accessor("roles"`

```suggestion
          {users?.length === 0
            ? "None"
            : users.map((u) => u.username).join(", ")}
```

app/src/styles/globals.css
david-mears-2: So actually this was deliberate - I want packit's default colours to be the same _kinds_ of colours as VIMC (or any other org) will use, that is, the accent colour should be on the dark side (and in real life it will probably be hued in line with the org's branding) and the foreground should be light. That means that we end up using the same colour 'logic' (I mean the variables play the same lightness/darkness roles) in development as in production instances. So the development/default colours should follow the guidelines for when organisations set custom colours (from packit-deploy's README):

> When choosing colours, consider contrast carefully, as described below. Do not rely solely on hue for contrast as this will create problems for users with certain types of colour blindness.

> The configurable colours are:

> Accent-colour. This may be used as the background for buttons or borders, and is expected to contrast in lightness with white (the background colour of the app in light mode), and with the accent-foreground colour (next bullet point). As such it should be middlingly dark or darker.
> Accent-foreground. This will be used as the text colour for anything whose background is the accent-colour, and as such is expected to contrast in lightness with the accent-colour (so it should be rather light, or simply white).
> Dark accent-colour. If configured, this plays the role of the accent-colour as above when the app is used in dark mode. As such it is expected to contrast in lightness with black (the background colour of the app in dark mode) and with the dark accent-foreground colour. It should thus be somewhat lighter than black.
> Dark accent-foreground. If configured, this plays the role of the accent-foreground as above when the app is used in dark mode. As such it should contrast in lightness with the dark accent-colour.

From `docs/custom-branding.md`:

> The default branding colours of Packit are different from the examples in the deploy tool, so that developers can easily see which colour is being used. The default colour values follow the expected brightness of the customisable colours: e.g. '--accent' is dark in colour and '--accent-foreground' is light in colour.

app/src/tests/components/contents/manageAccess/ManageUsers.test.tsx
```diff
@@ -42,6 +42,8 @@ describe("ManageUsers", () => {
     await waitFor(() => {
       mockUsersWithPermissions.forEach((user) => {
         expect(screen.getAllByText(user.username, { exact: false })[0]).toBeVisible();
+        expect(screen.getAllByText(user.email ?? "no email", { exact: false })[0]).toBeVisible();
+        expect(screen.getAllByText(user.displayName ?? "no display name", { exact: false })[0]).toBeVisible();
```
david-mears-2, line 46: This isn't quite right I think, because if there are multiple users with no email, but only one of those users actually displays the correct text "no email", all of those users will still pass.

So can we do something like, first find the div containing the right username for the user (or the right id in the key?), then check that the correct text exists within that particular div?

app/src/tests/components/contents/manageAccess/ManageUsers.test.tsx
```diff
@@ -103,7 +105,7 @@ describe("ManageUsers", () => {
       expect(filterInput).toHaveValue("");
     });
     mockUsersWithPermissions.forEach((user) => {
-      expect(screen.getByText(user.username)).toBeVisible();
+      expect(screen.getAllByText(user.username)[0]).toBeVisible();
```
david-mears-2, line 108: What's happening here - are there multiple users with the same username?

app/src/tests/components/contents/manageAccess/ManageUsers.test.tsx
david-mears-2: Also is the comment `    // only username roles are rendered` now outdated?

api/app/src/test/kotlin/packit/integration/controllers/UserRoleControllerTest.kt
```diff
@@ -90,6 +90,8 @@ class UserRoleControllerTest : IntegrationTest() {
         assertEquals(roles[0].name, "testRole1")
         assertEquals(roles[1].name, "testRole2")
         assertEquals(user.username, "test1@gmail.com")
+        assertEquals(user.displayName, "test user")
+        assertEquals(user.email, null)
```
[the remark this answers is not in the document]
absternator: haha yeah i downvoted it... ive added test for this as part of toUserWithPermissions test

absternator:
> Since we now display the username and the email address, and those seem to often be identical, the table could end up looking quite redundant. So would it make sense to omit the email address line if it's identical to the username?

this is only the case when we use basic auth... in montague auth or github they will not be the same.... and montague and github auth will be used mostly in production... thus makes sense to keep for overall consistency

david-mears-2 (approved):
LGTM

PR #225, opened 2025-06-18
Mrc-6548 Account for missing run status from orderly.runner

absternator:
This PR is twinned with orderly.runner https://github.com/mrc-ide/orderly.runner/pull/23. 

This now accounts for the fact that runner now returns statuses of only the taskIds that it finds. The reason for this is that the deployment of packit is taken down we lose these jobs, and packit requests these job IDs as its in the PostgreSQL database. We need to handle this case more gracefully!!! Thus packet will be able to just use database info on the missing redis task ids.

Testing:
Run up packit with dependencies. Run a packet and wait for it to finish. Then exec into the redis container and run `redis-cli del orderly.runner.queue:task:expr`. This will remove that task from queue.
Then check logs and logs page to see it still shows up as expected

### DIFF
--- a/scripts/common
+++ b/scripts/common
@@ -3,7 +3,7 @@ set -exu
 
 # Export env vars needed for running test dependencies
 export OUTPACK_SERVER_IMAGE=docker.io/mrcide/outpack_server:main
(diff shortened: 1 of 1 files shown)


M-Kusumgar (approved):
LGTM

PR #230, opened 2025-06-20
Fix holes in previewable file hover card content

david-mears-2:
The css class 'bg-card' was broken because the tailwind variable 'card' was typo'd.

# before

Screenshot from 2025-06-20 13-46-52

# after

Screenshot from 2025-06-20 13-46-42


absternator (approved):

david-mears-2:
thanks for the quick reponse mr @absternator - however docker is at its limit for today

PR #231, opened 2025-06-23
Mrc-6435 Hide tags on permissions form

absternator:
Hide tags when updating permissions. 

Testing: 
Create new role/user. Then update permission, and see there is no tag radio circle


app/src/tests/components/contents/manageAccess/updatePermissions/AddPermissionForUpdateForm.test.tsx
```diff
@@ -67,33 +66,34 @@ describe("AddPermissionForUpdateForm", () => {
     });
   });
 
-  it("should submit form with correct values for packet.read permission and tag scope", async () => {
```
absternator, line 70: the diff looks weird... its just commenting out 1 test and updating the other test to reference a packet rather than a tag

EmmaLRussell (approved):
Nice :sunglasses:

PR #232, opened 2025-06-23
mrc-6560 use gh registry;

EmmaLRussell:
Use docker github registry to avoid annoying rate limit errors...

Once this is merged, packit-deploy etc will be temporarily broken until we update those. I made a start on a packit-deploy branch targeting this branch, but it's a bit too painful to fix up in its entirety, with tests expecting "main" branch targets etc, so it's probably easier just to merge this first...


api/scripts/common
```diff
@@ -13,7 +13,8 @@ else
     GIT_BRANCH=$(git symbolic-ref --short HEAD)
```
M-Kusumgar, line 13: this wont do what you think it does, in a PR, the runner actually runs on a merge commit rather than the actual commit of the branch, see https://github.com/mrc-ide/packit/pkgs/container/packit not having the correct tags, this is what i had to do in hint, https://github.com/hivtools/hint/blob/eef7ed0a14c1e52378f21369b619e228b7e8e843/.github/actions/make-shared-env/action.yml#L39:
```
- id: ci-env
      name: Setup Environment
      shell: bash
      run: |
        if [ "${{github.event_name}}" = "pull_request" ];
        then
          long_sha=${{ github.event.pull_request.head.sha }}
          echo "CI_BRANCH=${{ github.head_ref }}" >> $GITHUB_OUTPUT
        else
          long_sha=${GITHUB_SHA}
          echo "CI_BRANCH=${{ github.ref_name }}" >> $GITHUB_OUTPUT
        fi
        echo "CI_SHA=${long_sha:0:7}" >> $GITHUB_OUTPUT
```

M-Kusumgar (changes requested):

M-Kusumgar (approved):
LGTM!

PR #233, opened 2025-06-24
mrc- 6559 resync packets

EmmaLRussell:
Resync packets endpoint and front end. User must have global packet manage permission to invoke this.

Resyncing is different from the usual incremental import of packets from outpack as it checks all packet ids, not just the ones that are new since the last import. This is an occasionally required function when outpack is updated from a different location, Any packets in outpack which are not in packit are added. Any packets in packit which are not in outpack are deleted. The numbers of packets added and deleted are logged. 

Resync can be invoked from a new "Resync Packets" tab in the Admin page (renamed from "Manage access") - we can make this "Manage Packets" if we ever add any other packet management functions here! There are some changes related to admin permissions since now some users may not have access to have all areas of the admin page  - if they have user.manage permission but not packet.manage or vice versa. You can test that this behaves properly by creating such a user and logging in as them.   

There are a lot of boring changes related to renaming the "manage-access" folder to "admin".


app/src/app/components/contents/admin/ManageRoles.tsx
```diff
@@ -10,10 +10,14 @@ import { setupManageRolesColumns } from "./utils/manageRolesColumns";
...
 
 import { NavLink } from "react-router-dom";
+import {Unauthorized} from "../common/Unauthorized";
 
 export const ManageRoles = () => {
   const { roles, users, mutate } = useManageAccessLayoutContext();
   const [filterValue, setFilterValue] = useState("");
+  if (!roles) {
+      return <Unauthorized />
+  }
```
EmmaLRussell, line 20: This, and corresponding addition in ManageUsers, is here because the parent AdminOutlet is now potentially accessible to users who have `packet.manage` permission (and can resync packets) but don't have `user.manage` permission. In this case the AdminOutlet does not attempt to get users and roles from the api. It's possible that such a user might manually navigate to `manage-roles` or `manage-users` in which case they will now see the Unauthorized screen rather than a blank page.

app/src/app/components/contents/admin/ResyncPackets.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const { authorities } = useUser();
+  if (!hasGlobalPacketManagePermission(authorities)) {
+    return <Unauthorized />;
+  } else {
+    return (
+      <>
+        <h2 className="text-2xl font-bold tracking-tight">Resync Packets</h2>
+        <p className="text-muted-foreground">
+          This operation can be used to synchronise Packit&apos;s data store with an outpack server which has been
+          pulled from another location. Resync may take a while if there are a lot of differences with outpack.
```
EmmaLRussell, line 16: Please suggest a better description!

app/src/app/components/contents/admin/ResyncPacketsButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+    setPending(false);
+  };
+
+  return (
+    <div className="">
+      <Button onClick={resyncPackets} type="button" className="w-[10rem]" disabled={pending}>
+        <RefreshCw className="p-1" /> Resync packets
+      </Button>
+      <div className={cn("text-sm pace-y-1 h-[1rem]", error ? "text-destructive" : "")}>
+        {pending ? "Resync in progress..." : error}
```
EmmaLRussell, line 32: This is pretty simplistic but probably ok if it's just us using it!

EmmaLRussell:
I've marked the service method as `@Transactional` as error was being thrown on uat on delete packet without this. Weirdly this didn't happen locally when there was a packet to delete - I'm not sure why not as I didn't think there was any difference in the spring/jpa configuration. @absternator any ideas? I also put the annotation on the scheduled import method as well as it seemed reasonable to do it there too.

app/src/app/components/contents/admin/AdminLayout.tsx
```diff
@@ -0,0 +1,31 @@
...
+        title: "Manage Roles"
+      },
+      {
+        to: "/manage-users",
+        title: "Manage Users"
+      }
+    );
+  }
+
+  if (hasGlobalPacketManagePermission(authorities)) {
```
absternator, line 23:
```suggestion
  if (canManageAllPackets(authorities)) {
```

app/src/app/components/contents/admin/AdminOutlet.tsx
```diff
@@ -7,9 +7,17 @@ import { UserWithPermissions } from "./types/UserWithPermissions";
...
+import { hasUserManagePermission } from "../../../../lib/auth/hasPermission";
 
-export const ManageAccessOutlet = () => {
-  const { roles, users, isLoading, error, mutate } = useGetRolesAndUsersWithPermissions();
+interface AdminOutletProps {
+  authorities: string[];
+}
+
+export const AdminOutlet = ({ authorities }: AdminOutletProps) => {
+  const isUserManager = hasUserManagePermission(authorities);
```
absternator, line 17:
```suggestion
  const hasUserManage = hasUserManagePermission(authorities);
```

absternator (commented):
Nice work!!! just some comments + questions

- maybe in person you can explain again when we would resync (sorry I'm being stupid)

api/app/src/main/kotlin/packit/service/PacketService.kt
```diff
@@ -67,9 +110,7 @@ class BasePacketService(
     }
 
     internal fun saveUniquePacketGroups(packetGroupNames: List<String>) {
-        val matchedPacketGroupNames = packetGroupRepository.findByNameIn(packetGroupNames).map { it.name }
```
absternator, line 70: reason for change?  also should be fetching all packetgroup names at once rather in a loop

EmmaLRussell: An opportunistic hack, and an unsatisfactory bug fix late in the day on Friday. Not really related to the resync, it was a problem with the Scheduler not being able to run to import new packets because the `findByNameIn` method has been made authorized and the Scheduler didn't have a user context. Looks like a change you made in late april as part of moving permission checking to the back end. #

But actually, it looks like nothing else is using that method now, so maybe we can just take the authz annotation off.

app/src/app/components/contents/admin/AdminOutlet.tsx
absternator: yup makes most sense

api/app/src/main/kotlin/packit/service/PacketService.kt
absternator: looked at 2 options... first may be easier, second may be better long term
1. make new repository method that says its unrestricted & use that in `saveUniquePacketGroups`
```
    @Query("SELECT pg FROM PacketGroup pg WHERE pg.name IN :names")
    fun findByNameInUnrestricted(names: List<String>): List<PacketGroup>
 ```
2. Update scheduler to run with auth context with all permissions
```
package packit.service

import org.slf4j.LoggerFactory
import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
import org.springframework.scheduling.annotation.Scheduled
import org.springframework.security.authentication.UsernamePasswordAuthenticationToken
import org.springframework.security.core.authority.SimpleGrantedAuthority
import org.springframework.security.core.context.SecurityContextHolder
import org.springframework.stereotype.Service

@Service
@ConditionalOnProperty(value = ["packit.scheduling.enabled"], havingValue = "true", matchIfMissing = true)
class Scheduler(
    private val oneTimeTokenService: OneTimeTokenService,
    private val packetService: PacketService,
    private val outpackServerClient: OutpackServerClient,
    private val deviceAuthRequestService: DeviceAuthRequestService
) {

    @Scheduled(fixedDelay = 10000, initialDelay = 0)
    fun checkPackets() = withSystemAuth {
        val current = packetService.getChecksum()
        val new = outpackServerClient.getChecksum()
        if (current != new) {
            log.info("Refreshing packet metadata")
            packetService.importPackets()
        }
    }

    @Scheduled(cron = "@daily")
    fun cleanUpExpiredTokens() = withSystemAuth {
        log.info("Cleaning up expired tokens")
        oneTimeTokenService.cleanUpExpiredTokens()
    }

    @Scheduled(cron = "@daily")
    fun cleanUpExpiredDeviceAuthRequests() {
        log.info("Cleaning up expired device auth requests")
        deviceAuthRequestService.cleanUpExpiredRequests()
    }

    internal fun <T> withSystemAuth(block: () -> T): T {
        val auth = UsernamePasswordAuthenticationToken(
            "system", null, listOf(
                SimpleGrantedAuthority("user.manage"),
                SimpleGrantedAuthority("packet.manage"),
                SimpleGrantedAuthority("packet.run"),
                SimpleGrantedAuthority("outpack.read"),
                SimpleGrantedAuthority("outpack.write"),
            )
        )
        val context = SecurityContextHolder.createEmptyContext()
        context.authentication = auth
        SecurityContextHolder.setContext(context)
        return try {
            block()
        } finally {
            SecurityContextHolder.clearContext()
        }
    }

    companion object {
        private val log = LoggerFactory.getLogger(Scheduler::class.java)
    }
}
```

**Comment by absternator** [comment]
also may want to pull in main

File: api/app/src/main/kotlin/packit/service/PacketService.kt [hunk 7]
```diff
@@ -46,19 +51,55 @@ interface PacketService {
...
+        runInfoRepository.deleteAllByPacketIdIn(notInOutpack)
+        packetRepository.deleteAllByIdIn(notInOutpack)
+
+        val notInPackit = outpackPacketIds subtract packitPacketIds
+        log.info("Saving ${notInPackit.size} new packets")
+        val newPackets = outpackPackets.filterKeys { it in notInPackit }.values
+        savePackets(newPackets)
+
+        // Clean up any packet groups which have been left childless
+        val currentPacketGroups = outpackPackets.values.map { it.name }.distinct()
```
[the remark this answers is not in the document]
EmmaLRussell: `currentPacketGroups` will be all the packet groups that outpack knows about, yes, so we shouldn't have any others in the packet group repository. 
Your suggestions wouldn't quite work I don't think: `packetGroupRepository.deleteAllByNameNotIn(notInOutpack.map { it.name } )` will remove groups not included as packet names in the "notInOutpack" list - i.e. packets that should be removed - so it's kind of the wrong way round. And even if we took out the "Not" it wouldn't be quite right, as we might still have outpack packets in the groups which included non-outpack packets, in which case we wouldn't want to delete them.

EmmaLRussell:
Thanks for checking out the Scheduler change required! I think this was not an issue when run locally, and not an issue the first time it ran on uat, but I've no idea why ! I'll put that change in and try testing it on uat.

absternator:
> think this was not an issue when run locally, and not an issue the first time it ran on uat, but I've no idea why ! I'll put that change in and try testing it on uat

I think because the first time it runs the database is empty and thus there is no packetgroups to fetch.... but I noticed I was getting the auth issue locally whenever I ran a packet (creating a new one).. this would then cause importPackets to run

EmmaLRussell:
> > think this was not an issue when run locally, and not an issue the first time it ran on uat, but I've no idea why ! I'll put that change in and try testing it on uat
> 
> I think because the first time it runs the database is empty and thus there is no packetgroups to fetch.... but I noticed I was getting the auth issue locally whenever I ran a packet (creating a new one).. this would then cause importPackets to run

Ah yes, that's right - and so the error was actually because it tried to insert duplicates because it didn't think they already exist yet, because the authz had hidden them...

app/src/app/components/contents/admin/AdminLayout.tsx
EmmaLRussell: Does this need bumping? Already mentioned in the below comment albeit I decided to leave it as "hasGlobalPacketManagePermission" rather than "hasGlobalPacketManage" to retain consistency with "hasUserManagePermission" in the same file. Could slice "Permission" off both if you prefer the shorter version?

EmmaLRussell:
I've put that Scheduler change in, and reinstate the authz on that repo method. Resync tested on uat - all seems to work ok.

absternator (approved):
thanks for sorting all that!! sorry again for vite merge headache

EmmaLRussell:
> thanks for sorting all that!! sorry again for vite merge headache

No worries! 😆

PR #236, opened 2025-06-26
Mrc-65470 Get read permissions per packet group

absternator:
Issue on UAT where too much data was being loaded on the home page. This was due to the read-permissions role and user info, as I had naively fetched for all packet groups.

The new update now only fetches roles and users per packet group only when the button is clicked to show the dialogue.

Testing:
Open home page and network tab on the side. See that when you click the update read permission button, a request is made that just gets roles and users for that packet group

image


plietar (approved):
This makes me so happy

PR #237, opened 2025-06-26
Mrc-6570 Generic API Handling

absternator:
This PR adds a fallback exception handler which will catch any `Exception::class` that are missed by all other exception handlers. The current `NoHandlerFoundException::class` was not working as expected with `FilterChainExceptionHandler`, thus has been replaced

Also have added logging with an id to that exception handler and this can be used to search up logs with that id for us.

It was not to set up integration test without adding a endpoint which throws an unexpected error, thus have left out.

Testing:
Update any controller code eg `getPacketGroupSummaries` on home page, and just add manual throw of random exception not caught by exception handler, e.g (SQLException, UnsupportedOperationException etc).... Then run app and check network requests of web. You will see generic error message with error code

### DIFF
--- a/api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
+++ b/api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
@@ -1,5 +1,6 @@
 package packit.exceptions
 
+import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.http.converter.HttpMessageNotReadableException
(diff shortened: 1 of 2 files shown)


api/app/src/main/kotlin/packit/exceptions/PackitExceptionHandler.kt
```diff
@@ -8,40 +9,62 @@ import org.springframework.security.authentication.InternalAuthenticationService
...
+    fun fallbackExceptionHandler(e: Exception): ResponseEntity<String> {
+        val errorId = UUID.randomUUID().toString()
+        log.error("ErrorId: $errorId - ${e.message}", e)
+        val message = "An unexpected error occurred. Please contact support with Error ID: $errorId"
+
+        return ErrorDetail(HttpStatus.INTERNAL_SERVER_ERROR, message)
+            .toResponseEntity()
+    }
+
+    @ExceptionHandler(MissingServletRequestParameterException::class)
```
absternator, line 42: these handlers were also missing

EmmaLRussell (approved):
This looks great, thanks. I wonder if the error code could maybe be a little shorter - maybe 8 or 12 characters like the device flow user codes. Probably doesn't need to be the full GUID. But probably doesn't really matter!

PR #240, opened 2025-07-04
Iterate pins feature: pin and unpin using UI

david-mears-2:
# TODO before merge:

- [x] Find the two instances of this comment and do it
```ts
// TODO: Update HttpStatus.BadRequest to HttpStatus.NotFound after https://github.com/mrc-ide/packit/pull/243 merged
```

# PR description

As it says on the tin, you can now pin packets and unpin them using an admin page. At this early stage in human history, pins refer to individual packets (by id) rather than by a query for a packet.

You should be prevented from pinning the same packet twice, and you should be prevented from pinning a non-existent packet.

Your user will need the packet.manage (global) permission.

The component structure of 'a button triggers a modal with form to open' is imitating the neighbouring Manage Users and Manage Roles pages' buttons and forms. I used the pattern twice, once for adding and once for removing pins.

There's another component, PinCheck, that is responsible for displaying helpful messages in the add pins form:
* No packet found with that ID
* Packet already pinned
* Matching packet found with name xyz (I thought that would be helpful since someone might not be sure they typed in the right id)


app/src/app/components/main/hooks/useGetPacketById.ts
```diff
@@ -2,10 +2,11 @@ import useSWR from "swr";
 import appConfig from "@config/appConfig";
 import { fetcher } from "@lib/fetch";
 import { PacketMetadata } from "@/types";
+import { packetIdRegex } from "@components/contents/admin/pins/utils/constants";
 
 export const useGetPacketById = (packetId: string | undefined) => {
   const { data, isLoading, error } = useSWR<PacketMetadata>(
-    packetId ? `${appConfig.apiUrl()}/packets/${packetId}` : null,
+    packetId && packetIdRegex.test(packetId) ? `${appConfig.apiUrl()}/packets/${packetId}` : null,
```
david-mears-2, line 9: I added this check here because otherwise every letter typed into the packet id input in the AddPinForm triggers a new request checking that there exists a packet with that id - but there never would be if the id isn't even well-formed, so we block those off at the pass.

api/app/src/main/kotlin/packit/controllers/PinController.kt
```diff
@@ -16,4 +23,25 @@ class PinController(
     fun getPinnedPackets(): ResponseEntity<List<PacketMetadata>> {
         return ResponseEntity.ok(pinService.findAllPinnedPackets())
     }
+
+    @PreAuthorize("@authz.hasGlobalPacketManagePermission(#root)")
```
absternator, line 27: should  be `canManageAllPackets`, :smile:

api/app/src/main/kotlin/packit/controllers/PinController.kt
```diff
@@ -16,4 +23,25 @@ class PinController(
...
+    fun pinPacket(
+        @RequestBody @Validated packetPin: PinDto,
+    ): ResponseEntity<PinDto> {
+        val packetId = packetPin.packetId
+        val pin = pinService.createPinByPacketId(packetId)
+
+        return ResponseEntity(PinDto(pin.packetId), HttpStatus.CREATED)
+    }
+
+    @PreAuthorize("@authz.hasGlobalPacketManagePermission(#root)")
```
absternator, line 38: should  be `canManageAllPackets`, :smile: .. as that will allow user.manage to hit as well... we assume this in other places as well... since user.manage can add/remove any permission

api/app/src/main/kotlin/packit/controllers/PinController.kt
```diff
@@ -16,4 +23,25 @@ class PinController(
...
+    @DeleteMapping()
+    fun deletePacket(
```
absternator, line 40: should this be `/{pinId}`.. then follow on and delete by pinid rather than packetId?

api/app/src/main/kotlin/packit/service/PinService.kt
```diff
@@ -20,4 +25,24 @@ class BasePinService(
             packetService.getMetadataBy(pin.packetId)
         }.sortedByDescending { it.time.start }
     }
+
+    override fun createPinByPacketId(packetId: String): Pin {
```
absternator, line 29: Just realised since pin is for a specific packet... the DB should have a foreign key to packet? And so we can point to the packet entity and reference... is there any reason for not doing that?

api/app/src/main/resources/errorBundle.properties
```diff
@@ -35,6 +35,8 @@ noRolesToUpdateWithPermission=No roles to update with permission
 notAllFilesFound=One or more file permissions not found
 packetGroupNotFound=Packet group not found
 packetNotFound=Packet not found
+pinAlreadyExists=Pin already exists
```
absternator, line 38: pinned packet? maybe reference to packet?

app/src/app/components/contents/admin/AdminLayout.tsx
```diff
@@ -21,10 +21,16 @@ export const AdminLayout = () => {
   }
 
   if (hasGlobalPacketManagePermission(authorities)) {
-    sidebarItems.push({
-      to: "/resync-packets",
-      title: "Resync Packets"
-    });
+    sidebarItems.push(
```
absternator, line 24: this pushing onto a array multiple times may be better extracted into own function.. `getAdminLayoutSidebarItems`

app/src/app/components/contents/admin/AdminLayout.tsx
```diff
@@ -21,10 +21,16 @@ export const AdminLayout = () => {
   }
 
   if (hasGlobalPacketManagePermission(authorities)) {
```
absternator, line 23: `canManageAllPackets`

app/src/app/components/contents/admin/AdminOutlet.tsx
```diff
@@ -42,9 +42,9 @@ export const AdminOutlet = ({ authorities }: AdminOutletProps) => {
   return roles || !hasUserManage ? <Outlet context={{ roles, users, mutate }} /> : null;
 };
 
-interface ManageAccessLayoutContext {
+interface AdminOutletContext {
```
absternator, line 45: thansk for rename!!!

absternator (commented):
Nice work looks awesome and works great!!! Just some comments .

Just realised would It be nicer UI/UX experience to give select dropdown to help users pick the packet they want?  the manage permissions dialog does this when  you select a a packet ..... we can also avoid all the checking of the pin packet id

1. sidebar item above resync packets? as it will be more used that resync packets. 
2. Also rename to `Manage Packet Pins` idk just  a way to sneak in its for packets. same
 with button
3. a bit hard to see red in dark mode for deleting pin packet. Also a tooltip would be good (remove pin). another idea up to you is moving to flex end of box like it is for manage users/roles deleting for consistency
image

5. on home page the pin icon makes it seem its an interaction. maybe best to remove
image
6. issue with checking packetid (which is my fault il check it out). since it cant find packet in authorization it actually throws illegalargumentexception 400.... so maybe thinking best to just remove this validation for now? because since its single input form anyways and you click add you will know if its valid or not. sending the request already checks if pin is valid... even though the validation does seem nice and snazzy maybe best not to add until I fix the issue.
image
7. where pins are showing should show packet id as well. because otherwise if you have 2 packets of the same group they look the same 
image
7. button should be destructive(red) IMO...... also title says delete but subtext says delete. maybe good to be consistent (remove IMO makes more sense)
image

api/app/src/main/kotlin/packit/controllers/PinController.kt
david-mears-2: I did try that at first, but it resulted in overly much overhead of passing pin ids around, so I reverted to using packet ids for all endpoints. The resulting consistency simplified the code.

api/app/src/main/resources/errorBundle.properties
david-mears-2: Changing to 'packet already pinned'

david-mears-2:
> Just realised would It be nicer UI/UX experience to give select dropdown to help users pick the packet they want? the manage permissions dialog does this when you select a a packet ..... we can also avoid all the checking of the pin packet id

Thinking about this again, I'm not sure it actually helps, because the number of packets to scroll through, even with a fairly small Packit instance, would be quite overwhelming.

david-mears-2:
> 1. sidebar item above resync packets? as it will be more used that resync packets.

done

> 2. Also rename to `Manage Packet Pins` idk just  a way to sneak in its for packets. same
>    with button

done

> 3. a bit hard to see red in dark mode for deleting pin packet. Also a tooltip would be good (remove pin). another idea up to you is moving to flex end of box like it is for manage users/roles deleting for consistency

done

> 5. on home page the pin icon makes it seem its an interaction. maybe best to remove

moved to be next to heading 'Pinned packets'

> 6. issue with checking packetid (which is my fault il check it out). since it cant find packet in authorization it actually throws illegalargumentexception 400.... so maybe thinking best to just remove this validation for now? because since its single input form anyways and you click add you will know if its valid or not. sending the request already checks if pin is valid... even though the validation does seem nice and snazzy maybe best not to add until I fix the issue.

> 7. where pins are showing should show packet id as well. because otherwise if you have 2 packets of the same group they look the same

Now the home page will show the packet IDs if any of the pinned packets belong to the same group.

> 7. button should be destructive(red) IMO...... also title says delete but subtext says delete. maybe good to be consistent (remove IMO makes more sense)
 
 done

api/app/src/main/kotlin/packit/service/PinService.kt
absternator: would that pin by query pin what? multiple packets? how would that work/look?

absternator:
> > Just realised would It be nicer UI/UX experience to give select dropdown to help users pick the packet they want? the manage permissions dialog does this when you select a a packet ..... we can also avoid all the checking of the pin packet id
> 
> Thinking about this again, I'm not sure it actually helps, because the number of packets to scroll through, even with a fairly small Packit instance, would be quite overwhelming.

like in permissions form its a combox... i.e you can search and select.. b

api/app/src/main/kotlin/packit/service/PinService.kt
david-mears-2: It would refer to a specific packet, but not by id. For example, the latest packet in some group (which would change when a new packet is added to the group). https://mrc-ide.github.io/orderly2/articles/query.html

absternator (approved):
thanks so much!!! lgtm

PR #245, opened 2025-09-02
mrc-6722 Support zip artefacts without descriptions in download page 

EmmaLRussell:
This fixes a bug Katy found, where an artefact with multiple files but no description caused packit's downoads page to fail on load. This is because the rendering code was attempting to substring a null description (in order to name the zip file download) and throwing an error. 

I've put in a change to check for falsy descriptions, and assign an artefact name based on artefact index instead. This is used both to name the zip file and as text for the artefact header. 

In order to test the fix, there is a new report in the demo dataset, `artefact-no-description`. There are two packets - the source for the later one sets `description = NULL` for the artefact, the earlier one does not provide description at all. I believe the later case translates to getting an empty string in the front end (which loaded the page ok, but with an empty heading for the artefact). The earlier one gets a javascript null description, which caused the page load to fail. These should both now load page and download the artefact with "artefact_0" as the artefact name.


app/src/tests/components/contents/downloads/orderly/Artefacts.test.tsx
```diff
@@ -32,7 +32,7 @@ const renderComponent = (packet?: PacketMetadata) => {
           <Route element={<Outlet context={{ packet }} />}>
             <Route
               path="/:packetName/:packetId/downloads"
-              element={<Artefacts artefacts={mockPacket.custom?.orderly.artefacts as Artefact[]} />}
+              element={<Artefacts artefacts={packet?.custom?.orderly.artefacts as Artefact[]} />}
```
david-mears-2, line 35: I assume the same correction applies a few lines back on 30, too.

app/src/tests/components/contents/downloads/orderly/Artefacts.test.tsx
```diff
@@ -93,4 +92,49 @@ describe("Artefacts component", () => {
...
+          ]
+        }
+      }
+    };
+    renderComponent(packetWithNamelessArtefacts);
+
+    expect(await screen.findByText("artefact_0")).toBeVisible();
+    expect(await screen.findByText("artefact_1")).toBeVisible();
+
+    // Renders download button for mult-file artefact
```
david-mears-2, line 128:
```suggestion
    // Renders download button for multi-file artefact
```

app/src/tests/components/contents/downloads/orderly/Artefacts.test.tsx
```diff
@@ -93,4 +92,49 @@ describe("Artefacts component", () => {
...
+    const artefactGroupDownloadButtonMatcher = /Download \(\d+\.\d+ KB\)/;
```
david-mears-2, line 129: This const is shared on line 78/79, could it be lifted up out of local scope?

david-mears-2 (commented):
Thanks for fixing it. Is zero-indexing the most user friendly? I say no

EmmaLRussell:
> Is zero-indexing the most user-friendly choice of file name? I say no

:laughing: Friendliness can take so many forms, sometimes it's a gentle encouragement to embrace zero-indexing...
(Yes, fair point, will change to 1-indexing).

app/src/tests/components/contents/downloads/orderly/Artefacts.test.tsx
EmmaLRussell: Yeah, although slightly complicated by the fact that `packet` might be undefined so you still need to give the router something to route with in that case. I'm also going to change the "when packet is not found.." test to check absence of generic /Download/ text rather than a specific artefact name.

david-mears-2 (approved):
Grand

PR #253, opened 2025-11-25
mrc-6789 Fix Download All failure with many files

EmmaLRussell:
Download All is failing on production for some packets with lots of downloadable files (see ticket for details). 
I've reproduced the issue in [associated base PR ](https://github.com/mrc-ide/packit/pull/252) which just adds a new demo report "many-downloads" with a single packet. If running locally, you should find that Download All for the packet in that branch fails, but that it succeeds in this branch, which fixes the problem. 

I've also deployed this branch to UAT, where a packet which previously would not Download All now succeeds: https://uat.montagu.dide.ic.ac.uk/packit/paper-four-figures/20251003-132910-dcadb7b4/downloads
(There seems to be an issue in this report with pngs being misnamed as pdfs - but I don't think that's anything to do with us!)

The issue was that all the filenames were being included in the query string of requests, both to get the one time token for the download, and also to do the download itself. So with enough filenames, the maximum url length was being exceeded. The fix for the two endpoints are slightly different:
- `/token` request: include all filenames in POST body rather than in url qs
- `/zip` request: this needs to be a GET because the download works by injecting an `<a>` tag with the url as its `href`. So I've removed the `paths` from this request entirely. We've saved the paths associated with the one time token, so we can assume that any requested zip download for the token should include all those files.

Arguably, we could take the same approach for the single file download, and not require the file path to be included in that request either as it will also be saved for its OTT. But I've left that request as is - it's unlikely we'll see a single filename which will be long enough  to cause this issue.


demos/orderly/src/many-downloads/many-downloads.R
```diff
@@ -0,0 +1,5 @@
+for (i in 1:200) {
+  file <- paste0("A-really-stupidly-lengthily-named-file-", i, ".rds")
+  saveRDS(list(a = 1, b = 2, c = 3), file)
+  orderly::orderly_artefact(description=paste("File", i), file)
+}
```
david-mears-2: The mrc-6789-repro branch at https://github.com/mrc-ide/packit/pull/252/ said not to merge it, but you're including the changes here anyway. That seems OK to me because it's helpful to have the bug repro case on main, but I'm not sure why https://github.com/mrc-ide/packit/pull/252/ said not to merge it.

david-mears-2 (approved):
Nice solution. Thanks for the reproduction case - all works as expected for me.

demos/orderly/src/many-downloads/many-downloads.R
EmmaLRussell: Oh yeah, sorry I just meant because the changes are in this PR too so there's not really any point in merging the other PR, which I intended to just close.

PR #255, opened 2026-02-27
Don't use browser autocomplete when adding users and roles

david-mears-2:
- Don't suggest role names using browser autocomplete

These will suggest the user add their own name as the name of the role.

- Don't suggest new users' name/email/password from browser autocomplete

The new user is not going to be the same person as the browser user, so we can't autocomplete their name/email address/password.

To actually switch off, we need to use the autocomplete attribute and set it to `"off"`


absternator (approved):
nice quality of life change

PR #257, opened 2026-03-10
Show R packages installed on runner

david-mears-2:
Using the new orderly.runner `/library/list` endpoint, this adds a 'Packages' tab to the 'Runner' page, which just lists the packages orderly.runner tells packit about.

## Manual testing

By default, when you run `scripts/dev-start`, the orderly runner image will look in the new `./fixtures/R` directory for its reporting of R libraries. This is different from the path given in [production](https://github.com/mrc-ide/orderly.runner/blob/253121d9c1120637a16bd791469fca15b2bca06c/docker/Dockerfile#L24) (`/library`), since it's unlikely this will turn up any packages in development as it likely won't exist.

If you want to look in a different location for R libraries (perhaps one with more packages), you can pass the path to your R library of choice using the --r-lib-path option when running dependencies, e.g.
```sh
./scripts/dev-start --super-user --db-build-skip --r-lib-path $HOME/R/x86_64-pc-linux-gnu-library/4.4
```
or:
```sh
PACKIT_HOST_R_LIBRARY_PATH=$HOME/R/x86_64-pc-linux-gnu-library/4.4 ./scripts/run-dependencies
```

And you may ask yourself, 'does this definitely work in an actual environment, looking at a non-fixture R library?' in which case I would direct you to look at uat where this branch has been deployed: https://uat.montagu.dide.ic.ac.uk/packit/runner/packages


david-mears-2:
@copilot+claude-opus-4.6 open a new pull request to apply changes based on [this feedback](https://github.com/mrc-ide/packit/pull/257#discussion_r2913795144) and [this feedback](https://github.com/mrc-ide/packit/pull/257#discussion_r2913795117) and [this feedback](https://github.com/mrc-ide/packit/pull/257#discussion_r2913795081)

app/src/app/components/contents/runner/PacketRunnerLayout.tsx
```diff
@@ -1,26 +1,41 @@
...
+      title: "Logs"
+    },
+    ...(hasGlobalPacketManagePermission(authorities)
+      ? [
+          {
+            to: "/runner/packages",
+            title: "Packages"
+          }
+        ]
+      : [])
```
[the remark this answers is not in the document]
david-mears-2: Updated the back end permissions to be stricter and match front end.

david-mears-2:
Copilot's PR comments were good and valuable feedback, but the PR implementing those suggestions had some differences from how I would have done it, so I instead implemented those myself

david-mears-2:
Closing this because the overall feature of installing packages would introduce a lot of complexity across the packit/orderly ecosystem, for little perceived gain to scientists.

Edit: reopened because this 'vertical slice' of the feature may have value for non-VIMC users of packit who don't have a yml-based workflow for R packages

api/app/src/main/kotlin/packit/config/RunnerConfig.kt
```diff
@@ -3,6 +3,7 @@ package packit.config
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.context.properties.ConfigurationProperties
 
+// NB 'repository' here refers to a git repository.
```
david-mears-2:
```suggestion
// 'Repository' here refers to a git repository, not a database repository.
```

app/e2e/runnerPage.demo.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    await page.goto("./");
+    content = await getContentLocator(page);
+    await page.getByRole("link", { name: "Runner" }).click();
+  });
+
+  test.describe("Packages tab", { tag: TAG_DEMO_PACKETS }, () => {
+    // Expect the example R package to be listed as installed; this package will not be present on
+    // demo or prod environments, nor if PACKIT_HOST_R_LIBRARY_PATH was set to anything other than
+    // the `./fixtures/R` path when running dependencies.
+    // See api/README.md for more details.
```
EmmaLRussell, line 18: A while ago I introduced the `@demoPackets` tag which was intended for tests running against local servers only, with the demo packet set. It seems like this would be a good candidate for this pattern too, although "demo packets" would then become a misleading tag since it would apply to more of the environment than just the packets. 

This is why some fixtures have both "demo" and "generic" versions. You'll see that the demo ones have the demo packets tag, and are deliberately excluded from the `test:e2e-dev` and `test:e2e-prod` scripts.

So I think at minimum you could move this test into `runnerPage.demo.spec` and give it the demo packets tag. And in an ideal world we'd rename the tag to something more inclusive (`@demoContext` or `@demoSystem`?)

app/src/app/components/contents/runner/PacketRunnerPackages.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <>
+      <div>
+        <h2 className="text-2xl font-bold tracking-tight">Package versions</h2>
+        <p className="text-muted-foreground">View installed R packages and versions</p>
+      </div>
+      <div className="space-y-4 mt-4">
+        <p>
+          This list includes only the packages installed in the R library exposed to the runner, excluding for instance
+          those installed on the runner itself.
```
EmmaLRussell: I'm not sure this is going to be all that comprehensible to the science team. How about something a little more general?

```suggestion
          This list includes only the additional packages installed in the runner's R library for use by reports, and excludes some standard packages. 
```

EmmaLRussell (commented):
Looks great! Code looks good. 

"fixtures" confused me a little as a top level folder name (though the README explanation is extremely helpful) - do you mean just as in test fixtures? I might be inclined to call that something a little more specific like "runnerDemoLib". Maybe even include it under the top level "demos" folder, if that won't collide with anything else? 

> This is different from the path given in [production](https://github.com/mrc-ide/orderly.runner/blob/253121d9c1120637a16bd791469fca15b2bca06c/docker/Dockerfile#L24) (/library), since it's unlikely this will turn up any packages in development as it likely won't exist.

Do you mean because there wont be any custom runner packages installed post-deployment? So is it the case that there are standard packages which are available to reports, but which aren't listed as they're preinstalled on the workers? (I think we did discuss, this, sorry can't recall the details.) I wonder if it would be good to eventually list those as well so that people don't wrongly think they are missing.

david-mears-2:
> "fixtures" confused me a little as a top level folder name (though the README explanation is extremely helpful) - do you mean just as in test fixtures? I might be inclined to call that something a little more specific like "runnerDemoLib". Maybe even include it under the top level "demos" folder, if that won't collide with anything else?

Yes, I mean as in "test fixtures". Originally I nested it under `scripts/` and I can't remember why I thought that was not ideal. I think it would be confusing to mix this kind of demo in with the orderly demos under `demos/`. I can't really nest it under a tests folder since it's used by both the back end and front end tests. I think I'll put it back into `scripts` and call the folder `runnerDemoLib` as you suggest. 

> 
> > This is different from the path given in [production](https://github.com/mrc-ide/orderly.runner/blob/253121d9c1120637a16bd791469fca15b2bca06c/docker/Dockerfile#L24) (/library), since it's unlikely this will turn up any packages in development as it likely won't exist.
> 
> Do you mean because there wont be any custom runner packages installed post-deployment? So is it the case that there are standard packages which are available to reports, but which aren't listed as they're preinstalled on the workers? (I think we did discuss, this, sorry can't recall the details.) I wonder if it would be good to eventually list those as well so that people don't wrongly think they are missing.

Actually I just mean that, on your machine, there won't be any files or folders at the route `/library`, because your machine is not a dockerized/deployed environment where something is mounted to that address. Thus, there would be nothing listed on the packages page in development, unless we explicitly point your local (dev) orderly.runner server to some address where we know an R library can be found (whether the fixture or your real local library that you use).

Nevertheless in answer to your question, yes, there are standard packages available to reports but which are not listed - orderly.runner server (currently) only reports on the packages installed in the shared library (`installed.packages` takes an optional argument specifying the library location). It would be quite simple to start reporting all packages + their location of installation. I think technically the standard packages installed on the server (which is where we query `installed.packages()`) could diverge from those on a runner, which would potentially make reporting all the packages misleading, but that seems unlikely to happen in reality.

Here is a PR on orderly.runner for reporting all the packages regardless of location: https://github.com/mrc-ide/orderly.runner/pull/40 - I would need to change this front-end to follow that.

app/e2e/runnerPage.demo.spec.ts
david-mears-2: Hm, I could make this demo-only, but that would prevent it from running in CI.

> A while ago I introduced the @demoPackets tag which was intended for tests running against local servers only, with the demo packet set.

There is also a second goal of the @demoPackets tag, which, per the README, is testing against deployed dev instances:

> We want to have e2e tests which test all packit features, but also to have a suite of tests which we can run against any packit server to test that its deployment has succeeded. In order to support detailed tests of all features, we want to have some tests which expect our standard demo dataset (available on localhost and perhaps on some dev instances)

I could _add_ a demo-only test that tests for certain packages being listed as installed on dev environments, but I'm not sure there are any packages we can guarantee as being installed on any dev instance - until we start listing all installations, at which point we can check for a standard package.

app/src/app/components/contents/runner/PacketRunnerPackages.tsx
david-mears-2: The word 'additional' is helpful.

app/e2e/runnerPage.demo.spec.ts
david-mears-2: Just realised the point you're making is that the test I wrote will be picked up by any `npm run test:e2e-prod` which is not what we want.

But my test can't be expected to pass for either of test:e2e-dev or test:e2e-prod. Checking on https://reside-dev.packit-dev.dide.ic.ac.uk/runner/packages, there aren't any packages installed on the shared library volume. I could install some small R package there and then update the test to check for that. Does that sound OK?

Once the test is demo-only, I can remove `PACKIT_HOST_R_LIBRARY_PATH` from from the front-end github workflow.

EmmaLRussell: > Hm, I could make this demo-only, but that would prevent it from running in CI.
> 
> > A while ago I introduced the @demoPackets tag which was intended for tests running against local servers only, with the demo packet set.
> 
> There is also a second goal of the @demoPackets tag, which, per the README, is testing against deployed dev instances:
> 
> > We want to have e2e tests which test all packit features, but also to have a suite of tests which we can run against any packit server to test that its deployment has succeeded. In order to support detailed tests of all features, we want to have some tests which expect our standard demo dataset (available on localhost and perhaps on some dev instances)
> 

I think the "perhaps on some dev instances" part is causing more complications than it deserves to here. as we don't generally do this in practice (AFAIK). The @demoPackets tests are really just intended for local and CI testing. 

> Just realised the point you're making is that the test I wrote will be picked up by any npm run test:e2e-prod which is not what we want.

Yes, and I don't think we should necessarily be installing packages on the deployed instances for the sake of the e2e tests. 

How about just adding another tag, @demoRPackages or whatever, and applying that to your test, and explicitly excluding it from the test:e2e-dev and test:e2e-prod scripts. That way we can treat the two separately if we need to, and we don't need to be misleading about what each test implies.

EmmaLRussell:
> Actually I just mean that, on your machine, there won't be any files or folders at the route `/library`, because your machine is not a dockerized/deployed environment where something is mounted to that address. Thus, there would be nothing listed on the packages page in development, unless we explicitly point your local (dev) orderly.runner server to some address where we know an R library can be found (whether the fixture or your real local library that you use).

👍 

> 
> Nevertheless in answer to your question, yes, there are standard packages available to reports but which are not listed - orderly.runner server (currently) only reports on the packages installed in the shared library (`installed.packages` takes an optional argument specifying the library location). It would be quite simple to start reporting all packages + their location of installation. I think technically the standard packages installed on the server (which is where we query `installed.packages()`) could diverge from those on a runner, which would potentially make reporting all the packages misleading, but that seems unlikely to happen in reality.
> 
> Here is a PR on orderly.runner for reporting all the packages regardless of location: [mrc-ide/orderly.runner#40](https://github.com/mrc-ide/orderly.runner/pull/40) - I would need to change this front-end to follow that.

I think that would be a good addition (future branch), if it was made clear in the UI what the distinction was between "core packages" and "installed packages" (don't think users will be interested in the location as such).

app/e2e/runnerPage.demo.spec.ts
david-mears-2: OK, great. I've done the separate tag idea.

EmmaLRussell (approved):
LGTM 👍

PR #258, opened 2026-03-10
Don't use user.manage permission for packet management permission checks

david-mears-2:
`canManageAllPackets` is the name of functions that existed both in the front end and the back end, which allow users to manage packets when they have global user.manage permission (including when they don't have packet.manage).

We should keep packet management permissions separate from user management, so I have removed this function and replaced it with a reference to `hasGlobalPacketManagePermission`.

The exception to keeping the two scopes separate is when granting users packet-related permissions; that really is both user management and packet management. It looks like currently these actions are defended only by 'user.manage' (e.g. at the top of RoleController), i.e. through packit I can give someone (myself excluded) permissions for any packet, regardless of whether I have permissions for that packet. This seems OK but I wanted to check.

### DIFF
--- a/api/app/src/main/kotlin/packit/security/PermissionChecker.kt
+++ b/api/app/src/main/kotlin/packit/security/PermissionChecker.kt
@@ -10,7 +10,6 @@ interface PermissionChecker {
     fun hasGlobalReadPermission(authorities: List<String> = emptyList()): Boolean
 
     /** Manage packets */
-    fun canManageAllPackets(authorities: List<String> = emptyList()): Boolean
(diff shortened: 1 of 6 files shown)


absternator (commented):
The initial idea was to make `user.manage` kind of like a god permission. As people who can manage users can pretty much do anything in the application. Thus not having to add `packet.manage` or any other permissions to people who can manage users. As this probably the highest level permission.

The issue here is if they want to add permissions to a user with packets but dont have the manage permission then they cant add those permissions. Beacuse pretty sure they wont see the list of packets and packet groups anymore when updating permissions. This is not right: a user with `user.manage` should be able to add/remove any permission to a user without needing `packet.manage`.  See screenshot below with a new user with only user. manage

Image

PR #259, opened 2026-03-10
Sync Kotlin plugin versions

david-mears-2:
Claude pointed out that all of these plugins - which variously extend Kotlin or configure the language features - should have the same versions, for maximum compatibility. This because they all come from https://kotlinlang.org/docs/compiler-plugins-overview.html - the version numbers refer to the version of Kotlin we target. I upgraded the ones that are further behind to match the most recent (2.0.21). By adding a `pluginManagement` block to settings.gradle.kts, we can help these stay synced in the future.

Before merging I'd like to be sure that:

* packit API will still build nicely from scratch: for this, please test using:
```sh
cd api
./gradlew --stop
./gradlew --refresh-dependencies clean
./gradlew build
```
* This doesn't wreck your IDE's ability to read Kotlin / use an LSP (mine hasn't been able to at any stage this year) - please close and reopen your IDE after re-gradle-ing to ensure it's not gone to pot

### DIFF
--- a/api/app/build.gradle.kts
+++ b/api/app/build.gradle.kts
@@ -1,17 +1,18 @@
+// https://docs.gradle.org/8.6/userguide/building_java_projects.html
 import io.gitlab.arturbosch.detekt.Detekt
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     // Apply the org.jetbrains.kotlin.jvm Plugin to add support for Kotlin.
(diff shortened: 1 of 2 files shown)


EmmaLRussell (approved):
Works for me. The build task itself fails because it runs tests as a dependency (for some reason!) and there's one Login controller test which I think has always failed locally. But the api runs fine from gradle, and IDE still understands kotlin.

PR #262, opened 2026-03-25
If runner feature disabled, don't show 'runner' navlink or pages

david-mears-2:
Non-VIMC Packit instances have the option to be read-only, i.e. no runner (this would be configured by setting the environment variable orderly.runner.enabled). If there's no runner we shouldn't show a link to the runner in the nav menu, nor show the runner pages.

### DIFF
--- a/api/app/src/main/kotlin/packit/service/RunnerService.kt
+++ b/api/app/src/main/kotlin/packit/service/RunnerService.kt
@@ -14,6 +14,7 @@ import packit.model.dto.*
 import packit.repository.RunInfoRepository
 
 interface RunnerService {
+    fun getEnabled(): Boolean = true
     fun getVersion(): OrderlyRunnerVersion
     fun gitFetch()
     fun getBranches(): GitBranches
@@ -167,6 +168,7 @@ class DisabledRunnerService : RunnerService {
         throw PackitException("runnerDisabled", HttpStatus.FORBIDDEN)
     }
 
+    override fun getEnabled(): Boolean = false
     override fun getVersion(): OrderlyRunnerVersion = error()
     override fun gitFetch() = error()
     override fun getBranches(): GitBranches = error()
--- a/app/src/app/components/contents/runner/PacketRunnerLayout.tsx
+++ b/app/src/app/components/contents/runner/PacketRunnerLayout.tsx
@@ -4,6 +4,7 @@ import { SidebarItem } from "@lib/types/SidebarItem";
 import { useUser } from "../../providers/UserProvider";
 import { Sidebar } from "../common/Sidebar";
 import { Unauthorized } from "../common/Unauthorized";
+import { useGetRunnerEnabled } from "@components/header/hooks/useGetRunnerEnabled";
 
 const sidebarItems: SidebarItem[] = [
   {
@@ -18,7 +19,9 @@ const sidebarItems: SidebarItem[] = [
 
 export const PacketRunnerLayout = () => {
   const { authorities } = useUser();
-  if (!hasPacketRunPermission(authorities)) {
+  const { isRunnerEnabled } = useGetRunnerEnabled(hasPacketRunPermission(authorities));
+
+  if (!hasPacketRunPermission(authorities) || isRunnerEnabled === false) {
     return <Unauthorized />;
   }
   return (
--- a/app/src/app/components/contents/runner/PacketRunnerLayout.tsx
+++ b/app/src/app/components/contents/runner/PacketRunnerLayout.tsx
@@ -4,7 +4,7 @@ import { SidebarItem } from "@lib/types/SidebarItem";
 import { useUser } from "../../providers/UserProvider";
 import { Sidebar } from "../common/Sidebar";
 import { Unauthorized } from "../common/Unauthorized";
-import { useGetRunnerEnabled } from "@components/header/hooks/useGetRunnerEnabled";
+import { useRunnerConfig } from "../../providers/RunnerConfigProvider";
 
 const sidebarItems: SidebarItem[] = [
   {
@@ -19,7 +19,7 @@ const sidebarItems: SidebarItem[] = [
 
 export const PacketRunnerLayout = () => {
   const { authorities } = useUser();
-  const { isRunnerEnabled } = useGetRunnerEnabled(hasPacketRunPermission(authorities));
+  const isRunnerEnabled = useRunnerConfig();
 
   if (!hasPacketRunPermission(authorities) || isRunnerEnabled === false) {
     return <Unauthorized />;
--- a/app/src/app/components/header/NavMenu.tsx
+++ b/app/src/app/components/header/NavMenu.tsx
@@ -8,35 +8,30 @@ import { DropdownMenu, DropdownMenuTrigger, DropdownMenuContent, DropdownMenuIte
 import { Menu } from "lucide-react";
 import { NavLink } from "react-router-dom";
 import { Button, buttonVariants } from "../Base/Button";
-
+import { useGetRunnerEnabled } from "./hooks/useGetRunnerEnabled";
 interface NavMenuProps extends React.HTMLAttributes<HTMLElement> {
   authorities: string[];
 }
 export const NavMenu = ({ className, authorities, ...props }: NavMenuProps) => {
-  const NavItems: { [key: string]: string } = {
-    runner: "Runner"
-    // accessibility: "Accessibility",
-  };
+  const canRunPackets = hasPacketRunPermission(authorities);
+  const { isRunnerEnabled, isLoading: isRunnerEnabledLoading } = useGetRunnerEnabled(canRunPackets);
 
-  const displayableItems = Object.keys(NavItems).filter((to) => {
-    if (to === "runner" && !hasPacketRunPermission(authorities)) {
-      return false;
-    } else {
-      return true;
-    }
-  });
+  const NavItems: { [key: string]: string } = {};
+
+  if (canRunPackets && !isRunnerEnabledLoading && isRunnerEnabled) {
+    NavItems.runner = "Runner";
+  }
 
   // Special case: route "Admin" to appropriate tab depending on user perms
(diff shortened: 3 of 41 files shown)


api/app/src/main/kotlin/packit/service/RunnerService.kt
```diff
@@ -167,6 +168,7 @@ class DisabledRunnerService : RunnerService {
         throw PackitException("runnerDisabled", HttpStatus.FORBIDDEN)
     }
 
+    override fun getEnabled(): Boolean = false
```
absternator, line 171: should we not be getting from appconfig? and have this set in application.properties?

app/src/app/components/contents/runner/PacketRunnerLayout.tsx
```diff
@@ -18,7 +19,9 @@ const sidebarItems: SidebarItem[] = [
 
 export const PacketRunnerLayout = () => {
   const { authorities } = useUser();
-  if (!hasPacketRunPermission(authorities)) {
+  const { isRunnerEnabled } = useGetRunnerEnabled(hasPacketRunPermission(authorities));
+
+  if (!hasPacketRunPermission(authorities) || isRunnerEnabled === false) {
```
absternator, line 24:
```suggestion
  if (!hasPacketRunPermission(authorities) || !isRunnerEnabled ) {
```
should probs not show if cant get this config as well.


Also this is pretty static so we can probably just fetch once and store in local storage like we do with other instance related stuff. (i think auth type)

app/src/app/components/header/hooks/useGetRunnerEnabled.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { data, error, isLoading } = useSWR<boolean>(
+    shouldFetch ? `${appConfig.apiUrl()}/runner/enabled` : null,
+    (url: string) => fetcher({ url }),
+    {
+      revalidateOnFocus: false
+    }
+  );
+
+  return {
+    isRunnerEnabled: data === true,
```
absternator:
```suggestion
    isRunnerEnabled: data,
```

absternator (commented):
good change to make!! just a few comments

api/app/src/main/kotlin/packit/service/RunnerService.kt
david-mears-2: It's set in application.properties already, and I think [this line](https://github.com/mrc-ide/packit/blob/1ca780adc36c0977864af4b3d75095a49c4eee60/api/app/src/main/kotlin/packit/config/RunnerConfig.kt#L12) is what reads it (setting config), though it bypasses AppConfig.kt itself. Then [this line](https://github.com/mrc-ide/packit/blob/1ca780adc36c0977864af4b3d75095a49c4eee60/api/app/src/main/kotlin/packit/service/RunnerService.kt#L190) checks if config was set, returns RunnerService if it is, and if not it returns the DisabledRunnerService.

My changes just add a property to the RunnerService and DisabledRunnerService which is static (enabled -> true for RunnerService, enabled -> false for DisabledRunnerService)

app/src/app/components/contents/runner/PacketRunnerLayout.tsx
david-mears-2: @copilot+gpt-5.3-codex 

Refer to app/src/app/components/providers/hooks/useGetAuthConfig.ts, app/src/app/components/providers/AuthConfigProvider.tsx, app/src/lib/localStorageManager.ts. Branch off this PR to make the fetch to the `/runner/enabled` endpoint happen only once, since it's a static value.

david-mears-2: I'm actually not convinced this is the best idea, because what if the instance decides they want to change the setting from enabled to disabled or vice versa? The setting would be cached in localStorage, unable to change.

Maybe sessionStorage would be a better place than localStorage. You'd do one fetch per session rather than once per browser.

app/src/app/components/header/NavMenu.tsx
```diff
@@ -8,17 +8,16 @@ import { DropdownMenu, DropdownMenuTrigger, DropdownMenuContent, DropdownMenuIte
...
 }
 export const NavMenu = ({ className, authorities, ...props }: NavMenuProps) => {
-  const canRunPackets = hasPacketRunPermission(authorities);
-  const { isRunnerEnabled, isLoading: isRunnerEnabledLoading } = useGetRunnerEnabled(canRunPackets);
+  const isRunnerEnabled = useRunnerConfig();
 
   const NavItems: { [key: string]: string } = {};
 
-  if (canRunPackets && !isRunnerEnabledLoading && isRunnerEnabled) {
+  if (hasPacketRunPermission(authorities) && isRunnerEnabled) {
```
david-mears-2: Reinstate the isLoading check

app/src/app/components/header/hooks/useGetRunnerEnabled.ts
```diff
@@ -2,17 +2,17 @@ import useSWR from "swr";
...
+    runnerEnabled === null ? `${appConfig.apiUrl()}/runner/enabled` : null,
     (url: string) => fetcher({ url }),
     {
       revalidateOnFocus: false
     }
   );
 
   return {
-    isRunnerEnabled: data === true,
+    data,
```
david-mears-2: Reinstate explicit data val name 'isRunnerEnabled'

app/src/app/components/providers/RunnerConfigProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+    if (data !== undefined) {
+      setRunnerEnabled(data);
+      localStorage.setItem(LocalStorageKeys.RUNNER_CONFIG, JSON.stringify(data));
+    }
+  }, [data]);
+
+  if (error) return <ErrorComponent message="failed to load runner config" error={error} />;
+
+  return <RunnerConfigContext.Provider value={runnerEnabled}>{children}</RunnerConfigContext.Provider>;
+};
```
david-mears-2, line 30: I don't see a unit test on this new provider. I want to be sure the storage is correctly updated and read, in both possible states of whether runner is enabled.

app/src/lib/storageManager.ts
```diff
@@ -10,3 +10,7 @@ export const getAuthConfigFromLocalStorage = (): AuthConfig | null => {
   const authConfig = localStorage.getItem(LocalStorageKeys.AUTH_CONFIG);
   return authConfig ? JSON.parse(authConfig) : null;
 };
+export const getRunnerConfigFromLocalStorage = (): boolean | null => {
+  const runnerConfig = localStorage.getItem(LocalStorageKeys.RUNNER_CONFIG);
+  return runnerConfig !== null ? JSON.parse(runnerConfig) : null;
+};
```
david-mears-2, line 25: Runner config should be stored in session storage rather than local storage. Rename file to storageManager. Also rename LocalStorageKeys to StorageKeys.

david-mears-2:
@copilot+gpt-5.3-codex Address my comments

app/src/lib/storageManager.ts
david-mears-2: @copilot storageManager should be responsible for remembering which keys are local storage and which session. Thus all instances of sessionStorage.setItem and localStorage.setItem should be routed through setters in storageManager.

app/src/lib/storageManager.ts
```diff
@@ -6,11 +6,38 @@ export const getUserFromLocalStorage = (): UserState | null => {
...
+  localStorage.removeItem(StorageKeys.THEME);
+};
+export const getRequestedUrlFromLocalStorage = () => localStorage.getItem(StorageKeys.REQUESTED_URL);
+export const setRequestedUrlInLocalStorage = (url: string | null) => {
+  if (url === null) {
+    localStorage.removeItem(StorageKeys.REQUESTED_URL);
+  } else {
+    localStorage.setItem(StorageKeys.REQUESTED_URL, url);
+  }
+};
```
david-mears-2: You moved some logic into this file that makes this setter unlike the others. Keep the logic in the provider.

david-mears-2 (commented):
Updating package-lock.json prevents automated tests. Use git to revert this file to its state in the last human-authored commit.

david-mears-2:
@copilot+gpt-5.3-codex address comments

app/src/app/components/contents/runner/PacketRunnerLayout.tsx
david-mears-2: > Done in [d123fc8](https://github.com/mrc-ide/packit/commit/d123fc8fe99711a46d691e8f414498d10fe3ba2f). Created a `RunnerConfigProvider` following the same pattern as `AuthConfigProvider` — it fetches `/runner/enabled` once, caches the result in localStorage, and exposes it via a `useRunnerConfig()` context hook. `NavMenu` and `PacketRunnerLayout` now use the context directly instead of calling the fetch hook themselves.

This has since been changed to use sessionStorage instead of localStorage.

david-mears-2:
@copilot+gpt-5.3-codex Test coverage is missing for most of useGetRunnerEnabled. Please unit test this file for all three possibilities: runnerEnabled as true, false, or null.

david-mears-2:
The failing CI e2e tests alerted me to the fact that this new endpoint `/runner/enabled` was not reachable (it 401s) if you're not already logged in. This is fixed in the commit [730fd77](https://github.com/mrc-ide/packit/pull/262/commits/730fd7770bae4a57ad184cbe396199700c79393f).

api/app/src/main/kotlin/packit/service/RunnerService.kt
absternator: ohh thanks for the clarification. i got  a bit confused by the setup of it all

app/src/app/components/contents/runner/PacketRunnerLayout.tsx
absternator: good idea

absternator (approved):
thanks for adressing comments. lgtm!

PR #263, opened 2026-04-01
Distinguish core/installed R packages

david-mears-2:
Add information to the runner > packages page about installed packages that were not installed specifically in the shared runner library. Although that shared library is the intended storage location for all packages used in reports, other packages are nevertheless available by default (e.g. they come out of the box with R, let us call these 'core' packages) and are probably used by reports in practice sometimes. The idea behind listing both installed and core packages separately is to prevent people having incorrect mental models about what packages are available, hopefully reducing confusion without encouraging people to make their reports depend on un-reproducible / contingent properties of the runner environment.

Depends on https://github.com/mrc-ide/orderly.runner/pull/40 in orderly.runner and must only be deployed in tandem with that change.


scripts/common
```diff
@@ -3,7 +3,7 @@ set -exu
 
 # Export env vars needed for running test dependencies
 export OUTPACK_SERVER_IMAGE=docker.io/mrcide/outpack_server:main
-export ORDERLY_RUNNER_IMAGE=ghcr.io/mrc-ide/orderly.runner:main
+export ORDERLY_RUNNER_IMAGE=ghcr.io/mrc-ide/orderly.runner:library-list-report-all-packages
```
david-mears-2: Revert this after the orderly.runner PR is merged, and before this is merged.
```suggestion
export ORDERLY_RUNNER_IMAGE=ghcr.io/mrc-ide/orderly.runner:main
```

app/e2e/runnerPage.demo.spec.ts
```diff
@@ -1,22 +1,16 @@
...
-
   test.describe("Packages tab", { tag: TAG_DEMO_R_LIBRARY }, () => {
     // Expect the example R package to be listed as installed; this package will not be present on
     // demo or prod environments, nor if PACKIT_HOST_R_LIBRARY_PATH was set to anything other than
     // the `./scripts/runnerDemoLib` path when running dependencies.
     // See api/README.md for more details.
     test("can see list of installed packages", async ({ page }) => {
+      await page.goto("./");
+      const content = await getContentLocator(page);
+      await page.getByRole("link", { name: "Runner" }).click();
```
EmmaLRussell, line 13: Could add e2e check of additional info visible in page too.

EmmaLRussell (approved):
Looks good, just wondered if you wanted to include the change in e2e test.

PR #264, opened 2026-04-01
Add tool-guardian hook for Copilot

david-mears-2:
Copied wholesale from https://github.com/github/awesome-copilot/blob/main/hooks/tool-guardian
See the README there to understand more, but the purpose is extra guardrails around Copilot, for when running it more [autonomously](https://docs.github.com/en/copilot/concepts/agents/copilot-cli/autopilot).

I think the main thing I'd be worried about with yolo-ing (allowing copilot to run any command) is it messing up my computer or uncommitted work in irreparable ways. It could do this directly, by modifying/deleting files or uninstalling things, or indirectly by installing a malicious package. There is also a very specific risk of it posting private keys to github, which I don't address here.

My general approach thus far is to whitelist safe commands e.g. `ls`, `npx vitest` as they come up; but this is perhaps over-cautious and prevents a lot of the benefits of using AI agents, namely being able to delegate and not have to supervise (so that working in parallel is made possible). In order to feel safe doing that I'd like to do something like run copilot inside a Docker sandbox: https://docs.docker.com/ai/sandboxes/. (If using Docker sandbox I would choose the 'balanced' policy with regard to network access: https://docs.docker.com/ai/sandboxes/security/policy/#network-policies). Another way to sandbox is to run Copilot in the cloud through the GitHub website, but then it's less able to be interacted with. Further research (not yet read): https://www.reddit.com/r/devops/comments/1q4pvy6/wrote_a_deep_dive_on_sandboxing_for_ai_agents/

The bash script added in this PR is not me-authored, it's copied from that tool-guardian repo. That repo is trusted enough by GitHub to be linked in the Copilot docs.

Copilot hooks docs: https://docs.github.com/en/copilot/concepts/agents/coding-agent/about-hooks

https://docs.github.com/en/copilot/reference/hooks-configuration

### DIFF
--- a/.github/hooks/tool-guardian/README.md
+++ b/.github/hooks/tool-guardian/README.md
@@ -1,0 +1,9 @@
+---
+name: 'Tool Guardian'
+description: 'Blocks dangerous tool operations (destructive file ops, force pushes, DB drops) before the Copilot coding agent executes them'
+tags: ['security', 'safety', 'preToolUse', 'guardrails']
+---
+
+# Tool Guardian Hook
+
+Copied from https://github.com/github/awesome-copilot/blob/main/hooks/tool-guardian/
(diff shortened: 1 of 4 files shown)


david-mears-2:
This feels more like a personal preference than something to encode in the repo, given it is possible to scope the hook to a user: https://github.com/github/copilot-cli/issues/2430#event-24275197112

